PR #1, opened 2015-12-22
Adds project structure around existing code

adriancole:
With this in place, you can enable travis, which can test pull requests in case they break the build.

Here's a list of other things:
 * Adds maven build structure
 * Defaults language level to 1.8, except main code in core which is 1.7
 * Adds apache license and plugin to verify it
 * Reformats code to square style (similar to google style)
 * Adds an example unit test we can use when there are concrete types

If you add me to this project or repo, I can help get travis working.


adriancole:
ps I don't mind if this history is lost (via squashing etc) if eventually gets to master.

mvnw
```diff
@@ -0,0 +1,234 @@
+#!/bin/sh
```
adriancole: fyi files and directories name mvn and maven were generated. This is called "maven wrapper" and avoids version conflicts with whatever version someone installed locally.

opentracing-core/src/java/io/opentracing/OpenTracer.java
```diff
@@ -1,19 +0,0 @@
-package io.opentracing;
```
adriancole: these classes miss git compare on format, Only change besides that and adding license header was a missing import in one of them.

src/etc/header.txt
```diff
@@ -0,0 +1,11 @@
+Copyright 2015 The OpenTracing Authors
```
adriancole: "The OpenTracing Authors" is a way out of enumerating everyone, so equally nice to all. I use this approach in many projects, and got the idea from guava.

mvnw
bensigelman: cool

bensigelman: if it's easy, can you add a pointer somewhere (in the README?) about how this was generated?

src/etc/header.txt
bensigelman: sgtm!

bensigelman:
Thanks so much for putting this together. BTW, feel free to just merge your own PRs for stuff like this at this early stage... I will have no objections.

PR #2, opened 2016-01-10
Adds project structure

adriancole:
This is the same as #1, except it doesn't base over any code. Code will be incrementally by porting from api-python and checking against other documentation.

With this in place, you can enable travis, which can test pull requests
in case they break the build.

* Adds README based on api-python
* Adds maven build structure
* Defaults language level to 1.8, except main code in core which is 1.7
* Adds apache license and plugin to verify it
* Reformats code to square style (similar to google style)
* Adds an example unit test we can use when there are concrete types

See https://github.com/opentracing/api-python


adriancole:
ok travis is working. goal here is to have the project structure in place, so that types can be added along with unit tests incrementally.

.travis.yml
```diff
@@ -0,0 +1,15 @@
...
+  - $HOME/.m2
+
+language: java
+
+jdk:
+  - oraclejdk8
+
+script:
+  # use the maven wrapper as opposed to whatever version travis has
+  - ./mvnw install -nsu
```
adriancole: this passed https://travis-ci.org/opentracing/api-java

adriancole:
eagerly merging as except for the readme, it is a copy of already reviewed work

PR #4, opened 2016-01-15
Adds initial interfaces based on api-go

adriancole:
This is mostly a port from python, except for a couple notes.
* Adds an IllegalArgumentException to `fromXXX`, if the input is malformed.
* Adds `EncodedTraceContext` type as Java cannot return multiple results.

Any other changes are unintentional.

See https://github.com/opentracing/api-go/tree/master/opentracing


opentracing/src/main/java/opentracing/Tracer.java
```diff
@@ -0,0 +1,45 @@
...
+ * is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express
+ * or implied. See the License for the specific language governing permissions and limitations under
+ * the License.
+ */
+package opentracing;
+
+/**
+ * Tracer is the entry point API between instrumentation code and the tracing implementation.
+ */
+public interface Tracer {
```
yurishkuro: In Go/Python versoins the Tracer extends TraceSourceContext. If we add the inheritance here, the method names may need to be changed to match Go/Python/semantics-doc, because they would look confusing: `tracer.fromBinary()` or `tracer.newRoot()`.

opentracing/src/main/java/opentracing/TraceContext.java
```diff
@@ -0,0 +1,51 @@
...
+   * <p>Note 2: attributes are sent in-band with every subsequent local and remote calls, so this
+   * feature must be used with care.
+   *
+   * <p>Note 3: keys are case-insensitive, to allow propagation via HTTP headers. Keys must match
+   * regex {@code (?i:[a-z0-9][-a-z0-9]*)}
+   *
+   * @param key attribute key, matching regex {@code (?i:[a-z0-9][-a-z0-9]*)}
+   * @param value value associated with the key.
+   */
+  TraceContext setAttribute(String key, String value);
```
prb: Is there an expected exception if a key is invalid (`IllegalArgumentException`?)?  Is there an expectation of last-`setAttribute`-wins or are attempts to set a previously set attribute blocked?

prb: Can a `null` value be set?

yurishkuro: good questions, @prb. cc @bensigelman, we should address in semantic doc.

My take is:
- invalid keys lead to unpredictable behavior (we mentioned it elsewhere), but not to an exception
- last `setAttribute` call overrides the attribute for future children spans
- null value *can* be set and is equivalent to deleting the attribute

dankosaur: Makes sense to me -- is the null-delete behavior documented anywhere?

bensigelman: I would argue against deletion semantics... it's already a confusing concept since attributes only propagate to *future* children, meaning they're not truly trace-wide (don't affect past children or other descendants of parents ("cousins", etc). Adding deletion into the mix just makes this harder to reason about. Finally, it's easier to add features than subtract them. Thoughts?

adriancole: I'm actually in favor of validating the keys on the way in (with illegal
argument exception). There are few things harder to debug than why
something was tossed very down stream for a documented and preventable
reason.

opentracing/src/main/java/opentracing/Tracer.java
adriancole: In effective java, which a lot modern java takes from, composition is
favored over inheritance. I was already a bit troubled, writing this, as
another inheritance wants an one object to also be a codec.

I think if we weren't explicit about requiring inheritance, we can write
and wire cleaner components in java.

opentracing/src/main/java/opentracing/TraceContext.java
prb: @bensigelman :+1: to forbidding deletion.  I can't think of a reason to support deletion given the immutability of parent contexts; it's always possible to not add the thing you'd otherwise delete.

yurishkuro: Deletion can be thought of as an update with a tombstone, so by extension we would need to disallow updates. I personally don't feel that a mid-trace update adds any more mental complexity than a mid-trace insert. I would expect well-behaving clients to set all required trace attributes upfront, before creating any child spans, as otherwise both inserts and updates are hard to reason about. On the other hand, I feel that disallowing updates puts clients at a disadvantage, they are unable to control their own execution call graph, because some process possibly way up the stack already set a value for an attribute that this client happened to use for another purpose with its children.

opentracing/src/main/java/opentracing/TraceContextDecoder.java
```diff
@@ -0,0 +1,41 @@
...
+
+import java.util.Map;
+
+/**
+ * Decodes a trace context from binary or text formats.
+ *
+ * @see TraceContextEncoder
+ */
+// See https://github.com/opentracing/opentracing.github.io/issues/21 about folding codec together
+public interface TraceContextDecoder {
```
prb: I'm a bit puzzled with the design of the encoder and decoder here.  Some wonderings:

1. Why have every en/de-coder responsible for the two kinds of operations here?
2. Why is the `EncodedTraceContext`

adriancole: >
>
>    1. Why have every en/de-coder responsible for the two kinds of
>    operations here?
>
> I don't recall the exact reason, but I think the reason for the split was
that not everyone is propagating context in text format (such as zipkin
does). Ex. in zipkin, propagation is via HTTP headers, so a map of string,
string works. the binary encoding isn't a feature we'd use.


>
>    1. Why is the EncodedTraceContext in the middle? Maybe straight from a
>    parameterized type T to aTraceContext implementation?
>
> There's probably an opportunity to link (or-repaste) rationale
documentation about the split of trace id vs other attributes. In following
other impls, they return a list of 2 collections each.

I agree that the encoded form could be more straight-forward if a type T.
Ex. I could use OkHttp's Headers object or a grpc Context directly that
way. If you feel interested enough, I could touch more on why I think
"straight-pipe" could be helpful for implementors.. want to raise an issue
here? https://github.com/opentracing/opentracing.github.io/issues

opentracing/src/main/java/opentracing/TraceContext.java
adriancole: >
> > +  TraceContext setAttribute(String key, String value);
>
> @bensigelman <https://github.com/bensigelman> [image: :+1:] to forbidding
> deletion. I can't think of a reason to support deletion given the
> immutability of parent contexts; it's always possible to not add the thing
> you'd otherwise delete.
>
In finagle, there's a flags propagation field, which includes debug. This
can be changed on a hop. Not necessarily in tracing, but I've seen other
side-channels which propagate TTL-style data, either hop count or deadline.
Since this isn't a general purpose context, we can punt such a feature I
suspect.

opentracing/src/main/java/opentracing/TraceContextDecoder.java
bensigelman: TL;DR: the oddities in the coding API grew out of the following:
- a requirement to avoid cruft and/or unnecessary copies of TraceContext representation data (not of the trace attributes as much as whatever "context ids" are intrinsic to the particular TraceContext impl... trace_id, span_id, that sort of thing)... i.e., if a TraceContext knows that it only has a trace_id and span_id, it can encode those positionally rather than as a map.
- to that end, the context identifiers are often bounded and – per a comment @yurishkuro made a little while back – sometimes encoded in a fixed-length buffer; the trace attributes are not, hence the way the go and python APIs separate those two pieces of the encoded `TraceContext` (*)
- re supporting text and binary but nothing else: TBH, I would probably have just gone for binary and told folks that wanted to stuff them in HTTP headers to base64 encode or whatever. But my understanding is that many zipkin users really like having the various fields split out in human-readable (and software-router-accessible) fashion in HTTP headers, hence the text format. We hope (?) that there will only be two, hence cramming them into one interface.

Now, presently the encoder and decoder are separate interfaces... there is an issue open to change that and I haven't heard about any opposition. https://github.com/opentracing/opentracing.github.io/issues/21

(*): @adriancole I just noticed that you are encoding/decoding to/from a *single* `byte[]` and a *single* `Map<...>`, not two like py and go. Is that deliberate? If not, let's change it; if so, I'm curious about the semantic divergence.

adriancole: > (): @adriancole I just noticed that you are encoding/decoding to/from a
*single byte[] and a single Map<...>, not two like py and go. Is that
deliberate? If not, let's change it; if so, I'm curious about the semantic
divergence.

There are in fact two values. I think you are mistaking the type parameter
as being single value.

We could make it `List<>`, but yeah it is weird enough already.

bensigelman: Ah, my brain skipped the EncodedTC intermediary. Disregard!

adriancole: > re supporting text and binary but nothing else: TBH, I would probably
have just gone for binary and told folks that wanted to stuff them in HTTP
headers to base64 encode or whatever. But my understanding is that many
zipkin users really like having the various fields split out in
human-readable (and software-router-accessible) fashion in HTTP headers,
hence the text format. We hope (?) that there will only be two, hence
cramming them into one interface.
I dont know how many are processing zipkin headers as intermediaries, but
there's certainly a support use case of taking the value of the trace id
header and plopping it into a search box.

opentracing/src/main/java/opentracing/Span.java
```diff
@@ -0,0 +1,67 @@
...
+   * Indicates that the work represented by this span has been completed or terminated, and is ready
+   * to be sent to the Reporter. <p>If any tags / logs need to be added to the span, it should be
+   * done before calling {@linkplain #finish}, otherwise they may be ignored.
+   */
+  void finish();
+
+  /**
+   * Attaches a key/value pair to the span.
+   *
+   * <p>The set of supported value types is implementation specific. It is the responsibility of the
```
bensigelman: I think we agreed that `value` must be a `BasicType`... i.e., a string, a bool, or a numeric type.

opentracing/src/main/java/opentracing/TraceContextDecoder.java
bensigelman: (good point)

opentracing/src/main/java/opentracing/Tracer.java
```diff
@@ -0,0 +1,45 @@
...
+   * context.
+   *
+   * <p>This method should be used by services that receive tracing info from upstream.
+   *
+   * @param operationName the service's own name for the end-point that received the request
+   * represented by this trace and span. The domain of names must be limited, e.g. do not use UUIDs
+   * or entity IDs or timestamps as part of the name.
+   * @param parent Trace Context of a client span started elsewhere and whose trace we're joining.
+   */
+  Span joinTrace(String operationName, TraceContext parent);
```
bensigelman: so the Go API has `startTrace` and `joinTrace` equivs, but also this:

https://github.com/opentracing/opentracing-go/blob/master/tracer.go#L46

Python has something similar in that a `Span` can be constructed given an explicit `TraceContext` to bind to.

@yurishkuro: did we add that so that the golang Zipkin impl could retain the same `span_id` on both sides of an RPC? Or was it just to be more general? In any case, we should remove it in Go/Python xor add it here.

yurishkuro: no, retaining the same span_id can be easily done as internal impl of zipkin when you do join_trace.

StartSpanWithContext was primarily added to support "debug" traces, where the instrumentation decides that a trace should be sampled no matter what, creates a new context, sets a debug attribute on the context, and then creates a span with that context.

opentracing/src/main/java/opentracing/Span.java
adriancole: this was a copy-paste docs. I'll revise to do this with overloading as there are no union types in java.

adriancole:
rejigged around the go project instead. This invalidated a lot of comments.. will go back and tidy what I can, but PTAL

adriancole:
cc @dsyer @michaelsembwever @marcingrzejszczak might be interested

opentracing/src/main/java/opentracing/Tracer.java
```diff
@@ -0,0 +1,54 @@
...
+   * #startTrace(String)} or a fresh child context like {@link #joinTrace(String, TraceContext)}).
+   *
+   * <p>Note that the following calls are equivalent
+   * <pre>{@code
+   * Span feed = tracer.startSpanWithContext("GetFeed", traceContextSource.newRoot());
+   * ...
+   * Span feed = tracer.startTrace("GetFeed");
+   * }</pre>
+   */
+  Span startSpanWithContext(String operationName, TraceContext context);
```
adriancole: @dsyer @marcingrzejszczak ps best I can tell, this is one you manually start a span (ex affect its start time)

opentracing/src/main/java/opentracing/Span.java
```diff
@@ -0,0 +1,93 @@
...
+   * <p>If there is a pre-existing tag set for {@code key}, it is overwritten.
+   */
+  // overloaded 3x to support the BasicType concern
+  Span setTag(String key, String value);
+
+  /** Same as {@link #setTag(String, String)}, but for boolean values. */
+  Span setTag(String key, boolean value);
+
+  /** Same as {@link #setTag(String, String)}, but for numeric values. */
+  // numbers kindof suck.. we've no idea if this is a float, how many bits, etc.
```
bensigelman: @adriancole should we instead have one `setTag` for `Long` and one for `Double`?

opentracing/src/main/java/opentracing/Span.java
```diff
@@ -0,0 +1,93 @@
...
+   *         Transaction.builder().amount(11).currency("USD").build(),
+   *       )
+   *     )
+   * );
+   * }</pre>
+   *
+   * @param message {@link Formatter format string} that can refer to fields in the args payload.
+   * @param args arbitrary payload
+   */
+  // See https://github.com/opentracing/opentracing.github.io/issues/30 about parameterization
```
bensigelman: (ignoring this until https://github.com/opentracing/opentracing.github.io/issues/30 resolves somewhere)

opentracing/src/main/java/opentracing/TraceContext.java
```diff
@@ -0,0 +1,66 @@
...
+   * <p>IMPORTANT NOTE #2: Use this thoughtfully and with care. Every key and value is copied into
+   * every local *and remote* child of this TraceContext, and that can add up to a lot of network
+   * and cpu overhead.
+   *
+   * <p>IMPORTANT NOTE #3: Trace attributes keys have a restricted format: implementations may wish
+   * to use them as HTTP header keys (or key suffixes), and of course HTTP headers are case
+   * insensitive.
+   *
+   * @param restrictedKey MUST match the regular expression `(?i:[a-z0-9][-a-z0-9]*)` and is
+   * case-insensitive. That is, it must start with a letter or number, and the remaining characters
```
bensigelman: @adriancole not sure you ever responded to my point about this... I'm happy for us to say (here) that the keys *must* be lowercase as long as we are clear that certain intermediaries may meddle with them (not before calling setAttribute, but when reassembling attributes when they come off the wire of HTTP headers, etc, etc). If that doesn't make sense let me know.

opentracing/src/main/java/opentracing/TraceContextCodec.java
```diff
@@ -0,0 +1,71 @@
...
+/**
+ * Encodes or Decodes a {@linkplain TraceContext trace context} in binary or text formats.
+ *
+ * <p>The toXXX methods are expected to serialize trace contexts into a pair of values representing
+ * separately the trace context / span identity, and the trace attributes. This is done specifically
+ * for binary protocols that may represent tracing identity in a dedicated fixed-length slot in the
+ * binary message format, so that it can be inspected efficiently by the middleware / routing layers
+ * without parsing the whole message.
+ */
+public interface TraceContextCodec {
```
bensigelman: yay (unified cod + dec)

opentracing/src/main/java/opentracing/Tracer.java
bensigelman: I don't think this has any better hope of affecting the start timestamp than the other methods on `Tracer`, but we do need to support that. See https://github.com/opentracing/opentracing.github.io/issues/20 (and while we should absolutely discuss the best way to address this, that issue is probably better than this PR as it is a poly-platform deficiency)

opentracing/src/main/java/opentracing/TraceContextSource.java
```diff
@@ -0,0 +1,48 @@
...
+ */
+package opentracing;
+
+import java.util.Map;
+
+/**
+ * Long-lived interface that knows how to create a root {@linkplain TraceContext} and encode/decode
+ * any other.
+ */
+public interface TraceContextSource {
```
bensigelman: I'm good with not-having `Tracer` inherit from this, but shouldn't it (`Tracer`, I mean) have a `TraceContextSource source()` or `TraceContextSource traceContextSource()` method?

bensigelman:
@adriancole aside from https://github.com/opentracing/opentracing.github.io/issues/20 (which we can deal with in a follow-on) and my question about how to get the `TraceContextSource` from the `Tracer`, this looks good to me. (Or maybe your thought was that the opentracing-java user would keep track of the TCSource and Tracer separately?)

opentracing/src/main/java/opentracing/TraceContextSource.java
yurishkuro: as well as `codec()`

bensigelman: `codec()` is accessible here -- though if we want to flatten everything into Tracer, that works for me, too.

opentracing/src/main/java/opentracing/Span.java
michaelsembwever: what benefit to you have in mind @bensigelman ?

Number is the superclass to both Long and Double.

adriancole:
IMHO, TraceContextSource (and TraceContextCodec for that matter) have no business being attached to other objects either via inheritance or composition. I'm quite certain there will be folks who do not want to make paths to these visible to others. I do know that there are folks who want things to be attached to each-other.

opentracing/src/main/java/opentracing/TraceContextSource.java
michaelsembwever: Sounds like then the class design for `Tracer` is not taking responsibility for its actions if users of an object require access to its context. And not providing it creates the simpler API which should always be favoured. (Adding it in the future if a good use case comes along doesn't break compatibility, but removing it does).

Do we have a use-case or example where `source()` is needed?

opentracing/src/main/java/opentracing/TraceContext.java
adriancole: I think it is better to say must be lowercase. if an intermediary changes them to camel-case, that sounds like an edge-case that could be addressed at the transport binding (a matter we don't have a place to discuss at the moment)

opentracing/src/main/java/opentracing/Span.java
bensigelman: This is in response to

> // numbers kindof suck.. we've no idea if this is a float, how many bits, etc.

If we force Long and Double, then we have an idea if this is a float, how many bits, etc. :)  [Not that impls couldn't do gross `instanceof` stuff instead]

Just a thought, no biggie.

opentracing/src/main/java/opentracing/TraceContext.java
bensigelman: We do have a possible place to discuss it: https://github.com/opentracing/opentracing-java/pull/4/files#diff-56ae83f39e27fe6f56501e4788fecda2R70

Namely, the `Map<String, String>` of trace attributes would need to have lowercase keys as well.

opentracing/src/main/java/opentracing/TraceContextSource.java
bensigelman: Sure, yeah... if you are writing the code that serializes a `TraceContext` to the wire (e.g., in HTTP headers, or stuffed into a binary protocol like thrift or grpc somewhere) – or deserializes the `TraceContext` on the other side – the `codec()` is needed. The non-`codec()` parts of the `TraceContextSource` are useful for code that can be shared by most `Tracer` implementations... less important than the `codec()` in my mind, but important nonetheless. If you speak Go, this is how the "standard" `Tracer` uses `TraceContextSource`: https://github.com/opentracing/opentracing-go/blob/master/standardtracer/standardimpl.go#L129 (there are other similar call sites in that file).

michaelsembwever: You'll have to forgive me if i'm behind the ball here.

Isn't tracing instrumentation just about collecting Spans?
Why put the Tracer, TracerContext, or TracerContextSource, on the wire at all?

(Which questions why `TraceContext traceContext();`  is in `Span`)

adriancole:
there are some existing threads on this PR that need to be forked into issues or resolved here. I'm going to be out for a while, so I'll need someone to decide offline which path to take, and either merge this or make changes to the shared branch w/o me.

michaelsembwever:
I've got a branch based off this that proposes a number of simplifications
and naming improvements.
Would be great if i could push that. That should let discussions keep
running.

On 21 January 2016 at 18:33, Adrian Cole <notifications@github.com> wrote:

> there are some existing threads on this PR that need to be forked into
> issues or resolved here. I'm going to be out for a while, so I'll need
> someone to decide offline which path to take, and either merge this or make
> changes to the shared branch w/o me.
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/opentracing/opentracing-java/pull/4#issuecomment-173483478>
> .
>



-- 
Mick Semb Wever
Australia

The Last Pickle
Apache Cassandra Consulting
http://www.thelastpickle.com

adriancole:
please do

michaelsembwever:
see https://github.com/opentracing/opentracing-java/pull/7

michaelsembwever:
I'm closing this PR  and re-opening it from scratch because most of the above comments are no longer relevant now that https://github.com/opentracing/opentracing-java/pull/10 https://github.com/opentracing/opentracing-java/pull/12  https://github.com/opentracing/opentracing-java/pull/8 and https://github.com/opentracing/opentracing-java/pull/9 have been merged in.

PR #7, opened 2016-01-22
In discrete steps, Simplify the API

michaelsembwever:
This PR contains a sequence of commits to simplify the API.

Class renames:
 - TraceContext to SpanContext
 - TraceContextSource to TraceContext
 - TraceContextCodec to TraceCodec

Removing porcelain api:
 - Span.createChild(..)
 - TraceContext.ChildSpanContext
 - logging methods in Span (replaced with `event(message, payload)`).

Look in the individual commits for the rationale to each.

It's clearer to see the value of the changes by looking at the HEAD of the branch rather than evaluating the patch.

### DIFF
--- a/opentracing/src/main/java/opentracing/Span.java
+++ b/opentracing/src/main/java/opentracing/Span.java
@@ -13,7 +13,7 @@
  */
 package opentracing;
 
-import java.util.Formatter;
+import java.util.Map;
 
 /**
  * Span represents an active, un-finished span in the opentracing system.
@@ -56,29 +56,7 @@ public interface Span {
   Span setTag(String key, Number value);
 
   /**
-   * {@code message} is a format string and can refer to fields in the payload by path, like so:
-   *
-   * <pre>{@code
-   *
-   * span.info("first transaction is worth ${transactions[0].amount} ${transactions[0].currency}",
-   *     ImmutableMap.of(
-   *       "transactions", asList(
-   *         Transaction.builder().amount(10).currency("USD").build(),
-   *         Transaction.builder().amount(11).currency("USD").build(),
-   *       )
-   *     )
-   * );
-   * }</pre>
-   *
-   * @param message {@link Formatter format string} that can refer to fields in the args payload.
-   * @param args arbitrary payload
-   */
-  // See https://github.com/opentracing/opentracing.github.io/issues/30 about parameterization
-  Span info(String message, Object... args);
-
-  /** Same as {@link #info}, but for warnings. */
-  Span warning(String message, Object... args);
-
-  /** Same as {@link #info}, but for errors. */
-  Span error(String message, Object... args);
+   * @todo
+   **/
+  Span event(String message, Map<String,String> payload);
 }
--- a/opentracing/src/main/java/opentracing/Span.java
+++ b/opentracing/src/main/java/opentracing/Span.java
@@ -28,15 +28,6 @@ public interface Span {
   SpanContext childContext();
 
   /**
-   * Denotes the beginning of a subordinate unit of work.
-   *
-   * @param operationName name of the operation represened by the new span from the perspective of
-   * the current service.
-   * @return a new child Span in "started" state.
-   */
-  Span startChild(String operationName);
-
-  /**
    * Sets the end timestamp and records the span.
    *
    * <p>This should be the last call made to any span instance, and to do otherwise leads to
--- a/opentracing/src/main/java/opentracing/Span.java
+++ b/opentracing/src/main/java/opentracing/Span.java
@@ -25,7 +25,7 @@ public interface Span {
   /**
    * Suitable for serializing over the wire, etc.
    */
-  TraceContext traceContext();
+  SpanContext childContext();
 
   /**
    * Denotes the beginning of a subordinate unit of work.
--- a/opentracing/src/main/java/opentracing/SpanContext.java
+++ b/opentracing/src/main/java/opentracing/SpanContext.java
@@ -14,23 +14,23 @@
 package opentracing;
 
 /**
- * TraceContext encpasulates the smallest amount of state needed to describe a Span's identity
- * within a larger [potentially distributed] trace. The TraceContext is not intended to encode the
+ * SpanContext encpasulates the smallest amount of state needed to describe a Span's identity
+ * within a larger [potentially distributed] trace. The SpanContext is not intended to encode the
  * span's operation name, timing, or log data, but merely any unique identifiers (etc) needed to
  * contextualize it within a larger trace tree.
  *
- * <p>TraceContexts are sufficient to propagate the, well, *context* of a particular trace between
+ * <p>SpanContexts are sufficient to propagate the, well, *context* of a particular trace between
  * processes.
  *
- * <p>TraceContext also support a simple string map of "trace attributes". These trace attributes
+ * <p>SpanContext also support a simple string map of "trace attributes". These trace attributes
  * are special in that they are propagated *in-band*, presumably alongside application data. See the
  * documentation for {@link #setAttribute(String, String)} for more details and some important
  * caveats.
  */
-public interface TraceContext {
+public interface SpanContext {
 
   /**
-   * Sets a tag on this TraceContext that also propagates to future children per {@link
+   * Sets a tag on this SpanContext that also propagates to future children per {@link
    * TraceContextSource#newChild(TraceContext)}.
    *
    * <p>Trace attributes enables powerful functionality given a full-stack opentracing integration
@@ -54,7 +54,7 @@ public interface TraceContext {
    * must be letters, numbers, or hyphens. undefined behavior results if the `restrictedKey` does
    * not meet these criteria.
    */
-  TraceContext setAttribute(String restrictedKey, String value);
+  SpanContext setAttribute(String restrictedKey, String value);
 
   /**
    * Gets the value for a trace tag given its key. Returns Null if the value isn't found in this
--- a/opentracing/src/main/java/opentracing/TraceCodec.java
+++ b/opentracing/src/main/java/opentracing/TraceCodec.java
@@ -16,7 +16,7 @@ package opentracing;
 import java.util.Map;
 
 /**
- * Encodes or Decodes a {@link SpanContext trace context} in binary or text formats.
+ * Encodes or Decodes a {@link SpanContext context} in binary or text formats.
  *
  * <p>The toXXX methods are expected to serialize trace contexts into a pair of values representing
  * separately the trace context / span identity, and the trace attributes. This is done specifically
@@ -24,7 +24,7 @@ import java.util.Map;
  * binary message format, so that it can be inspected efficiently by the middleware / routing layers
  * without parsing the whole message.
  */
-public interface TraceContextCodec {
+public interface TraceCodec {
 
   /**
    * Implementation-specific format of a span's identity along with any span attributes.
--- a/opentracing/src/main/java/opentracing/TraceContext.java
+++ b/opentracing/src/main/java/opentracing/TraceContext.java
@@ -13,8 +13,6 @@
  */
 package opentracing;
 
-import java.util.Map;
-
 /**
  * Long-lived interface that knows how to create a root {@link SpanContext} and encode/decode
  * any other.
@@ -34,15 +32,4 @@ public interface TraceContext {
    */
   SpanContext newRoot();
 
-  /**
-   * Creates a child context for {@code parent}, and returns both that child's own
-   * TraceContext as well as any Tags that should be added to the child's Span.
-   */
-  ChildTraceContext newChild(SpanContext parent);
-
-  interface ChildTraceContext {
-    SpanContext child();
-
-    Map<String, Object> tags();
-  }
 }
--- a/opentracing/src/main/java/opentracing/TraceContext.java
+++ b/opentracing/src/main/java/opentracing/TraceContext.java
@@ -19,7 +19,7 @@ import java.util.Map;
  * Long-lived interface that knows how to create a root {@link SpanContext} and encode/decode
  * any other.
  */
-public interface TraceContextSource {
+public interface TraceContext {
 
   /**
    * Encodes or Decodes a {@link SpanContext trace context} in binary or text formats.
--- a/opentracing/src/main/java/opentracing/TraceContext.java
+++ b/opentracing/src/main/java/opentracing/TraceContext.java
@@ -24,7 +24,7 @@ public interface TraceContext {
   /**
    * Encodes or Decodes a {@link SpanContext trace context} in binary or text formats.
    */
-  TraceContextCodec codec();
+  TraceCodec codec();
 
   /**
    * Create a SpanContext which has no parent (and thus begins its own trace).
--- a/opentracing/src/main/java/opentracing/TraceContextSource.java
+++ b/opentracing/src/main/java/opentracing/TraceContextSource.java
@@ -16,32 +16,32 @@ package opentracing;
 import java.util.Map;
 
 /**
- * Long-lived interface that knows how to create a root {@linkplain TraceContext} and encode/decode
+ * Long-lived interface that knows how to create a root {@link SpanContext} and encode/decode
  * any other.
  */
 public interface TraceContextSource {
 
   /**
-   * Encodes or Decodes a {@linkplain TraceContext trace context} in binary or text formats.
+   * Encodes or Decodes a {@link SpanContext trace context} in binary or text formats.
    */
   TraceContextCodec codec();
 
   /**
-   * Create a TraceContext which has no parent (and thus begins its own trace).
+   * Create a SpanContext which has no parent (and thus begins its own trace).
    *
    * <p>A TraceContextSource must always return the same type in successive calls to
    * NewRootTraceContext().
    */
-  TraceContext newRoot();
+  SpanContext newRoot();
 
   /**
    * Creates a child context for {@code parent}, and returns both that child's own
    * TraceContext as well as any Tags that should be added to the child's Span.
    */
-  ChildTraceContext newChild(TraceContext parent);
+  ChildTraceContext newChild(SpanContext parent);
 
   interface ChildTraceContext {
-    TraceContext child();
+    SpanContext child();
 
     Map<String, Object> tags();
   }
--- a/opentracing/src/main/java/opentracing/Tracer.java
+++ b/opentracing/src/main/java/opentracing/Tracer.java
@@ -36,7 +36,7 @@ public interface Tracer {
   /**
    * Like {@link #startTrace(String)}, but the returned span is made a child of {@code parent}.
    */
-  Span joinTrace(String operationName, TraceContext parent);
+  Span joinTrace(String operationName, SpanContext parent);
 
   /**
    * StartSpanWithContext returns a span with the given {@code operationName} and an association
@@ -50,5 +50,5 @@ public interface Tracer {
    * Span feed = tracer.startTrace("GetFeed");
    * }</pre>
    */
-  Span startSpanWithContext(String operationName, TraceContext context);
+  Span startSpanWithContext(String operationName, SpanContext context);
 }
--- a/opentracing/src/main/java/opentracing/TraceContextCodec.java
+++ b/opentracing/src/main/java/opentracing/TraceContextCodec.java
@@ -16,7 +16,7 @@ package opentracing;
 import java.util.Map;
 
 /**
- * Encodes or Decodes a {@linkplain TraceContext trace context} in binary or text formats.
+ * Encodes or Decodes a {@link SpanContext trace context} in binary or text formats.
  *
  * <p>The toXXX methods are expected to serialize trace contexts into a pair of values representing
  * separately the trace context / span identity, and the trace attributes. This is done specifically
@@ -27,13 +27,13 @@ import java.util.Map;
 public interface TraceContextCodec {
 
   /**
-   * Implementation-specific format of a span's identity along with any trace attributes.
+   * Implementation-specific format of a span's identity along with any span attributes.
    *
    * @param <E> encoding, for example {@code byte[]} for binary, or {@code Map<String, String>} for
    * text.
    */
-  // Can instead explicitly create BinaryEncodedTraceContext, TextEncodedTraceContext, just.. cluttery
-  interface EncodedTraceContext<E> {
+  // Can instead explicitly create BinaryEncodedSpanContext, TextEncodedSpanContext, just.. cluttery
+  interface EncodedSpanContext<E> {
     /** Encoded span identifier. */
     E spanIdentity();
 
@@ -42,30 +42,30 @@ public interface TraceContextCodec {
   }
 
   /**
-   * Encodes the trace context into a binary representation of the span's identity and trace
+   * Encodes the span context into a binary representation of the span's identity and trace
    * attributes.
    */
-  EncodedTraceContext<byte[]> toBinary(TraceContext tc);
+  EncodedSpanContext<byte[]> toBinary(SpanContext tc);
 
   /**
-   * Decodes a trace context from a binary representation of the span's identity and trace
+   * Decodes a span context from a binary representation of the span's identity and trace
    * attributes.
    *
    * @throws IllegalArgumentException if the encoded data is malformed.
    */
-  TraceContext fromBinary(EncodedTraceContext<byte[]> encoded);
+  SpanContext fromBinary(EncodedSpanContext<byte[]> encoded);
 
   /**
-   * Encodes the trace context into a text representation of the span's identity and trace
+   * Encodes the span context into a text representation of the span's identity and trace
    * attributes.
    */
-  EncodedTraceContext<Map<String, String>> toText(TraceContext tc);
+  EncodedSpanContext<Map<String, String>> toText(SpanContext tc);
 
   /**
-   * Decodes a trace context from a text representation of the span's identity and trace
+   * Decodes a span context from a text representation of the span's identity and trace
    * attributes.
    *
    * @throws IllegalArgumentException if the encoded data is malformed.
    */
-  TraceContext fromText(EncodedTraceContext<Map<String, String>> encoded);
+  SpanContext fromText(EncodedSpanContext<Map<String, String>> encoded);
 }


opentracing/src/main/java/opentracing/Span.java
```diff
@@ -65,29 +56,7 @@
...
-  Span info(String message, Object... args);
-
-  /** Same as {@link #info}, but for warnings. */
-  Span warning(String message, Object... args);
-
-  /** Same as {@link #info}, but for errors. */
-  Span error(String message, Object... args);
+   * @todo
+   **/
+  Span event(String message, Map<String,String> payload);
```
michaelsembwever: The api of payload is still unclear to me
 ref: https://github.com/opentracing/opentracing.github.io/issues/30#issuecomment-173786466

opentracing/src/main/java/opentracing/TraceCodec.java
```diff
@@ -24,16 +24,16 @@
  * binary message format, so that it can be inspected efficiently by the middleware / routing layers
  * without parsing the whole message.
  */
-public interface TraceContextCodec {
+public interface TraceCodec {
```
michaelsembwever: I'm unsure about the need for this class.

What is the need to expose serialisation details to the implementer?

opentracing/src/main/java/opentracing/SpanContext.java
```diff
@@ -0,0 +1,66 @@
...
+   * and cpu overhead.
+   *
+   * <p>IMPORTANT NOTE #3: Trace attributes keys have a restricted format: implementations may wish
+   * to use them as HTTP header keys (or key suffixes), and of course HTTP headers are case
+   * insensitive.
+   *
+   * @param restrictedKey MUST match the regular expression `(?i:[a-z0-9][-a-z0-9]*)` and is
+   * case-insensitive. That is, it must start with a letter or number, and the remaining characters
+   * must be letters, numbers, or hyphens. undefined behavior results if the `restrictedKey` does
+   * not meet these criteria.
```
michaelsembwever: There was a discussion on gitter regarding removing this restriction.

It should be an implementation detail. For example if the implementation is going to put these attributes keys into HTTP headers than naturally it's the implementation's responsibility to escape them accordingly.

Rather than enforce these restrictions the documentation should provide the recommendation to use such a regular expression so to avoid such possible escaping side-effects.

opentracing/src/main/java/opentracing/TraceCodec.java
yurishkuro: @michaelsembwever  If we don't expose a codec, how else would a framework or an application serialize the context into whichever protocol they are using for RPC?

yurishkuro: If we assume we change the name to SpanContext, why not call this SpanContextCodec?

michaelsembwever: > how else would a framework or an application serialize the context into whichever protocol they are using for RPC?

Any way they like.

The specification explains the need for serialisation.
But I don't (yet) see the need for the API to impose how it should be done.

michaelsembwever: >  If we assume we change the name to SpanContext, why not call this SpanContextCodec?

See the Rationalie in the commit msg: https://github.com/opentracing/opentracing-java/commit/02d5d15bb51866afc7b8ed9d035798032c0b0848

opentracing/src/main/java/opentracing/Span.java
```diff
@@ -25,16 +25,7 @@
...
-
-  /**
-   * Denotes the beginning of a subordinate unit of work.
-   *
-   * @param operationName name of the operation represened by the new span from the perspective of
-   * the current service.
-   * @return a new child Span in "started" state.
-   */
-  Span startChild(String operationName);
+  SpanContext childContext();
```
yurishkuro: what is the use case where one needs a child context, but does not need a child span?

opentracing/src/main/java/opentracing/TraceContextSource.java
```diff
@@ -1,48 +0,0 @@
...
-   */
-  TraceContext newRoot();
-
-  /**
-   * Creates a child context for {@code parent}, and returns both that child's own
-   * TraceContext as well as any Tags that should be added to the child's Span.
-   */
-  ChildTraceContext newChild(TraceContext parent);
-
-  interface ChildTraceContext {
```
michaelsembwever: @adriancole I removed this interface and the method above. See https://github.com/opentracing/opentracing-java/commit/6cc58b53bb9cf3065a6efcf234fd406bfcf4883a

Can you help me understand the need for ChildTraceContext (or ChildSpanContext)?

opentracing/src/main/java/opentracing/Span.java
yurishkuro: there must be a way to retrieve span's own context, so that it can be serialized on the wire.

michaelsembwever: > what is the use case where one needs a child context, but does not need a child span?

That's not the change here, if I understand you correctly.

The method `Span startChild(String operationName);` was removed, because it's a convenience API method.

The `TraceContext traceContext();` method renamed to `SpanContext childContext();` so to avoid confusion as to whether this is to obtain the context for this object or the context for its children.

adriancole: Some libraries divorce the concern of creating a span vs adding to it. Ex. this the case in Google Cloud Trace, Sleuth and Brave: you have a reference to add to a span another object created for you.

When trying to implement OT in spring-cloud-sleuth, was that adding a branch ability to the span itself (vs the tracer), requires some relatively heavy changes, such as needing to hold references to what's currently separate instrumentation code. It makes span more a "scopable tracer" and less a DTO that has helpers around time.

adriancole: cc @dsyer lemme know if I mis-spoke here.

michaelsembwever: > what is the use case where one needs a child context, but does not need a child span?

In fact that is the rationale I used for removing `TraceContext.newChild(SpanContext);`
 ref: https://github.com/opentracing/opentracing-java/pull/7#discussion_r50498361

opentracing/src/main/java/opentracing/Span.java
adriancole: ack. before that comment, I'd guess `@Nullable Object payload`

opentracing/src/main/java/opentracing/SpanContext.java
adriancole: ack

opentracing/src/main/java/opentracing/SpanContext.java
```diff
@@ -0,0 +1,66 @@
...
+   */
+  SpanContext setAttribute(String restrictedKey, String value);
+
+  /**
+   * Gets the value for a trace tag given its key. Returns Null if the value isn't found in this
+   * TraceContext.
+   *
+   * @param restrictedKey see {@link #setAttribute(String, String)} notes.
+   */
+  String getAttribute(String restrictedKey);
```
adriancole: ps something here I noticed, not necessarily a problem.. one of the few places we have a getter.

opentracing/src/main/java/opentracing/TraceCodec.java
adriancole: I agree that this doesn't need to be defined, and defining it is distracting at this point. When an OpenTracing implementation is realized against a real system, I'm sure a subtype will do something like this in their own way that makes sense for them. Ex cloud trace [special-cases http codec](https://github.com/GoogleCloudPlatform/cloud-trace-java/blob/master/cloud-trace-sdk-java-core/src/main/java/com/google/cloud/trace/sdk/TraceContext.java#L53), and I expect future finagle things to defer to their context propagation api.

opentracing/src/main/java/opentracing/Tracer.java
```diff
@@ -36,7 +36,7 @@
   /**
    * Like {@link #startTrace(String)}, but the returned span is made a child of {@code parent}.
    */
-  Span joinTrace(String operationName, TraceContext parent);
+  Span joinTrace(String operationName, SpanContext parent);
```
adriancole: I like SpanContext better than TraceContext, it is scoped more reliably. I think maybe at some point people were thinking about propagating attributes upstream (ex. in http, a *response* header), maybe that led to TraceContext? but still it ends up an attribute of one span.

opentracing/src/main/java/opentracing/TraceContextSource.java
adriancole: this was c&p from go. basically there was an idea that you want to seed a child with a set of tags. Since tags aren't in a tracecontext, this is an issue.

This particular issue is implementation-specific.. ex in [Census Tracing](https://github.com/grpc/grpc/pull/4750), tags are in the same set, regardless of whether they are propagated or not. There certainly may be a filtering constraint when a tracing api is layered on that (ex is a child a fork of the parents tags, just changing id?) Maybe @bogdandrutu has ideas on how they will handle this.

In zipkin today, I don't know if any tracers propagate tags implicitly from parent to child, as a function of creating a child. Only exception I can think of is the endpoint of an annotation, but that's not actually scoped to the parent, more often the tracer itself.

Anyway, maybe someone else can elaborate on this feature..

opentracing/src/main/java/opentracing/Span.java
michaelsembwever: >  there must be a way to retrieve span's own context, so that it can be serialized on the wire.

That makes more sense to me.

One creates the initial SpanContext from the `TraceContext.newRoot()`.
From a SpanContext one creates child Spans using `Tracer.joinTrace(..)` or `Tracer.startSpanWithContext(..)`.

So if we rename the method instead to `Span.context()` (since the original `traceContext` is now inaccurate and `spanContext` superfluous), when you add tags to a SpanContext it's either when you have
 - the root SpanContext (from `TraceContext.newRoot()`), or
 - a current Span, and can add tags to its SpanContext (from `Span.context()`)¹

¹ This is not entirely intuitive because you're adding tags to the Span's context. That Span isn't going to take on these tags (because it's already been constructed) but further children of the Span will take on these tags).

As far as exposing the Span's current context for the sake of serialisation I don't (yet) see the need for that for that same reasons as I don't see the need for `TraceCodec`.

> The specification explains the need for serialisation.
> But I don't (yet) see the need for the API to impose how it should be done.

michaelsembwever: > Some libraries divorce the concern of creating a span vs adding to it. 

By removing `Span startChild(String operationName);` are we not helping to solve this.
Keeping the concern of creating spans in the `Tracer`.

opentracing/src/main/java/opentracing/SpanContext.java
michaelsembwever: good point. +1 to remove it.

opentracing/src/main/java/opentracing/Span.java
michaelsembwever: > Then from a SpanContext one creates child Spans using Tracer.joinTrace(..) or Tracer.startSpanWithContext(..).

I have no idea what the difference between these two methods is.
Can someone explain it to me please.

adriancole: > > -  Span startChild(String operationName);
> > +  SpanContext childContext();
>
> Some libraries divorce the concern of creating a span vs adding to it.
>
> By removing Span startChild(String operationName); are we not helping to solve this.
> Keeping the concern of creating spans in the Tracer.

I suppose I inferred that SpanContext childContext() was still an
operation to create a child span? Did you mean SpanContext context()?

opentracing/src/main/java/opentracing/TraceContextSource.java
michaelsembwever: As I (feebly) tried to explain in https://github.com/opentracing/opentracing-java/pull/7#discussion_r50499357 there seems to be two ways to have obtained the current SpanContext.

1) You've created a brand new trace, and still adding spans to the root, are working with the SpanContext from `Tracer.newRoot()`. (I would rename this to rootContext(), since the creation function is really done via `Tracer.startTrace(..)`.

2) You're in the middle of a trace, have hold of a current Span, and are working with a `SpanContext.childContext()`. 

> In zipkin today, I don't know if any tracers propagate tags implicitly from parent to child, as a function of creating a child.

Yes, to me it looks like each SpanContext would inherit tags from any parent SpanContext, while tags added to a SpanContext do not propagate back to the parent SpanContext. A need to not inherit tags from a parent SpanContext would warrant something along the lines of `SpanContext.clear()` and `SpanContext.remove(key)`.

opentracing/src/main/java/opentracing/Span.java
michaelsembwever: > I suppose I inferred that SpanContext childContext() was still an operation to create a child span?

No. The use of `childContext()` is better explained in https://github.com/opentracing/opentracing-java/pull/7#discussion_r50500237

The creation of child Spans is still done via 
 `Tracer.joinTrace("new-span-name", currentSpan.childContext());`

opentracing/src/main/java/opentracing/SpanContext.java
yurishkuro: the whole point of trace attributes is to be accessible to application code at deeper levels of the call tree.

opentracing/src/main/java/opentracing/Span.java
adriancole: ok I get it now. basically this would be documented as a buddy of joinTrace, which would be ancillary to routine tasks like addTag. Implementations would implement this to make their choices wrt which (if any) attributes to pass down. Is the returned childContext immutable in this design? Or would we expect someone to do something like... 

```java
childContext = currentSpan.childContext();
childContext.setAttribute("foo", "bar");
Tracer.joinTrace("new-span-name", childContext);
```

Or are we just saying that this is the hook to make babies.

opentracing/src/main/java/opentracing/SpanContext.java
michaelsembwever: > the whole point of trace attributes is to be accessible to application code at deeper levels of the call tree.

Sure, but that's defined by the specification, rather than by making methods public.

For example: the specification needs to state the Spans created always include all attributes currently in the current SpanContext.

opentracing/src/main/java/opentracing/Span.java
yurishkuro: > The `TraceContext traceContext();` method renamed to `SpanContext childContext();` so to avoid confusion as to whether this is to obtain the context for this object or the context for its children.

The original TraceContext class had two primary purposes:
  1. It itself serves as the span identity, meaning it encapsulates some form of trace-id and span-id and needs to be serialized on the wire.
  1. It was a container for Trace Attributes, or "distributed context", or 'baggage", which also needs to be serialized on the wire.

@michaelsembwever  From various comments on this thread, including the one quoted above, it seems your SpanContext is not meant to perform either of these functions.

opentracing/src/main/java/opentracing/TraceCodec.java
bensigelman: @michaelsembwever @adriancole 

Most organizations I've spoken with do most of their IPC via ad hoc HTTP calls. By including a codec facility in the API, those ad hoc HTTP calls propagate the TraceContext (or SpanContext, here) from process to process *without relying on knowledge of the particular OpenTracing implementation*. Breaking that is a non-starter.

opentracing/src/main/java/opentracing/Span.java
michaelsembwever: > Is the returned childContext immutable in this design?

No.
For example, using the two scenarios in https://github.com/opentracing/opentracing-java/pull/7#discussion_r50500237

1) 

    SpanContext spanContext = traceContext.newRoot();  // or traceContext.spanContext();
    spanContext.setAttribute("foo", "bar");
    Tracer.startTrace("new-span-name", spanContext);

2) 

    childContext = currentSpan.childContext();
    childContext.setAttribute("foo", "bar");
    Tracer.joinTrace("new-span-name", childContext);

I do think once we settle on on SpanContexts are used, we can improve how we define Tracer.

opentracing/src/main/java/opentracing/TraceContext.java
```diff
@@ -14,53 +14,22 @@
...
- * processes.
- *
- * <p>TraceContext also support a simple string map of "trace attributes". These trace attributes
- * are special in that they are propagated *in-band*, presumably alongside application data. See the
- * documentation for {@link #setAttribute(String, String)} for more details and some important
- * caveats.
+ * Long-lived interface that knows how to create a root {@link SpanContext} and encode/decode
+ * any other.
  */
 public interface TraceContext {
```
bensigelman: this is not a context object (anymore), and it's not concerned with a specific trace (given its long-lived-ness)... as such I am puzzled about this name...

opentracing/src/main/java/opentracing/Tracer.java
bensigelman: I am fine with the name SpanContext. I am also fine with the name TraceContext. :)

No strong feelings, though perhaps a slight leaning towards TraceContext since it's a term that's used in multiple tracing systems already. From first principles, SpanContext seems more accurate as a descriptor, though.

opentracing/src/main/java/opentracing/Span.java
michaelsembwever: >  From various comments on this thread, including the one quoted above, it seems your SpanContext is not meant to perform either of these functions.

No. SpanContext solves both those functions.
Can you elaborate which comments allude otherwise?

opentracing/src/main/java/opentracing/SpanContext.java
bensigelman: Yes, the point of this API is that an OpenTracing user can set an attribute high in the stack of their (distributed) system then retrieve it at the bottom of the stack since it's sent in-band. There are numerous use cases for this which I can describe if that's helpful. But removing the getter makes the API pointless.

opentracing/src/main/java/opentracing/SpanContext.java
bensigelman: Tracing implementations have no control over how HTTP middleware works; that's the problem that motivated this restriction.

For what it's worth, I don't really have a dog in this particular fight... But it's important to understand that OpenTracing *implementors* are not the ones who deal with wire protocols.

opentracing/src/main/java/opentracing/Span.java
bensigelman: @adriancole that's the idea... it is an arbitrary object that implementations can record in total, in part, or not at all.

opentracing/src/main/java/opentracing/TraceCodec.java
michaelsembwever: > By including a codec facility in the API, those ad hoc HTTP calls propagate the TraceContext (or SpanContext, here) from process to process without relying on knowledge of the particular OpenTracing implementation.

So what you're saying is the SPI must define what's required for serialisation of SpanContexts across different implementations to the one tracing platform?

Or are we expecting SpanContexts to go over the wire between different tracing platforms? Because that doesn't make sense to me.

opentracing/src/main/java/opentracing/TraceContext.java
michaelsembwever: I agree. And i'm unclear on why it's a different object to `Tracer`.

But I'd hoped to come to that later, in correspondence with  https://github.com/opentracing/opentracing-java/pull/7#discussion_r50501336

bensigelman:
@michaelsembwever thank you for sending this out. Per gitter, we are not on the same page.

It is apparent that either the goals of OpenTracing haven't been communicated well or you don't agree with them. :)  Especially given the comments that have come up in this PR, what you propose here is not an API simplification: it removes entire swaths of semantic functionality. If that's intended, call a spade a spade. If it's not, imagine writing (or actually write if you have the time!) example calling code that
- serializes a Span to and from the wire (c.f. your enthusiasm for removing the Codec),
- sets a trace attribute in a client process and reads it in a server/peer process (c.f. your enthusiasm for removing the attribute getter),
- creates a child Span from a parent Span when the Tracer is not accessible (c.f. your removal of `Span.startChild`),
- and creates a child Span that has a parent_id tag (c.f. your removal of `ChildSpanContext`)

... all without peeking through to the OpenTracing implementation.

As for naming (less important given the scope of the problems above):
- SpanContext is an ok (if unconventional w.r.t. existing systems) name in my eyes... no strong feelings on my part, though
- what you *now* call TraceContext (formerly `TraceContextSource`) is not about a single trace, nor is it a context object, so that name confuses me
- TraceCodec does not encode or decode traces, it encodes or decodes a slender set of ids that refer to a span and a trace... and for that reason seems like a naming regression as well.

opentracing/src/main/java/opentracing/SpanContext.java
michaelsembwever: > Yes, the point of this API is that an OpenTracing user can set an attribute high in the stack of their (distributed) system then retrieve it at the bottom of the stack since it's sent in-band.

Right. So it's expected that the use of `spanContext.getAttribute(key)` goes beyond that of just the tracing implementation. That is, the application code using the tracing client library also wants access to these attributes.

If that is the case then I do see a need in keeping the getter.

An example I can think of is in Zipkin how the trace_id is used in each service and put into the logging context, so that tools like Kibana and Zipkin re-use the same correlation id.

opentracing/src/main/java/opentracing/TraceCodec.java
adriancole: >
> By including a codec facility in the API, those ad hoc HTTP calls
> propagate the TraceContext (or SpanContext, here) from process to process
> without relying on knowledge of the particular OpenTracing implementation.
>
>
We are defining context as a split across exactly two buffers. If there is
indeed a lot of practice doing this, perhaps we can find maybe even 3
tracers doing this?

I'm sure I can talk at high level to someone and they'd agree that
propagating context implies serialization. Devil is in the details... Did
the parties you've spoken to nod that they want a list of 2 byte arrays?
This is the primary source of my concern with this issue. I'd rather leave
it unspecified that awkwardly not like most apis.

opentracing/src/main/java/opentracing/SpanContext.java
michaelsembwever: > But it's important to understand that OpenTracing implementors are not the ones who deal with wire protocols.

Can you elaborate on that sentence please. Who do you mean does then?

michaelsembwever:
> It is apparent that either the goals of OpenTracing haven't been communicated well …

Quite possibly. I am late to the game, but so will every subsequent person after me.
If there's documentation to which I obviously haven't read; do please reference them during the discussion.

> ... all without peeking through to the OpenTracing implementation.

There's a difference between "peeking through" the OpenTracing implementation, and using that OpenTracing implementation. I have presumed that the goal of OpenTracing is to provide a specification for different tracing implementations to implement. Not to provide universal documentation for all users of all different platforms.

> - what you *now* call TraceContext (formerly `TraceContextSource`) is not about a single trace, nor is it a context object, so that name confuses me

We're on the same page here, and that discussion continues in this PR.

The other bulleted points in your comment I'll reply to in context to the code and the previous ongoing discussions.

opentracing/src/main/java/opentracing/TraceCodec.java
michaelsembwever: From https://github.com/opentracing/opentracing-java/pull/7#issuecomment-173808490
> - TraceCodec does not encode or decode traces, it encodes or decodes a slender set of ids that refer to a span and a trace... and for that reason seems like a naming regression as well.

See the rationale for not calling it SpanContextCodec in the commit msg: https://github.com/opentracing/opentracing-java/commit/02d5d15bb51866afc7b8ed9d035798032c0b0848
I see your point, but by calling it the more intuitive SpanContextCodec it alludes to different SpanContexts within the one trace allowing different codec implementations.

Other suggestions?

opentracing/src/main/java/opentracing/SpanContext.java
michaelsembwever: From https://github.com/opentracing/opentracing-java/pull/7#issuecomment-173808490
> - sets a trace attribute in a client process and reads it in a server/peer process (c.f. your enthusiasm for removing the attribute getter),

This is a good example to use.
Removing the attribute getter does not make implementing this any more difficult, presuming that you mean by "reads it in a server/peer process" reading it within a tracing implementation.

opentracing/src/main/java/opentracing/TraceContextSource.java
michaelsembwever: From https://github.com/opentracing/opentracing-java/pull/7#issuecomment-173808490
> - and creates a child Span that has a parent_id tag (c.f. your removal of `ChildSpanContext`)


A child span is created (with the appropriate parent_id tag) using 

    Span span = Tracer.joinTrace("new-span-name", context);

opentracing/src/main/java/opentracing/Span.java
```diff
@@ -25,16 +25,7 @@
...
-  TraceContext traceContext();
-
-  /**
-   * Denotes the beginning of a subordinate unit of work.
-   *
-   * @param operationName name of the operation represened by the new span from the perspective of
-   * the current service.
-   * @return a new child Span in "started" state.
-   */
-  Span startChild(String operationName);
```
michaelsembwever: From https://github.com/opentracing/opentracing-java/pull/7#issuecomment-173808490
> - creates a child Span from a parent Span when the Tracer is not accessible (c.f. your removal of `Span.startChild`),

I do not understand how it is possible that a Tracer is not accessible, nor the point of creating a Span in any environment when no Tracer exists.

opentracing/src/main/java/opentracing/TraceCodec.java
michaelsembwever: From https://github.com/opentracing/opentracing-java/pull/7#issuecomment-173808490
> - serializes a Span to and from the wire (c.f. your enthusiasm for removing the Codec),

I'd envisioned that this type of information is specified by the Tracing platform, not by the OpenTracing specification.

bensigelman:
I'm torn between a point-by-point response to all of the various issues here and something higher-level. I'll go for the latter since it'll be less noisy for everyone involved...

I am thankful that you sent this out, as this review and some of the gitter traffic demonstrates (to me) that the current set of OpenTracing primitives apparently leads to mass confusion: surely a fault of the API/scheme, not the masses. :) Or maybe it's just missing documentation, but my instinct tells me that the problem is deeper than that.

I am going to try to send out a PR that basically gets rid of `SpanContext` as an API primitive... overall I think/hope it will make for an equally portable tracing API and less confusion for OpenTracing contributors present and future. In order to evaluate that I need to make sure I can adapt the (portable) OpenTracing client code I've already written with the revised/trimmed API... If I fail, then I'll come back and we can go through a lengthy micro-review (i.e., where this is headed at the moment).

Sound reasonable?
Ben

yurishkuro:
Thanks, Ben. In turn, I will post a PR with documentation of the Common Use Cases we've discussed when designing the API, with code examples of how each use case is solved by the current API.

michaelsembwever:
Sounds reasonable @bensigelman 
Outside of just API there's a number understandings I raised in https://github.com/opentracing/opentracing-java/pull/7#issuecomment-173814992
If any of these are wrong can you highlight them.

michaelsembwever:
@yurishkuro Can you push these as documentation into opentracing.github.io ?

This really would be invaluable information for newcomers.

yurishkuro:
@michaelsembwever already in PR  opentracing/opentracing.github.io#34

michaelsembwever:
@yurishkuro thanks for the examples. 

They all match my previous understandings of what we're trying to achieve. So far I haven't been able to find any reasons why the API simplifications offered in the file changes here can't be adopted.

The examples do add context to two separate discussions in this PR: the getter method in SpanContext and the codec class. I'll add my thoughts to these existing discussions.

opentracing/src/main/java/opentracing/SpanContext.java
michaelsembwever: bq. Right. So it's expected that the use of spanContext.getAttribute(key) goes beyond that of just the tracing implementation. That is, the application code using the tracing client library also wants access to these attributes.

This was confirmed in @yurishkuro work in https://github.com/opentracing/opentracing.github.io/pull/34/files#r50616901

opentracing/src/main/java/opentracing/TraceCodec.java
michaelsembwever: @yurishkuro does a good job of demonstrating what's the intended use of the codec in https://github.com/opentracing/opentracing.github.io/pull/34/files#r50616999

My misunderstanding to this has been based on my interpretation of what serialisation entails.

Serialisation is the writing of bytes or characters into streams/buffers for the sake of storage or cross-process communication.

What the API here is trying to do is transform state from one object (eg the http headers) into the tracing's SpanContext. This is a transformation process not serialisation. While http and rpc frameworks offer state structures that we'll need to transform into SpanContexts it's a leap to presume this transformation is always connected to serialisation, and worse to leak that connection to the API user.

An example of this transformation not being second-handedly connected to serialisation is when reading offline tracing data, for example from log files, where the data goes through a number of transformation/cleansing steps before being finally transformed into SpanContexts.

opentracing/src/main/java/opentracing/Span.java
michaelsembwever: @bensigelman None of the examples raised by @yurishkuro nor anything else i've seen demonstrates any need for the method `span.startChild(..)` in when processing spans postmortem from a different system.

michaelsembwever:
**Summarising** the status of the PR so far.

 - That TraceContext can be renamed to something better like SpanContext, and the renaming of Span.traceContext(), has been superseded by @bensigelman proposing to offer a patch that removes it altogether.
 - Removing the info/error methods in Span, has been agreed upon. 
 - Removing Span.createChild(..) and ChildSpanContext remains a valid proposal, as there's no examples raised (yet) of where they are needed.
 - Renaming TraceContextSource to TraceContext. This depends on Ben's proposed patch to remove the TraceContext.
 - Rename TraceContextCodec to TraceCodec. There's confusion here because of the terms codec and serialisation. This can be spun off into a separate PR.

**Actions** to take.

 - Remove the "Rename TraceContextCodec to TraceCodec" from this PR. I will spawn a new PR suggesting alternative names, like SpanExtractor.
 - Wait for (agreement upon) Ben's patch to remove the TraceContext. 
 - Evaluate naming of TraceContextSource (maybe Ben solves it).

**Other suggestions** that have been raised in this PR.

 - Tracer.startTrace(..) and tracer.jointrace(..) would be simpler as one method.
 - change `Span event(String message, Map<String,String> payload);` to `Span event(String message, @Nullable Object payload);` 
 - Change the documentation on setAttribute(..) so the regular expression of keys is only a recommendation. This has been agreed upon.

Is this clear? Does everyone agree?

bensigelman:
I sent out https://github.com/opentracing/opentracing-go/pull/40 (for better or worse, `!Java` since I wanted to compile against a working example implementation and a working caller that propagates across HTTP)...

@michaelsembwever if people like the direction of my RFC above, we can get rid of a lot of stuff here. And also save a lot of time and energy arguing about names that would no longer exist :)

PS: @michaelsembwever sorry I didn't respond to all of the various minor points above where you @bensigelman'd me -- I thought it would be wasted effort for both of us since so much of the debated material is presently in jeopardy :)

michaelsembwever:
That's fine @bensigelman , i'll be happy to see the work here come to fruition any way possible.

As listed in my previous comment everything that this PR proposes is eventuating one way or another, plus more, and that makes me happy.

I'd only hoped that you got many more of these types of reviews opened before you started considering anything vaguely resembling a 1.0 release.

opentracing/src/main/java/opentracing/TraceContextSource.java
adriancole: On parent_id particularly

in zipkin parent_id is simultaneously a tag (parent_id -> Span.parentId) and a propagated one (parent_id -> X-B3-ParentSpanId). The only propagation tag which isn't also a span tag is the sampled header.

That's because zipkin currently doesn't take on the abstract task of propagating data for use outside trace implementations. For example, right now that functionality rests in tools like Finagle's context propagation api.

opentracing/src/main/java/opentracing/SpanContext.java
adriancole: the getAttribute/propagatedTag part here is not a feature to allow things like logging integration to work. It is a feature that requires things like logging integration to work *without looking under the covers*.

For example, zipkin formerly did not specify a specific means to propagate context, yet most tracers have in fact managed to do things like MDC tagging of trace id. How is that? They do because they reach under to the propagation context and take a peek.

opentracing/src/main/java/opentracing/TraceContextSource.java
michaelsembwever: So ChildTraceContext is needed to pass down the parent_id tag, which can't be treated like a propagated tag (or attribute).

My point here is that the application code won't ever interact with this ChildTraceContext.
Therefore it does not belong in the public SPI (nor the public API for that matter).

See my explanation of components to the API in https://github.com/opentracing/opentracing-java/pull/7#issuecomment-173814992

adriancole: >
> My point here is that the application code won't ever interact with this
> ChildTraceContext.
> Therefore it does not belong in the public SPI (nor the public API for
> that matter).
>
> See my explanation of components to the API in #7 (comment)
> <https://github.com/opentracing/opentracing-java/pull/7#issuecomment-173814992>
>
agreed, at least in spring-cloud-sleuth, there's no desire to require users
to understand the details of how a child context, or any other span for
that matter, is constructed.

michaelsembwever:
Closing this PR, the first step in a simpler API it superseded with https://github.com/opentracing/opentracing-java/pull/8

opentracing/src/main/java/opentracing/Span.java
michaelsembwever: see https://github.com/opentracing/opentracing-java/pull/9

PR #8, opened 2016-01-31
Simplify the API: removing TraceContext and TraceContextSource

michaelsembwever:
- Remove TraceContext, as has already been done in the specification.
- Remove TraceContextSource, as methods from it now belong in Tracer.
- Remove Tracer.joinTrace(..) as Span.startChild(..) already does this.
- Simplify the SpanPropagator. Having to split responsibilities for span and trace identification versus trace attributes is not something the end-user cares about.

ref:
 - https://github.com/opentracing/opentracing.github.io/pull/42
 - https://github.com/opentracing/opentracing-java/pull/


michaelsembwever:
@adriancole @yurishkuro @bensigelman

opentracing/src/main/java/opentracing/SpanPropagator.java
```diff
@@ -0,0 +1,53 @@
...
+ * binary message format, so that it can be inspected efficiently by the middleware / routing layers
+ * without parsing the whole message.
+ */
+public interface SpanPropagator {
+
+  /**
+   * Encodes the span into a binary representation of the span's identity and trace
+   * attributes.
+   */
+  byte[] toBinary(Span span);
```
yurishkuro: Merging trace attributes with span identity makes it much more expensive for routing layers (think haproxy, but for binary protocols) to add extra data to the spans. The routing now needs to unpack the whole Trace Attributes blob that it's not interested in and has no intention of changing just to access and possibly change the span ID.

opentracing/src/main/java/opentracing/SpanPropagator.java
```diff
@@ -0,0 +1,53 @@
...
+
+  /**
+   * Decodes a span from a binary representation of the span.
+   *
+   * @throws IllegalArgumentException if the encoded data is malformed.
+   */
+  Span fromBinary(byte[] encoded);
```
yurishkuro: in Go version, calling the method joinTraceFromBinary emphasizes the fact that a span is returned in a "started" state, and the user is expected to call `span.finish()` when done with it.

opentracing/src/main/java/opentracing/SpanPropagator.java
adriancole: Mick, this is not really interesting in zipkin as it doesn't propagate variable length data (or at least the variable length is relatively small ex few hundred bytes)

However, OT does (user attributes). That's the big difference. If an implementation chooses to support user attributes, and they are added.. the transport can toss them by ignoring the blob that contains them.

If tracer attributes (parent id, sampled flag) are combined in the same blob as user attributes (who knows!), then there's not flexibility in if or how to send the variable length data.

adriancole: I should clarify slightly.. Zipkin backends don't require or have an opinion on extra propagated things. Individual tracers may or may not.

Traditionally, zipkin tracers have not intentionally exposed their DCP apis to users directly. They do use common header names for interop, but for example don't specificy a naming convention or otherwise to carry other data.

adriancole:
Nuance about framing user attributes in a different buffer than tracer,
looks like progress.

opentracing/src/main/java/opentracing/SpanPropagator.java
michaelsembwever: Let me clarify.

I'm not removing the separation between span identity and trace attributes here, but simply removing it from the end-user.

I'm aware of the notion of the separation and its benefits.

But the end-user really doesn't care about how the byte[] or the map of strings is split up, that's up to the Specification to tell the implementation. Hence it has been removed from here. My understand is that the encoding specifics and any routing layer is not an end-user.

opentracing/src/main/java/opentracing/SpanPropagator.java
michaelsembwever: When would `from Binary(..)` return a finished span?

opentracing/src/main/java/opentracing/SpanPropagator.java
```diff
@@ -0,0 +1,53 @@
...
+/**
+ * Encodes or Decodes a {@linkplain Span} in binary or text formats.
+ *
+ * <p>The toXXX methods are expected to serialize spans into a pair of values representing
+ * separately the span / span identity, and the trace attributes. This is done specifically
+ * for binary protocols that may represent tracing identity in a dedicated fixed-length slot in the
+ * binary message format, so that it can be inspected efficiently by the middleware / routing layers
+ * without parsing the whole message.
+ */
+public interface SpanPropagator {
```
michaelsembwever: While there's simplifications in this class that can make the API a lot easier to use, I'm not comfortable with the design of SpanPropagator.

The pushing of `byte[]` and `Map<String,String>` seems too presumptuous to me. How do we know if this is what all protocol middleware frameworks will work with? I do believe this needs to be a more SPI interface to the concept of a middleware framework, and not a presumption that we will always be able to reach into its guts and get a hold of the wire's byte or character stream.

I raised this concern before, and implemented stuff like this before. My suggestion is to keep brainstorming on this problem. The Apache Request framework did something similar here so to adapt to different http middleware frameworks: Servlet, Portlet, JSP, Mustache, Velocity and FreeMarker, I'm aware that its tackling a problem in a different domain but it still draws strong enough parallels.

The Apache Request framework does it like with the [AttributeExtractor](http://tiles.apache.org/tiles-request/apidocs/org/apache/tiles/request/attribute/AttributeExtractor.html) class. By letting the unknown be the subclass instead of the parameter the api design gets to focus on what it does know; **each protocol middleware framework that we support must provide a way (a subclass) to extract the span identity and trace attributes out for us.**

opentracing/src/main/java/opentracing/SpanPropagator.java
adriancole: > But the end-user really doesn't care about how the byte[] or the map of
strings is split up, that's up to the Specification to tell the
implementation. Hence it has been removed from here. My understand is that
the encoding specifics and any routing layer is not an end-user.

I think I've got you! Yeah, so now everything is in the span, including
attributes. The user shouldn't be bothered with some notion of dual
buffers, since they really only get a single item back.

What you are saying is to really remove framing or encoding details from
the user api. For example, underneath this, a tracer could know a binary
format, Ex serialized protobuf message, and that could have been decoded
skipping attributes or whatever. In other words these details are 100% impl
specific.

Do I understand?

opentracing/src/main/java/opentracing/SpanPropagator.java
michaelsembwever: For example

    public interface SpanExtractor {
        Span getSpan();
    }

Again the notion of separation between span identity and trace attributes is just not needed in the end-user API.

If the implementation wants to provide optimisations at the wire level (something that should not be polluting end-user API) then this can be done inside the implementation of the Span that's returned from the SpanExtractor.

At the end of the day it's neither the middleware framework nor the OT implementation that determines where the barrier between span identity and trace attributes lies, but the wire protocol itself.

I'm also not sure what type of optimisation such a routing layer could provide anyway, will it know the length of bytes/characters in the trace attribute section and be able to skip over it? Will it have access to this level in the protocol middleware to do such skipping? Or will the common implementation be in fact still to read through these bytes/characters but just not put them back into wire? If it's the latter it's only a very small amount of cpu optimisation that's relevant here so sounds like a case of pre-mature optimisation. 

For the situation of adding trace attributes at a proxy routing layer it all sounds a little flakey to be playing around with bytes in this manner; the presumption that bytes can simply be tacked on to the end of an existing byte stream without having read it. And it's obvious you can't do that with Map<String,String>.

opentracing/src/main/java/opentracing/SpanPropagator.java
michaelsembwever: Yes!

yurishkuro: who is the "user" in this case?  In most cases it's a person implementing an RPC framework, who needs to take span and encode it into the RPC-specific format that may require Trace Attributes to be separate from trace ID.

opentracing/src/main/java/opentracing/SpanPropagator.java
yurishkuro: @michaelsembwever I think we have this as unresolved issue https://github.com/opentracing/opentracing.github.io/issues/22.  Basically, there are layers that may want to add stuff to a span, but not affect the span's timestamp/duration, which is implied by the `finish` method. But at this point I think we indeed expect `fromBinary` to always return a started span that must be finished (as that's the only way to cause it to be reported).

But there's a second problem here. The former `join` method was abstracting away the difference in how tracing systems model RPCs, Dapper with two spans, Zipkin with one span. The tracer was deciding whether `joinFromBinary` would return the same spanID (Zipkin) or a child span ID (Dapper), without forcing the end user to make that choice. Expecting the same semantics from `fromBinary

opentracing/src/main/java/opentracing/SpanPropagator.java
bensigelman: There are two reasons to split core "tracer state" from the "trace attributes":

1. allowing OT clients (whether they be application owners or maintainers of IPC frameworks) to opt-out of the "trace attribute" wire cost, regardless of the OT implementation, and
2. enabling the sorts of introspection that @yurishkuro has described many times now (no need to reiterate it here)

@michaelsembwever you say:

> I'm aware of the notion of the separation and its benefits.

... and you go on to say:

> My understanding is that the encoding specifics and any routing layer is not an end-user.

This is the crux of the problem. You don't think of RPC framework authors as users of OpenTracing, yet they are. I don't know how to convince you of this since it seems self-evident to me: they must use the OT APIs to get bits over the wire, and it is a non-starter to try and add custom methods to `SpanPropagator` for every popular IPC framework and/or middleware layer.

I think you are asserting (or at least implying) that the only "user" that matters is the application programmer who doesn't write RPC systems. If we pursue that path, I don't know how OpenTracing will be able to provide tracing interop for at the RPC/IPC layer, at least when trace attributes need to be treated separately.

opentracing/src/main/java/opentracing/TraceContextCodec.java
```diff
@@ -1,71 +0,0 @@
...
-/**
- * Encodes or Decodes a {@linkplain TraceContext trace context} in binary or text formats.
- *
- * <p>The toXXX methods are expected to serialize trace contexts into a pair of values representing
- * separately the trace context / span identity, and the trace attributes. This is done specifically
- * for binary protocols that may represent tracing identity in a dedicated fixed-length slot in the
- * binary message format, so that it can be inspected efficiently by the middleware / routing layers
- * without parsing the whole message.
- */
-public interface TraceContextCodec {
```
bensigelman: yay red code!

opentracing/src/main/java/opentracing/Tracer.java
```diff
@@ -33,22 +33,5 @@
...
-   *
-   * <p>Note that the following calls are equivalent
-   * <pre>{@code
-   * Span feed = tracer.startSpanWithContext("GetFeed", traceContextSource.newRoot());
-   * ...
-   * Span feed = tracer.startTrace("GetFeed");
-   * }</pre>
-   */
-  Span startSpanWithContext(String operationName, TraceContext context);
+  SpanPropagator getSpanPropagator();
```
bensigelman: @dkuebric and I were chatting about this name... how about just `Propagator getPropagator()`? Just a thought.

bensigelman:
@michaelsembwever the removal of `TraceContext` per se looks very red and good. Thank you for that.

I don't know if you'll be willing to live with this, but it's worth asking: how about trying to get the Java API into parity with the latest Go stuff and splitting the trace attributes out, then moving the debate we're starting in this PR to somewhere appropriate under https://github.com/opentracing/opentracing.github.io/issues ?  I am happy to delay any "OpenTracing 1.0RC" stamp until we've come to some point of resolution about said debate, but I wish those sorts of discussions which are not language-specific could happen in a central place. If/when we decide to make a change, we can then change the various platform APIs and the `spec.md` (etc) all at once.

opentracing/src/main/java/opentracing/Tracer.java
michaelsembwever: +1

But keep in mind my questioning of the design of Propagator, and suggestion of an Extractor design instead.

opentracing/src/main/java/opentracing/SpanPropagator.java
michaelsembwever: > This is the crux of the problem. You don't think of RPC framework authors as users of OpenTracing, yet they are.

Yes this has been an on-going misunderstanding, and i've tried a number of times to bring about better nomenclature into the discussion to help progress it. 

1. The end-user is the application developer. The api they use is the public API/SPI.
2. The authors of OT implementations. The api they use is the internal API, and do so to flesh out the public SPIs.

End-users read the Specification to gain an understanding on how to use the public API.
Authors read the Specification to know what their implementation must adhere to.

The java interfaces being introduced here are proposed as only the public API.

The development, evolution, and stability of the public and internal APIs is different, so not only will using clear nomenclature between the two help the discussions, but help us not shoot ourselves in the foot with API compatibilities issues over time. And lastly, by separating them we focus on building the simplest and most attractive end-user API possible.

I am definitely not saying that the author does not matter, but rather giving them their own arena.

michaelsembwever: And the easiest way forward is to

first: define and agree upon what's the simplest public API,
second: from there build upon it the simplest internal API.

opentracing/src/main/java/opentracing/SpanPropagator.java
michaelsembwever: `joinFromBinary` didn't exist in the interfaces branch this PR is based off.

This discussion can be moved to the more appropriate place (any PR attempting to merge the interfaces branch into master).

michaelsembwever:
@bensigelman As I mentioned above in https://github.com/opentracing/opentracing-java/pull/8#discussion_r51369125 this PR is only trying to tackle the end-user public API side of things. And it's based off the interfaces branch, not the master branch.

opentracing/src/main/java/opentracing/SpanPropagator.java
bensigelman: Really glad we are discussing this. It's critical to understand that the author or maintainer of an IPC subsystem (eg, thrift, TChannel, some arbitrary ad hoc http client, etc) is a **consumer** of the public API here. They are the end user.

The only implementors of this public OpenTracing API are the tracing systems themselves (eg, Zipkin). That is, thrift is not an OpenTracing implementation.

(And I do agree that the *author* of OpenTracing implementations is out lowest priority from a convenience standpoint... Their job should be possible but need not be easy.)

Sorry if this is rambly, I'm typing on my phone :)

michaelsembwever: Ok. Then we have different notions as to who will implement the IPC components.

I've always been under the impression that this is primarily the concerns of the implementors. For example Zipkin will provide OpenTracing instrumentation libraries to work with http and with thrift and what else.

I still struggle to see why the application developer (even those at the lower level that deal with the IPC code) need to be playing around with the internals of the SpanPropagator, and in the situation where they are why we would not expect them to be doing so using the internal API.

I'm trying to get my head around it, but as far as I can see, adding attributes or bytes to the wire within the SpanPropagator is beyond the concerns of instrumentation: beyond the concerns of the end-user; and moving into more nuts and bolts and hence more the internal API. I wouldn't be surprised either if this type of development is stuff: for example proxy routing where trace attributes are added; we'd be hoping to see contributed back to the particular OpenTracing implementation.

michaelsembwever: It might also be more productive if we flesh out the concerns around adding trace attributes against a IPC subsystem, and what-not, against the `SpanExtractor` proposal I raised above. I suspect going down that path with provide us that cleaner separation in the API between the high-level instrumentation stuff (what i've been referring to as end-user) and the lower level stuff (what i've been referring to as internal).

yurishkuro: > For example Zipkin will provide OpenTracing instrumentation libraries to work with http and with thrift and what else.

This seems hardly practical given how many different RPC frameworks may be out there. Why should it be the duty of OepnTracing implementor and not of the implementor of the actual RPC framework? Even if we take http, there are hundreds of different server and client frameworks built on top of HTTP, all with their own APIs to prepare an outbound request and set its headers. The only commonality among them from OT point of view is that they deal with a text protocol, so OT provides a way to encode span into virtual "headers" as string key/value pairs. Then the actual framework decides how it can take that map of key/value pairs and translate it into the format of HTTP request it operates on.

Easier to show in the code: https://github.com/uber-common/opentracing-python-instrumentation/blob/master/opentracing_instrumentation/http_client.py#L59

This is an instrumentation layer that tries to be "generic" by relying on a certain abstraction of an http request. The actual http libraries all have different APIs for http request class, so adapters are provided to be able to reuse this "generic" wrapper around outbound request. This code is not an OpenTracing API, it's a user of that API.

michaelsembwever: > The only commonality among them from OT point of view is that they deal with a text protocol, so OT provides a way to encode span into virtual "headers" as string key/value pairs.

This, and the byte stream, and the decision as to which to use just doesn't gel with me.

You've taken a Span, then transformed it to a k/v, then handled to the RPC framework with the presumption the map is the right thing for it to work with. That's presumptuous and creating an API that doesn't evolve well. 

On the other hand if you take the SpanExtractor and work with the problem of headers coming out (same thing really)

    public interface SpanExtractor {
        Span getSpan();
    }
Then add on to that the use-case for the commonality you expect you get (something like)

    public class SpanFromMapExtractor {
        public SpanFromMapExtractor(Map<String,String> spanIds, Map<String,String> headers) { ..}
        public Span getSpan() {
            return new Span(spanIds.get("trace_id"), spanIds.get("span_id"), headers);
    }
You get the picture. Here you haven't forced the presumption of the text or byte array or map up to the top level of the api but kept it within suggested helper classes. From here the API can evolve to offer SpanFromServletExtractor, SpanFromThriftExtractor, SpanFromSpringExtractor, etc. And from these classes you get to start offering more the nuts&bolts stuff like optimisations to skip reading headers while extracting, without polluting the top-level design. These helper classes are easy to depose of as well over time, or move into their corresponding frameworks even. End-user developers also have full access to this functionality, again without complicating the initial API we expose them to.
> The actual http libraries all have different APIs for http request class…

And that too can be better designed for, as pointed out in the SpanExtractor example.

> opentracing-python-instrumentation/opentracing_instrumentation/http_client.py#L59

Yes that is end-user code, and I can see a better design there possible that steers away from byte streams and k/v maps, and other presumptions about what RPC frameworks offer you.

yurishkuro: We chose `Map[String, String]` as the lowest common denominator, because the tracing implementation does not know how the actual RPC framework is going to serialize that map on the wire. If you have a better API that can be used in the `http_client.py` I included above, I would love to see it. I am not clear how `SpanFromMapExtractor` could be used, because it is implementation-specific, since it has knowledge of what keys can be found in the map, like `trace_id` and `span_id` (there's no requirement in OT as to what keys will be in the map once the span is converted to it). If you're saying that `SpanFromMapExtractor` is just an interface that RPC framework can somehow retrieve from the tracer, then I don't see how it's different from the tracer directly exposing `toMap` and `fromMap` methods that we have today (the only difference is composition, but I think you're objecting to the notion of map, not where the methods are defined).

It is to avoid these confusions that we published the [use cases](http://opentracing.io/use-cases/) on the site, which show concrete examples of end-user code using OT API. Can you try to rewrite the following part from the Tracing Client Calls use case using a different API?

```python
        # Goal: framework encodes the Span into HTTP headers of the outbound request
        h_ctx, h_attr = tracer.propagate_span_as_text(span)

        for key, value in h_ctx.iteritems():
            request.add_header(key, value)
        if h_attr:
            for key, value in h_attr.iteritems():
                request.add_header(key, value)
```

michaelsembwever: > the only difference is composition, but I think you're objecting to the notion of map, not where the methods are defined …

This brings us closer together, as it is where the methods are defined (the design of the api) that i'm objecting to. By moving the concept of the k/v down into a helper subclass, you're providing a simpler top-level design, and a design that gives more horizon.

I hear you in regards to use-cases and concrete examples, I'll come back with what I can on that front.

bensigelman: (One other idea now that I'm not typing on my :phone:!)

We could alternatively define an iterator interface for propagation... if done carefully, this would make it possible to distinguish between named and positional parameters, "Tracer Fields" vs the user-defined KV pairs, etc. Something very roughly like this:

```
public interface Propagator {
    PropagationFieldIterator propagate(Span span);

    ....
}

public interface PropagationFieldIterator {
    boolean done();
    PropagationFieldIterator next();

    FieldType fieldType();

    // Returns -1 when not defined... but trace_id and span_id could be `0` and `1` respectively, for example.
    int fieldPosition();

    // Always defined, but not necessarily used when `position() >= 0`:
    String fieldKey();

    // Always defined.
    String fieldValue();

    public enum FieldType {
        TRACER_IMPL,
        ANNOTATION
    };
}
```

^^^ please excuse the terrible Java style... just trying to illustrate the concept, I realize the naming is terrible, etc.


Anyway, I kind of like that idea but think the other direction (deserialization) would be messy. But wanted to throw it out there.

Definitely second @yurishkuro request for a diff on the relevant use-case example.

michaelsembwever: Keep in mind that the only reason the separation exists is
 **if** the middleware exposes a binary stream and is using binary serialisation
 **if** the binary protocol uses fixed-length slots,
 **if** the IPC interaction wants to skip trace attributes
 **for** a small performance benefit or stability fix.

> This is done specifically
> for binary protocols that may represent tracing identity in a dedicated fixed-length slot in the
> binary message format, so that it can be inspected efficiently by the middleware / routing layers
> without parsing the whole message.


Surely, you can understand my questioning as to whether this really needs to belong in OpenTracing API v1.0

I'm trying to keep horizons as open as possible to cater for the adoption of as many tracing frameworks as possible. But it seems we're getting blind-sighted by niche performance optimisations for an additional DCP add-on… 

I'm in favour of clear documentation and Specification, and a simple API.

Using the extractor pattern it'ld be something along the lines of

         # Goal: framework encodes the Span into HTTP headers of the outbound request
        tracer.propagate(span, request)

michaelsembwever: I've restored SpanPropagator to how it was so that this PR can get merged.

This discussion, which i'm feeling quite strongly for, can continue elsewhere.

michaelsembwever:
> how about trying to get the Java API into parity with the latest Go stuff and splitting the trace attributes out, then moving the debate we're starting in this PR to somewhere appropriate under https://github.com/opentracing/opentracing.github.io/issues ?

done.

bensigelman:
LGTM, thanks!

opentracing/src/main/java/opentracing/SpanPropagator.java
bensigelman: thank you for splitting it off -- happy to continue the discussion and ACK that it's not resolved.

yurishkuro:
lgtm

michaelsembwever:
I will merge this in 24hrs if no objections are raised.

opentracing/src/main/java/opentracing/TraceContext.java
```diff
@@ -1,66 +0,0 @@
...
- *
- * <p>TraceContexts are sufficient to propagate the, well, *context* of a particular trace between
- * processes.
- *
- * <p>TraceContext also support a simple string map of "trace attributes". These trace attributes
- * are special in that they are propagated *in-band*, presumably alongside application data. See the
- * documentation for {@link #setAttribute(String, String)} for more details and some important
- * caveats.
- */
-public interface TraceContext {
```
adriancole: can't wait till we have a general DCP decoupled from Tracing, so can stop dancing around where we put things like this.

adriancole:
@michaelsembwever feel free to merge. I'm going to abstain from further comments on this topic.

PR #9, opened 2016-01-31
Simplify the API: logging and timestamped annotations

michaelsembwever:
Simplify the logging and timestamped annotations down to a `event(message, payload)` method.

Rationale:
 This is the simplest solution, and an appropriate plumbing API, that solves the use-case of
info/error and parameterised message formatting.
 OpenTracing API may adopt a porcelain API for those info/error logging-style methods down the road once the plumbing api has been battle tested a bit more.

(cherry picked from commit 5630ab2389e054f9a353b35642acb87eac5cd00c)

### DIFF
--- a/opentracing/src/main/java/opentracing/Span.java
+++ b/opentracing/src/main/java/opentracing/Span.java
@@ -13,8 +13,6 @@
  */
 package opentracing;
 
-import java.util.Formatter;
-
 /**
  * Span represents an active, un-finished span in the opentracing system.
  *
@@ -61,30 +59,24 @@ public interface Span {
 
   Span setTraceAttribute(String key, String value);
 
+  /**   *
+   * Add a new log event to the Span, accepting an event name string and an optional structured payload argument.
+   * If specified, the payload argument may be of any type and arbitrary size,
+   * though implementations are not required to retain all payload arguments
+   * (or even all parts of all payload arguments).
+   *
+   * The timestamp of this log event is the current time.
+   **/
+  Span log(String eventName, /* @Nullable */ Object payload);
+
   /**
-   * {@code message} is a format string and can refer to fields in the payload by path, like so:
+   * Add a new log event to the Span, accepting an event name string and an optional structured payload argument.
+   * If specified, the payload argument may be of any type and arbitrary size,
+   * though implementations are not required to retain all payload arguments
+   * (or even all parts of all payload arguments).
    *
-   * <pre>{@code
-   *
-   * span.info("first transaction is worth ${transactions[0].amount} ${transactions[0].currency}",
-   *     ImmutableMap.of(
-   *       "transactions", asList(
-   *         Transaction.builder().amount(10).currency("USD").build(),
-   *         Transaction.builder().amount(11).currency("USD").build(),
-   *       )
-   *     )
-   * );
-   * }</pre>
-   *
-   * @param message {@link Formatter format string} that can refer to fields in the args payload.
-   * @param args arbitrary payload
-   */
-  // See https://github.com/opentracing/opentracing.github.io/issues/30 about parameterization
-  Span info(String message, Object... args);
-
-  /** Same as {@link #info}, but for warnings. */
-  Span warning(String message, Object... args);
-
-  /** Same as {@link #info}, but for errors. */
-  Span error(String message, Object... args);
+   * The timestamp is specified manually here to represent a past log event.
+   * The timestamp in microseconds in UTC time.
+   **/
+  Span log(long instantMicroseconds, String eventName, /* @Nullable */ Object payload);
 }


michaelsembwever:
@adriancole @bensigelman @yurishkuro 

as agreed in https://github.com/opentracing/opentracing.github.io/issues/30#issuecomment-173766654

opentracing/src/main/java/opentracing/Span.java
```diff
@@ -65,29 +63,7 @@
...
-  Span info(String message, Object... args);
-
-  /** Same as {@link #info}, but for warnings. */
-  Span warning(String message, Object... args);
-
-  /** Same as {@link #info}, but for errors. */
-  Span error(String message, Object... args);
+   * @todo
+   **/
+  Span event(String message, /* @Nullable */ Object payload);
```
yurishkuro: I think we should be consistent with Go and have

```java
logEvent(String event)
logEvent(String event, Object payload)
log(?? timestamp, String event, Object payload)
```

Although I am not against using a simpler `event` name for the method, as long as we do it in all languages.

adriancole: logEvent is intuitive if we consider the implicit expectation that all
things logged result in the same struct. Ex I think somewhere else in OT,
theres a log struct that has an optional field for event (in anticipation
that there will be later features which don't record event, I suppose)

We tried logEvent in Sleuth, and it hasn't raised alarm, but of course it
hasn't been released either! we chose not to add a parameter for payload,
as we have no reason to support that.

Interestingly, if you look at log4j 2, there's a logEvent op with ties in
with its propagation api. Ex people can put tags into a context used for
later formatting of the event.

https://logging.apache.org/log4j/2.0/manual/eventlogging.html

On timestamp, I would guess use long as the first arg and either specify
micros or say 'implementation decides'. Ex in log4j 2, their timestamped
message interface doesn't indicate precision, yet uses a long vs a
polarizing choice of Date types.

michaelsembwever: I like `event` as it's short and sweet, 

Shall we use separate methods or accept nulls?

    Span event(String message);
    Span event(String message, Object payload);
    Span event(String message, ZonedDateTime timestamp);
    Span event(String message, Object payload, ZonedDateTime timestamp);
OR

    Span event(String message, 
                         /*@Nullable*/ Object payload, 
                         /*Nullable*/ ZonedDateTime timestamp);
I would normally be in favour of separate methods, but for a porcelain and premature api i think i'm in favour of having just one method.

adriancole: ZonedDateTime enters the building ;) this will be polarizing to some,
particularly folks who use their own date representations (ex twitter
scala) and explicitly makes this a java 8+ api. Not sure what Android
version has this api, but maybe we can check.

That said, there's no reason why folks who can't use java 8 types can't
write their own interfaces. Iirc, in grpc-java, Android and netty use
different interfaces for the same job.

I've no problem with choosing this type, and setting minimum language level
accordingly. Main reason is that this api requires significant tracer
surgery anyway.

adriancole: No strong opinion on event vs logEvent

I can see value in overloading mainly due to isolate the payload arg, which
I think is most likely to break over time.

adriancole: Prefer timestamp at position zero or otherwise before payload as I've seen
folks elude to a varargs. Also, I know timestamp has an immediate use case
in zipkin.

Id like to be able to write an impl that completely ignores payload and
putting nulls in the middle looks worse. Ex event(CLIENT_SEND, null,
timestamp)

yurishkuro: to clarify, the long form method with 3 args is called `log`, not `event`, because it is a catch-all / last resort method. It's semantic signature is
```java
Span log(long timestamp, String event,  /*@Nullable*/ Object payload);
```
both timestamp and event are required. It specifically exists for people who want to provide their own timestamp (as in historical recording).

[Instant](https://docs.oracle.com/javase/8/docs/api/java/time/Instant.html) is the common way of representing timestamps. I think java-7 has a way to backport those via jsr310

I agree with Adrian that using `long` for the timestamp is the simplest way to avoid getting into the mess of Java time interfaces,

bensigelman: I am too out-of-the-loop in Java to justify strong feelings about naming/spelling. That said, it is important that we have a version of this method that supports an explicit timestamp (for historical recording, a la https://github.com/opentracing/opentracing.github.io/issues/22).

FWIW, I personally prefer `log` as the catch-all "plumbing" name and `logEvent` for the slightly easier-to-use event-oriented API. I'm also fine only-supporting `log` for now. Per https://github.com/opentracing/opentracing.github.io/issues/30, the idea is that we'll come back and deal with logging porcelain once the more important (and far more subtle) propagation APIs are settled.

michaelsembwever: Ok.

>  I'm also fine only-supporting log for now.

I'll change all the methods to `log`. (Having different names only adds schema and unnecessary overhead to the newcomer).

> Prefer timestamp at position zero

I'll put the timestamp first.

> putting nulls in the middle looks worse

I don't like using nulls either, but that's porcelain api we can easily add later on.

> Instant is the common way of representing timestamps.

I was thinking we will have services running in different timezones and will need to deal with that, but Instant forces us to use UTC time. I like that.

michaelsembwever: Also replace the @Nullable with Optional<Object>

bensigelman:
With the changes discussed above, this PR looks good to me, thanks!

michaelsembwever:
I'll merge this tomorrow if no objections are raised.

yurishkuro:
I would prefer an extra method `log(String message)` instead of forcing users into passing null or None, since the language allows overloading.

Otherwise lgtm.

adriancole:
I'd prefer to use an override without payload declaring Optional payload. In other words, if `log/event(String)` and `log/event(String, Object)` were there vs the Optional stuff, I'd probably say :shipit: even if I'm totally not in the payload camp at all.

---

The Optional type does exist in java 8+, however functional composition layers have their own option monads. I'd really prefer to dodge the topic completely, but not using anyone's Option.

I have similar thoughts about Instant, to be honest, but since recording events that already occured is already special-case, feel less about it.

michaelsembwever:
My thoughts about the use of Option, and just one method, is that this is the simplest method to start with.

With API evolution it's easy to add the second method later on without breaking any compatibility.

That way we are making the simplest smallest steps required. And we more adopters many of these opinions about what other methods would be "nice" become more heard across a larger group.

Sticking with one method though would mean that the log with a payload method would always use Optional.

Further thoughts?

michaelsembwever:
Coming back to this. 

> The Optional type does exist in java 8+, however functional composition layers have their own option monads. I'd really prefer to dodge the topic completely, but not using anyone's Option.

@adriancole surely the Java8 Optional is permissible here (being the *default* Optional class pe Java8), at least for the initial OT release?

my concern remains that while overriding methods is nice it presumes that the methods are correct, which we cannot for the first release. and that it's smarter for this release to add the fewest possible methods, even if it means we're left over time with an Java8 Optional in two of the methods.

adriancole:
I'd prefer nullable to anyone's optional, but I don't feel strongly enough to debate this. I'd rather have someone later complain than no-one complain because it never finished.

IOTW go for it.

michaelsembwever:
It's come up that OpenTracing will need to support Android.
For that reason I'll back to code down to Java7.

opentracing/src/main/java/opentracing/Span.java
```diff
@@ -62,29 +60,12 @@
...
-   * @param args arbitrary payload
-   */
-  // See https://github.com/opentracing/opentracing.github.io/issues/30 about parameterization
-  Span info(String message, Object... args);
-
-  /** Same as {@link #info}, but for warnings. */
-  Span warning(String message, Object... args);
+   * Add a timestamped annotation, with an optional payload.
+   **/
+  Span log(String message, /* @Nullable */ Object payload);
```
bensigelman: per https://github.com/opentracing/opentracing.github.io/issues/30#issuecomment-188464752, this signature may not be enough to disambiguate against a possible debug-logging signature in the near future... obvious suggestion would be to call this `logEvent(String eventName, /* @Nullable */ Object payload)` and reserve `log(...)` for the fallback catch-all method below.

opentracing/src/main/java/opentracing/Span.java
```diff
@@ -62,29 +60,12 @@
...
 
-  /** Same as {@link #info}, but for errors. */
-  Span error(String message, Object... args);
+  /**
+   * Add a timestamped annotation, with the specified microseconds in UTC time, and an optional payload.
+   **/
+  Span log(long instantMicroseconds, String message, /* @Nullable */ Object payload);
```
bensigelman: per the above, the current spec supports an event name but *not* a formatted debug message... e.g., this comment in the Go repo: https://github.com/opentracing/opentracing-go/blob/master/span.go#L103

There's an inferior description in the spec as well: http://opentracing.io/spec/#span

opentracing/src/main/java/opentracing/Span.java
michaelsembwever: But any future debug-logging signature in the future would override the above signature, being more exact.

bensigelman: This is what I had in mind, thinking forward past https://github.com/opentracing/opentracing.github.io/issues/30#issuecomment-188464752:

```
Span logEvent(String eventName, /* *Nullable */ Object payload);
Span info(String message, Object... params);
Span error(String message, Object... params);
// maybe / maybe-not warn(), etc... TBD.
Span log(
  /* @Nullable */ String eventName,
  /* @Nullable */ String formattedMessage,  // *not* a pre-formatted message+params pair
  bool error,  // could also be a Level if we go for more than info+error... or a @Nullable Bool
  /* @Nullable */ Object payload,
  /* @Nullable */ Instant timestamp)
```

i.e., there is an utterly unwieldy plumbing method (as you would put it) that handles the union of everything else, then specialized logging methods for the common cases of event and debug message logging.

bensigelman: (no strong feelings about how specifically to combine everything into the unwieldy lower-level `log()` method... without named parameters it's inevitably a little gross)

michaelsembwever: I see don't see any reason to rename the current log to logEvent…? Where's the clash?

opentracing/src/main/java/opentracing/Span.java
michaelsembwever: Same reply as above.

I'll fix the apidoc.

opentracing/src/main/java/opentracing/Span.java
bensigelman: Good question, and this may be an agree-to-disagree situation... If the debug-logging methods are named `info` and `error` (or just `info`, or also `warning`, or whatever), then there's no bona fide conflict. The benefit is just making it clear that `log()` is strictly the superset plumbing method and that the other special-purpose convenience methods have more descriptive names to distinguish them.

It sounds like you'd prefer to have the event-logging method just overload the plumbing API with a signature like `log(String, Object)`. My preference was to call it `logEvent(String, Object)` or even just `event(String, Object)`... that way `event` (or `logEvent`) is on equal footing with `info` and `error`, the other convenience methods.

Another non-possibility is to have all of the above overload `log`, but my original point was that this would be impossible since the event-logging and debug-logging signatures would look the same to the compiler.

bensigelman: TL;DR: what you have here is ok with me if you feel strongly. If it's all the same, I would prefer `logEvent` or `event` for the event-logging version of `log`. LGTM regardless.

michaelsembwever:
This looks good to merge if there's no further blockers.
 (it will wait for PR#11 and PR#10 though)

bensigelman:
@michaelsembwever why not merge this now? It doesn't formally depend on #10 or #11, correct?

michaelsembwever:
> It doesn't formally depend on #10 or #11, correct?

github practice more than theory. because the current branching looks like

master -> interfaces -> pr#9 -> pr#10 -> pr#11

that is, it's nice to fold in from the right.
otherwise the destinations to the PRs become invalid, and the merge button invalid.
although it really only means just rebasing then merging the PRs manually against interfaces.

bensigelman:
@michaelsembwever I really don't want to merge #11 first -- there are tons of issues to discuss in there that don't even affect the public API. IMO it makes the most sense to merge this and rebase #10 accordingly, though I don't feel as strongly about that.

michaelsembwever:
ok. have merged this into interfaces.

PR #10, opened 2016-02-21
Simplify the API: update to the new Specification

michaelsembwever:
Here's the top level API fulfilling the new Specification.

For example implementation on top of this see https://github.com/opentracing/opentracing-java/pull/11


opentracing/src/main/java/opentracing/propagation/SplitTextFormatInjector.java
```diff
@@ -0,0 +1,24 @@
...
+package opentracing.propagation;
+
+import java.util.Map;
+
+/**
+ *
+ */
+interface SplitTextFormatInjector extends Injector<Map> {
+    static final String TRACE_IDENTIFIERS = "Open-Tracing-Context-Id-";
+    static final String TRACE_ATTRIBUTES = "Open-Tracing-Trace-Tags-";
```
yurishkuro: I don't think these constants belong in the public API. The objective of the carrier is to provide a data structure that the end-user code can use to serialize into the outgoing request, without knowing implementation details of the tracer. In this case this data structure is a map with a priori defined key prefixes, which does not meet the objective because, for example, many Zipkin tracers expect the final headers to be in the form of `B3-xxx`.

opentracing/src/main/java/opentracing/Tracer.java
```diff
@@ -31,7 +36,9 @@
    *                   .setTag("lucky_number", 42);
    * }</pre>
    */
-  Span startTrace(String operationName);
+  Span startSpan(String operationName, Optional<Span> parent);
+
+  <T> Injector<T> getInjector(Class<T> carrierType);
```
bensigelman: @michaelsembwever @yurishkuro 

I was having trouble sleeping last night and found myself thinking about this API as well... I had a partially-overlapping idea: namely that there's nothing gained from an abstraction standpoint by having both `getInjector(x)` and `Injector.injectSpan(span, x)`. An `Injector` impl written for `Tracer` implementation `TracerA` cannot be reused for `Tracer` implementation `TracerB`... maybe there are helpers that could be factored out for the benefit of `TracerA` and `TracerB`, but we can't expect the various `Injector` and `Extractor` subclasses to work for an arbitrary `Tracer` implementation.

If we went with the approach here – i.e., no "format" arguments, which does work naturally in Java per previous discussions – the API would look more like

```
public interface Tracer {
  ...

  <T> void injectSpan(Span span, T carrier);
}
```

... or, if we wanted to make things more loosely coupled:

```
public interface Injector {
  <T> void injectSpan(Span span, T carrier);
}

public interface Tracer extends Injector {
  ...
}
```

Thoughts? (cc @tschottdorf since similar things could be done in Go with or without the `format` strings... it seems like we often have these format-specific carrier types in Go and the `reflect` package could be used to reduce the number of concepts in that case)

opentracing/src/main/java/opentracing/propagation/SplitTextFormatInjector.java
michaelsembwever: > I don't think these constants belong in the public API.

Yes i agree. And nothing else in the API here requires them to be public.
I'll move them to a more appropriate less visible place.

opentracing/src/main/java/opentracing/Tracer.java
tschottdorf: Maybe I'm not reading this right, but are you basically suggesting going towards https://github.com/opentracing/opentracing-go/pull/48#issuecomment-181047181? I.e. (this is copy-pasta from there):

```go
type Tracer interface {
  // Join attempts to obtain a Span from the given interface.
  Join (interface{}) (Span, error)
}

type Span interface {
  // Inject attempts to store the Span's information in the given interface.
  Propagate(interface{}) error
}

t := something.NewTracer()
sp := t.StartTrace("foo")
var req http.Request
// This will add the trace information to the request's headers.
if err := t.Propagate(&req); err != nil {
  return err
}

// Receiving end uses the Tracer to join a new Span onto the one
// encoded in the Request.
if sp, err := t.Join(&req); err != nil {
  // ...
}
```

bensigelman: @tschottdorf morally speaking, yes. Though of course the naming is different and they're both ultimately methods on Tracer.

The biggest difference is more a matter of convention than a matter of the literal interface, at least in Go: the "required carriers" would include types like `opentracing.SplitBinaryCarrier` and so forth: i.e., definitely not `map[string]string`. `http.Header` might be ok, though that's on the border. In dynamically-typed languages there might still be the need for something like the current API's `format`.

Anyway, as much as I like sweeping Injector and Extractor under the rug (since most tracing instrumentation _eventually_ is intra-process rather than inter-process), the 1.0RC API is arguably too abstract. Curious to hear what others think.

(I realize it's potentially annoying to vacillate about this, but better to adjust things now than later if we come to consensus about a better path)

bensigelman: let's continue this here: https://github.com/opentracing/opentracing.github.io/issues/70

opentracing/src/main/java/opentracing/propagation/SplitTextFormatInjector.java
michaelsembwever: Oh, I forgot, but this is actually **not** the public API.
The interface is intentionally package-protected.

opentracing/src/main/java/opentracing/Tracer.java
```diff
@@ -31,7 +34,8 @@
    *                   .setTag("lucky_number", 42);
    * }</pre>
    */
-  Span startTrace(String operationName);
+  Span startSpan(String operationName, Optional<Span> parent);
 
-  SpanPropagator getSpanPropagator();
+  <T> void inject(Span span, T carrier);
+  <T> Span join(String operationName, T carrier);
```
bensigelman: elsewhere this has been called `joinTrace` -- any reason for the deviation?

opentracing/src/main/java/opentracing/propagation/Extractor.java
```diff
@@ -0,0 +1,23 @@
...
+ * or implied. See the License for the specific language governing permissions and limitations under
+ * the License.
+ */
+package opentracing.propagation;
+
+import opentracing.Span;
+
+/**
+ */
+public interface Extractor<T> {
```
bensigelman: this (and `Injector<T>`) don't need to be part of `opentracing-java` now that `Tracer` has `inject(...)` and `join[Trace](...)`.

opentracing/src/main/java/opentracing/propagation/SplitTextFormatInjector.java
```diff
@@ -0,0 +1,24 @@
...
+ * the License.
+ */
+package opentracing.propagation;
+
+import java.util.Map;
+
+/**
+ *
+ */
+interface SplitTextFormatInjector extends Injector<Map> {
```
bensigelman: if we're using `Map<String, String>` as the carrier type then there's no need for this to be part of `opentracing-java`... that said, there *does* need to be a comment somewhere about the carrier types (and semantics thereof) that `opentracing.Tracer` impls must support for injecting/joining.

Note that the other languages use two string maps to represent the "split" injected/extracted Span... in @adriancole's initial OT-java PR it looked like this: https://github.com/opentracing/opentracing-java/pull/4/files#diff-e4b78bf6849befce360697858137714dR37 (naming is OOD at this point, of course).

I think I prefer something like

```
interface SplitCarrier<E> {
  /** Implementation-specific Tracer State */
  void setTracerState(E);
  E getTracerState();

  /** Any Baggage that needs to be injected/extracted between processes */
  void setBaggage(E);
  E getBaggage();
}
```

(Or maybe even a concrete class rather than an interface; it's really just a struct, right?)

Then the aforementioned comment about mandatory carrier types would basically boil down to `SplitCarrier<Map<String, String>>` and `SplitCarrier<byte[]>`.

yurishkuro: @bensigelman your suggestion is another example where `format` argument is useful, because if we did what you suggest, there is no way to determine the actual type of `<E>` at runtime, due to generic type erasure.

bensigelman: Ah, right, Java generics!

My point remains, though, that the user of `SplitTextCarrier` or `SplitCarrier<Map<String, String>>` (i.e., someone writing the client or server side of an IPC handoff) will not be able to distinguish the "tracer state" from the "baggage" unless we document the `TRACE_*` prefix strings and expose them publicly. So maybe this means I'd advocating for a `format` identifier, too, following suit with the other platforms...

yurishkuro: > will not be able to distinguish the "tracer state" from the "baggage" unless we document the TRACE_* prefix strings and expose them publicly.

that was exactly my point below and in the other PR with actual injectors impl. In order to be a "split" carrier, it needs to have two maps, otherwise end user cannot distinguish which items in the map are which. Sharing the prefix strings is not an option because the prefixes _must_ be different for different tracers (the `B3_xxx` example). The two maps should be the final, ready for wire representation, the end user should not need to peek into them.

opentracing/src/main/java/opentracing/Tracer.java
michaelsembwever: only because there's no reference in the spec as to what it is. 

in https://github.com/opentracing/opentracing.github.io/issues/70 `injectSpan` had been shortened to `inject`, so i'd presumed the same for `join`.

opentracing/src/main/java/opentracing/propagation/Extractor.java
michaelsembwever: True, they are more part of the internal api now. 

Sitting here they are meant as that: nothing in the primary package references them.
I think i'll create the separate submodule "opentracing-internal-api" and move them to that.

opentracing/src/main/java/opentracing/propagation/SplitTextFormatInjector.java
michaelsembwever: I'm not in favour of adding additional methods into the class. This is the point of the delegation model proposed.

In the subsequent PR#11 you see the split by submodules between public api and internal api more clearly.

The user does not care which of the map entries are what unless they are working/implementing off the internal api, hence the prefixes are appropriate.

opentracing/src/main/java/opentracing/Tracer.java
bensigelman: Ah, got it. Makes sense.

opentracing/src/main/java/opentracing/propagation/Extractor.java
bensigelman: Per my comment (https://github.com/opentracing/opentracing-java/pull/11#discussion_r54204792) on #11: without taking a particular stance on any of the "internal" APIs, it would simplify (and accelerate) this review process if we could focus on the public-facing aspects of *this* PR and get it in as a much-needed stake-in-the-ground for `opentracing-java`. Sound ok?

opentracing/src/main/java/opentracing/propagation/SplitTextFormatInjector.java
bensigelman: @michaelsembwever:

> The user does not care which of the map entries are what unless they are working/implementing off the internal api, hence the prefixes are appropriate.

The entire purpose of the "split" carriers is that the user *does* care which entries are tracer state and which are baggage. This is described here: http://opentracing.io/spec/#required-injectorextractor-formats

(FYI, this is not some imagined use-case: Uber's protocols need to put the (fixed-length) tracer state in one place and the baggage in another when RPCs are made. Also, some organizations would be freaked out about Baggage in general, and by exposing it separately it gives OT users an opportunity to keep trace state propagation but guarantee that Baggage never gets sent over the wire.)

bensigelman: It will probably save us some round-trips if I'm more specific about where I'm at with this...

We need to support the split carriers like we are doing in every other platform. (And the `Map<String, String>` carrier you have here is *not* a "split carrier" since the OpenTracing user cannot reference the two pieces independently.

BTW, I can see an argument to add required carrier format(s) for "non-split" binary and/or text data, but that would be *in addition to* (not instead of) the existing required split-text and split-binary formats.

michaelsembwever: > The entire purpose of the "split" carriers is that the user does care which entries are tracer state and which are baggage.

That user then involves itself with the internal api. Like i've described before its a very niche example that's crept into OpenTracing-1.0 (when OpenTracing should be focusing on much broader issues). But it remains completely feasible and elegant to solve, just with the internal api. This keeps a niche use-case from complicating the public api.

bensigelman: @michaelsembwever it's not a niche example... you have previously raised your concern (which I agree with, btw) about the potentially unbounded cost of the Baggage feature in a distributed system... owners of said distributed systems should be able to strip out baggage without requiring changes to the OT implementation they're using, and that's something the `Split*` carriers accomplish.

(There is also the more esoteric but nevertheless real requirement from Uber regarding the separation of the fixed-width tracer state from the baggage, but that is secondary in my mind)

opentracing/src/main/java/opentracing/Span.java
```diff
@@ -54,10 +45,17 @@
...
-  // numbers kindof suck.. we've no idea if this is a float, how many bits, etc.
   Span setTag(String key, Number value);
 
-  Span setTraceAttribute(String key, String value);
+  /**
+   * Set a Baggage item, represented as a simple string:string pair.
+   *
+   * Note that newly-set Baggage items are only guaranteed to propagate to future children of the given Span.
+   */
+  Span setBaggage(String key, String value);
```
bensigelman: `setBaggageItem` per the spec and other languages... the "Baggage" itself is a set.

michaelsembwever:
the destination branch has been merged into the `interfaces` branch.
this PR must now manually be also merged into `interfaces` branch.

opentracing/src/main/java/opentracing/Span.java
michaelsembwever: fixed.

opentracing/src/main/java/opentracing/Tracer.java
```diff
@@ -13,25 +13,41 @@
...
    */
-  Span startTrace(String operationName);
+  Span startSpan(String operationName, /* @Nullable */ Span parent);
+
+  /** Takes two arguments:
+   *    a Span instance, and
+   *    a “carrier” object in which to inject that Span for cross-process propagation.
+   *
+   */
+  <T> void inject(Span span, T carrier);
```
bensigelman: I see that the controversial carriers are no longer part of this PR -- works for me. :)  I like incremental PRs and it's great if we can merge the less controversial parts first.

That said, we do have a bit of a fork-in-the-road re the "carriers"... regardless of the Split/NonSplit stuff we have been discussing, is it realistic to say that a `byte[]` is *always* a specific carrier type? I would argue that it's certainly not realistic... so we will end up with a situation where thin classes need to wrap around things like a `byte[]` in order to get the necessary type information into `inject`. Not the end of the world, but I do want to make sure this is a deliberate choice... in other languages we have been using explicit identifiers/strings/whatever to determine the type/semantics of the carrier parameter.

opentracing/src/main/java/opentracing/Tracer.java
```diff
@@ -13,25 +13,41 @@
...
-   * Span feed = tracer.startTrace("GetFeed");
-   * Span http = tracer.startTrace("HandleHTTPRequest")
+   * Span feed = tracer.startTrace("GetFeed", null);
+   * Span http = tracer.startTrace("HandleHTTPRequest", feed)
    *                   .setTag("user_agent", req.UserAgent)
    *                   .setTag("lucky_number", 42);
    * }</pre>
    */
-  Span startTrace(String operationName);
+  Span startSpan(String operationName, /* @Nullable */ Span parent);
```
bensigelman: There are a few other nullable/optional params for the startSpan plumbing interface... for reference:

https://github.com/opentracing/opentracing-go/blob/master/tracer.go#L76

(Users recording Spans after-the-fact will need to set an explicit timestamp, and there are times when the `startSpan` impl may want to return different concrete `Span` impls depending on the tags specified at `startSpan`-time)

opentracing/src/main/java/opentracing/propagation/SplitTextFormatInjector.java
michaelsembwever: > owners of said distributed systems should be able to strip out baggage without requiring changes to the OT implementation they're using

The dropping of baggage is an operational optimisation, a safeguard, not something that's intrinsic part of the public api.

Furthermore i believe that it's something that otherwise anyway can be done in a more elegant, non-intrusive to the public api, manner.

For example look at how the  [SplitTextFormatInjectorImpl.java](https://github.com/opentracing/opentracing-java/pull/11/files#diff-f2ad9c00db83aa2576fe12ea2f7f0413) solves it.

opentracing/src/main/java/opentracing/Tracer.java
michaelsembwever: I'll add the timestamp.

returning alternative span impls is done by overriding AbstractTracer (or implementing Tracer from scratch).

opentracing/src/main/java/opentracing/Tracer.java
michaelsembwever: > is it realistic to say that a byte[] is always a specific carrier type?

the IoC approach, ie `injector.canHandle(byteBuffer)`, is a strategy forward to solving preference and finer details to injector impls against carrier objects. There's also plenty of room in how the Tracer impl is configured and initialised in regards to the registration of injectors and extractors.

opentracing/src/main/java/opentracing/Tracer.java
bensigelman: the point is that the alternative Span impl depends on the tags specified by `startSpan` -- not an issue of whether one implements AbstractTracer or Tracer or whatever else, it's an issue of having the tags on-hand when making the decision about which Span impl to choose. (I.e., a single Tracer impl can create two different Span impls)

opentracing/src/main/java/opentracing/Tracer.java
bensigelman: That's not my point. Consider the case where there are two raw binary formats someday.

```
  byte[] byteBuffer = ...;
  tracer.inject(someSpan, byteBuffer);
```

How does the Tracer impl determine which of the two binary formats to choose from if they both use `byte[]` as a carrier? It's underspecified.

One approach:

```
  byte[] byteBuffer = ...;
  tracer.inject(someSpan, SomeScope.FORMAT_A_ID, byteBuffer);
```

Another approach:

```
class CarrierA {
  public byte[] buffer() ...;
  public void setBuffer(byte[] b) ...;
}
...
  CarrierA carrier = ...;
  tracer.inject(someSpan, carrier);
```

Either of the above is feasible IMO (I prefer the former, but whatever). It's not reasonable to expect that `byte[]` always implies a specific format, though.

yurishkuro: To add to this, it also cannot be solved by static configuration of the tracer, because the same process may use different RPC frameworks simultaneously (e.g. different storage backends), and each RPC framework may interpret `byte[]` format differently.

opentracing/src/main/java/opentracing/propagation/SplitTextFormatInjector.java
bensigelman: ... I'm going to leave this discussion alone until #10 is merged since the controversial bits are not part of this PR at the moment. For whatever reason we are failing to communicate about some fairly essential aspects of this (what's public, what's not, that sort of thing) and I think it makes more sense to resolve and merge the significant amount of the API we *do* agree on before tackling the remaining bits.

opentracing/src/main/java/opentracing/Tracer.java
michaelsembwever: > … depending on the tags specified at startSpan-time

Are you talking about an additional map<string,object> tags parameter, or those tags specified within the parent span parameter?

opentracing/src/main/java/opentracing/Tracer.java
michaelsembwever: First, and foremost, I don't see the need for this today. The format parameter remains optional, and can be added harmlessly if needed when the time is appropriate.

Otherwise I'm more in favour the the later approach where the ByteBuffer is accordingly (and thinly) wrapped. If we get to requiring this it also brings up whether we should be returning to the tracer.getInjector(..) and tracer.getExtractor(..) approach. 

Note that it was not me that advocated for the removal of those intermediate methods, for a smaller api simplification that what the delegation model here provides.

michaelsembwever: >  it also cannot be solved by static configuration of the tracer, because the same process may use different RPC frameworks simultaneously

There's still plenty of room to deal with that scenario effectively. The inject method  an analyse both the span and the carrier.

opentracing/src/main/java/opentracing/propagation/SplitTextFormatInjector.java
michaelsembwever: > makes more sense to resolve and merge the significant amount of the API we do agree on before tackling the remaining bits.

+1

opentracing/src/main/java/opentracing/Tracer.java
bensigelman: I'm talking about the additional map tags parameter.

opentracing/src/main/java/opentracing/Tracer.java
bensigelman: If this is a war of attrition then I give up. :)

If you don't want to support the format parameter, let's just get this in without it and we can write concrete, non-templated carrier types rather than passing ambiguous `byte[]` buffers to inject/join (which will not work in practice).

PS: you wrote...

> Note that it was not me that advocated for the removal of those intermediate methods, for a smaller api simplification that what the delegation model here provides.

The intermediate methods (and types) are not relevant here... the question is whether runtime type information alone is sufficient to identify a carrier. The intermediate layer of Injector/Extractor interfaces did not make that any easier or harder so I'm not sure why it's being brought up here.

opentracing/src/main/java/opentracing/Tracer.java
michaelsembwever: Do we have any examples of this?

Something like this is already happening [here](https://github.com/opentracing/opentracing-java/pull/11/files#diff-f202d93da32a2cfe723cbf28dbb95522R30) and  [here](https://github.com/opentracing/opentracing-java/pull/11/files#diff-a90b8918cb3610641d03a6d8bf014704R43).

But i'm unclear as to what its purpose or need is exposed higher up in `Tracer`?
When would the caller want to decide which Span implementation to use? (that seems like an anti-pattern).

If the Tracer can't decide which Span is required without first holding attributes it'll have, then I think the Builder pattern is more appropriate.

opentracing/src/main/java/opentracing/Tracer.java
michaelsembwever: I've advocated that the format parameter be optional, not removed.
I just don't feel for adding public methods that we're forever bound to prematurely.

> The intermediate methods (and types) are not relevant here …

What I meant by that is… the problem introduced here regarding formats, was solved with the tracer.getInjector(..) and tracer.getExtractor(..)

(the tying of generics around carrier did over emphased the binding, but I would have removed that, or permitted tracer.getInjector(string) calls so to satisfy the case of overlapping carrier types. once again i would have advocated for it when needed.)

Taking a look at the current [java Public API](https://github.com/opentracing/opentracing-java/tree/mck/delegation-design-0/opentracing/src/main/java/opentracing) I'm quite content with how it has simplified and become incredibly easy to quickly comprehend. It will grow of course, but this is a great place for "1.0" in my opinion. Post a released 1.0 hopefully, with an api that everyone can quickly grab uncluttered by specific use-cases, we'll see further implementing into the different tracing instrumentations out there, which can provide us with even more use-cases and corner-cases.

bensigelman: Re the format parameter: I'm hoping we just don't need to add it to `opentracing-java` in the future. Since it's idiomatic to have an explosion of small facade classes in Java, I think that's realistic, too.

It had its merits, but having both `Tracer.getInjector(...)` and `Injector.injectSpan(...)` added too much cognitive load on the new OT user... my (very bright) coworker took a look at that scheme a few weeks ago and was quite confused, for example (that's actually why I opened https://github.com/opentracing/opentracing.github.io/issues/70). The transition to `Tracer.Inject()` in Go consolidated [these](https://github.com/opentracing/opentracing-go/pull/65/files#diff-66487bb56b848b3f1635aa3aa38f0f90L30) [three](https://github.com/opentracing/opentracing-go/pull/65/files#diff-66487bb56b848b3f1635aa3aa38f0f90L37) [comments](https://github.com/opentracing/opentracing-go/pull/65/files#diff-46db78603208f38452c0333bec12ab2cL30) into a single straight-line explanation of [Tracer.Inject()](https://github.com/opentracing/opentracing-go/pull/65/files#diff-46db78603208f38452c0333bec12ab2cR30)... much easier to grok, I think/hope, and that's a great thing for the fundamentally tricky (and essential) propagation API.

opentracing/src/main/java/opentracing/Tracer.java
bensigelman: @michaelsembwever:

> Do we have any examples of this?

There's one written up here: http://opentracing.io/use-cases/#setting-sampling-priority-before-the-trace-starts

I can see what you mean about the builder pattern, esp since that's so widely used in Java-land. The alternative of passing a hodge-podge of `null`s to `startSpan` is a little gross. Then again, what a bummer to add a builder pattern just for the sake of this one esoteric corner-case...

I can't claim to have particularly strong feelings about this one, but hopefully the above is motivating in some way. My understanding is that this is all somewhat related to the `debug=true` case in Zipkin traces... maybe that rings a bell?

> When would the caller want to decide which Span implementation to use? (that seems like an anti-pattern).

No, that's not what I meant... it's the *Tracer* that would decide which Span implementation to use, and it would do so by consulting these tags provided at `startSpan()` time.

Consider the case where the Tracer impl has a zipkin-like sampling scheme. Let's further assume that the Tracer impl has two concrete Span subclasses it uses: `SampledSpan` and `UnsampledSpan`. Now, what if that Tracer impl wants to allow its user to request a sampled trace without exposing anything about `SampledSpan`/`UnsampledSpan`: this is what [the opentracing.io example](http://opentracing.io/use-cases/#setting-sampling-priority-before-the-trace-starts) is about, of course. So the user can write something like...

```
Span s = tracer.startSpan(
    "...",
    null,
    // Not valid Java, just trying to communicate the idea:
    {"sampling.priority": 1});
```

... and the `tracer` will use the tag to select the `SampledSpan` impl.

Hope that makes the motivation clearer.

bensigelman:
Well, the PR comments are officially a longer than the code. :)

The only real blocker for me at the moment is the lack of a `/* @Nullable */ Map<String, String> tags` parameter to `startSpan`... the builder pattern is tempting but overkill in my book, especially since 99% of the time the `tags` param will be left off entirely. So I would personally favor plain-old overloading...

```
startSpan(String opName, /* @Nullable */ Span parent);
startSpan(String opName, /* @Nullable */ Span parent, /* @Nullable */ Map<String, String> tags);
```

I am up for agreeing to disagree about the `format` param to inject()/join(). There are also no actual "OT-sponsored" carrier types in this PR, but those can – and probably should – be added and discussed after this PR is merged.

Thanks, getting closer...

opentracing/src/main/java/opentracing/Tracer.java
michaelsembwever: I am too for the far simpler Tracer .inject() and .join(). It absolutely is less cognitive load on the new user.

I'd be keen to evaluate the different options open (eg format vs explicit Injector/Extractor) when the time arises.

opentracing/src/main/java/opentracing/Tracer.java
michaelsembwever: Yes, that was clear. Thanks.
I will add the tags map.

michaelsembwever: ^I said I'd add the tags but first…

Once I started to code this it's already pretty obvious how more intuitive (and needed) the builder pattern is. Rather than add it to this PR I've spawned it off as a new https://github.com/opentracing/opentracing-java/pull/12.

The Builder pattern has my vote. But it depends on how the Specification can be interpreted. If there's any objections I'll add the map as I said I would.

michaelsembwever:
see https://github.com/opentracing/opentracing-java/pull/10/files#r54381791
 and https://github.com/opentracing/opentracing-java/pull/12

opentracing/src/main/java/opentracing/Tracer.java
bensigelman: Sold re #12. Thanks.

michaelsembwever: This PR is getting on indeed, but I feel that it's often going from strength to strength.

It looks like I'll merge in #12 soon, the builder pattern of course came with its own little flurry of design questions.

michaelsembwever:
If there's no objections with this code (including that from https://github.com/opentracing/opentracing-java/pull/12, which will get merged into this first) I will merge this to `interfaces` in 12+hrs.

bensigelman:
LGTM

michaelsembwever:
Has been merged into `interfaces`

PR #11, opened 2016-02-21
Example helping implementation of the top-level API

michaelsembwever:
This code provides a basic implementation meeting the requirements that the specification imposes without adding additional API.

The code also provides that basic implementation in an non-tracer specific way.


yurishkuro:
This implementation is akin to `GoHTTPHeader` format in Go API, but it's not really a _split_ implementation. For example, the injector does not provide the ability to opt out of propagating trace attributes. The only way to achieve opt out would be to pass an intermediary Map, and then loop through its keys and filter them out by prefix, which breaks the other design goal of the end user not knowing which header names individual implementations use to propagate the context.

michaelsembwever:
> I don't think these constants belong in the public API.


Implementing the *split* is trivial.

You could parameterise the `SplitTextFormatInjectorImpl`, 
eg `splitTextFormatInjectorImpl.setTraceAttributesIncluded(false);`

or parameterise the tracer and have it register a different injector for maps, 
eg `registry.register(Map.class, new TraceStateOnlyFormatInjectorImpl());`

opentracing-java8/src/main/java/opentracing/propagation/SpanImpl.java
```diff
@@ -0,0 +1,115 @@
...
+    }
+
+    @Override
+    public String getBaggage(String key) {
+        return traceAttributes.get(key);
+    }
+
+    @Override
+    public Span log(String message, /* @Nullable */ Object payload) {
+        Instant now = Instant.now();
```
adriancole: cool. hiding this allows use of the same interface regardless of java 8 vs java 7 (or even regardless of library)

opentracing-api/src/main/java/opentracing/propagation/Extractor.java
```diff
@@ -17,7 +17,7 @@
 
 /**
  */
-public interface Extractor<T> {
+    public interface Extractor<T> {
```
bensigelman: sorry for the ultra-nitpick, but is there some Java whitespace convention happening here or is it just an errant `<Tab>`?

opentracing-java8/src/main/java/opentracing/propagation/SpanImpl.java
```diff
@@ -0,0 +1,115 @@
...
+        return this;
+    }
+
+    @Override
+    public String getBaggage(String key) {
+        return traceAttributes.get(key);
+    }
+
+    @Override
+    public Span log(String message, /* @Nullable */ Object payload) {
```
bensigelman: not sure of the precise relationship between this PR and the other logging-specific PR, but same comments apply here... (https://github.com/opentracing/opentracing-java/pull/9#discussion_r54151986)

opentracing-java8/src/main/java/opentracing/propagation/AbstractTracer.java
```diff
@@ -0,0 +1,108 @@
...
+import java.util.Optional;
+import java.util.concurrent.ConcurrentHashMap;
+import java.util.concurrent.ConcurrentMap;
+import opentracing.Span;
+import opentracing.Tracer;
+import opentracing.propagation.Extractor;
+import opentracing.propagation.Injector;
+import opentracing.propagation.SpanImpl;
+
+abstract class AbstractTracer implements Tracer {
```
bensigelman: I like the idea – per http://opentracing.io/use-cases/#opentracing-priorities – of truly minimizing the number of classes/concepts in the core `opentracing-java` repo. We could do something like `basictracer-java` (a la https://github.com/opentracing/basictracer-go) if we want to create helpers for future implementors.

opentracing-api/src/main/java/opentracing/propagation/Extractor.java
michaelsembwever: my bad.

opentracing-java8/src/main/java/opentracing/propagation/AbstractTracer.java
michaelsembwever: this class sits in the opentracing-java8 submodule and artifact, as opposed to the opentracing-api submodule and artifact. That's intended to address that concern.

opentracing-java8/src/main/java/opentracing/propagation/SpanImpl.java
michaelsembwever: this PR is based off the top of PR#10, which in turn is based off the top of PR#9

opentracing-java8/src/main/java/opentracing/propagation/AbstractTracer.java
bensigelman: Let's figure out the public `opentracing-java` API (in the `opentracing-api` artifact) first and separately. There's no need to couple this partial/abstract implementation with the interface itself, and it adds noise to the review thread as well as the `opentracing-java` repository.

michaelsembwever: yes, as described in the original description that is the purpose of this PR as opposed to PR#10. Although there are some things in PR#10 that can get moved to this PR.

bensigelman: ok, I will spare us the back-and-forth on #11 until #10 is in (if that sounds reasonable).

opentracing-impl-java8/src/main/java/opentracing/propagation/AbstractTracer.java
```diff
@@ -0,0 +1,111 @@
...
+
+    public <T> Injector<T> register(Class<T> carrierType, Injector<T> injector) {
+        return registry.register(carrierType, injector);
+    }
+
+    public <T> Extractor<T> register(Class<T> carrierType, Extractor<T> extractor) {
+        return registry.register(carrierType, extractor);
+    }
+
+    private static class PropagationRegistry {
```
adriancole: this is basically like a mini-spring. I like that there's an implementation for a DI-free environment. you could imagine a scenario where register doesn't exist and inject/join make calls to a BeanFactory

ex.
```java 
class SpringTracer implements Tracer {

   // satisfy the api contract which requires runtime lookups by class.
   @Override
   public <T> void inject(Span span, T carrier, Class<T> carrierType) {
     Injector<T> injector = beanFactory.getBean(typeOf(Injector.class, carrierType));
     return injector.injectSpan(span, carrier);
   }
```
```

Comment (adriancole) [inline opentracing-impl-java8/src/main/java/opentracing/propagation/AbstractTracer.java]: in reality, such a tracing system would only support ^^ for interop reasons. In real life, the Injector/Extractor types would be injected directly into the call sites.

Ex. 
```java
// ignore the interfaces, and how the fields are initialized
class InjectCurrentSpanIntoMetadata implements Func1<Metadata> {
  Context.Key<Span> spanKey;
  Injector<Metadata> injector;

  void apply(Metadata input) {
    injector.injectSpan(spanKey.get(), input);
  }
}
```

Main thing is that the above type really needs no reference to a tracer to do its job, and it can be modified independently without impacting the tracer.

michaelsembwever:
I'm closing this pull request in favour of opening it again with the correct destination and to start discussion afresh.

PR #12, opened 2016-02-29
Tracer.startSpan(..) with a builder for specifying the arguments

michaelsembwever:
See https://github.com/opentracing/opentracing-java/pull/10/files#r54330197

### DIFF
--- a/opentracing/src/main/java/opentracing/Tracer.java
+++ b/opentracing/src/main/java/opentracing/Tracer.java
@@ -25,20 +25,19 @@ public interface Tracer {
    *
    * <p>Example:
    * <pre>{@code
-   * Tracer tracer = ...
-   * Span feed = tracer.startTrace("GetFeed", null);
-   * Span http = tracer.startTrace("HandleHTTPRequest", feed)
-   *                   .setTag("user_agent", req.UserAgent)
-   *                   .setTag("lucky_number", 42);
-   * }</pre>
-   */
-  Span startSpan(String operationName, /* @Nullable */ Span parent);
+    Tracer tracer = ...
 
-  /**
-   * Same as {@link #startSpan(String, Span)},
-   * but allows to specify a past timestamp in microseconds when the Span was created.
+    Span feed = tracer.buildSpan("GetFeed")
+                      .start();
+
+    Span http = tracer.buildSpan("HandleHTTPRequest")
+                      .withParent(feed)
+                      .withTag("user_agent", req.UserAgent)
+                      .withTag("lucky_number", 42)
+                      .start();
+    }</pre>
    */
-  Span startSpan(String operationName, long microseconds, /* @Nullable */ Span parent);
+  SpanBuilder buildSpan(String operationName);
 
   /** Takes two arguments:
    *    a Span instance, and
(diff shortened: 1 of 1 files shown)


yurishkuro:
I like this approach. It's different from other languages, but then in Python optional named params are the norm, and in Go a SomethingWithOptions() pattern is pretty common, so they don't really need builder pattern. But in Java this seems fairly idiomatic.

The only change I would suggest is renaming `startSpan` to `createSpan` or  `buildSpan`.

opentracing/src/main/java/opentracing/Tracer.java
```diff
@@ -25,20 +25,19 @@
...
+
+    Span http = tracer.startSpan("HandleHTTPRequest")
+                      .withParent(feed)
+                      .withTag("user_agent", req.UserAgent)
+                      .withTag("lucky_number", 42)
+                      .start();
+    }</pre>
    */
-  Span startSpan(String operationName, long microseconds, /* @Nullable */ Span parent);
+  SpanBuilder startSpan(String operationName);
```
bensigelman: nit: how about `buildSpan(...).withX(...).withY(...).withZ(...).start()`?

opentracing/src/main/java/opentracing/Tracer.java
```diff
@@ -56,4 +55,22 @@
    * and others consider the caller to be the parent and the receiver to be the child)
    */
   <T> Span join(String operationName, T carrier);
+
+  interface SpanBuilder {
```
bensigelman: I like it... since java has neither named params nor quick ad-hoc structs, builder patterns work well.

opentracing/src/main/java/opentracing/Tracer.java
```diff
@@ -56,4 +55,22 @@
    * and others consider the caller to be the parent and the receiver to be the child)
    */
   <T> Span join(String operationName, T carrier);
```
bensigelman: ... and join() would return a SpanBulider, too, right?

bensigelman:
LGTM!

opentracing/src/main/java/opentracing/Tracer.java
michaelsembwever: buildSpan(..) naturally reads better. I was only worried about how the Specification needs to be met.

I will change to buildSpan(..)

opentracing/src/main/java/opentracing/Tracer.java
michaelsembwever: I'm not sure. I guess the same example of forcing sampling via tags before the span is created applies here too. Will make it return a SpanBuilder.

michaelsembwever:
code updated. i still need to fix the apidocs around the methods.

michaelsembwever:
If using builders is permitted against the Specification, then I would also look into using a builder on Span.log(..)

That would reduce it to be only one Span.log(..) method.
If you also provide better horizon to the types of logging (ie info/error, etc) advocated for in future releases.

opentracing/src/main/java/opentracing/Tracer.java
```diff
@@ -55,5 +54,23 @@
    * (Note that some OpenTracing implementations consider the Spans on either side of an RPC to have the same identity,
    * and others consider the caller to be the parent and the receiver to be the child)
    */
-  <T> Span join(String operationName, T carrier);
+  <T> SpanBuilder join(String operationName, T carrier);
```
bensigelman: I'm realizing now that the `operationName` param is a little subtle... it's often but not always available at `join()`-time; the simple thing would be to make it `/* @Nullable */`, but we could also add it to `SpanBuilder` and just emphasize the importance of remembering to call it. (Of course there is already `Span.setOperationName()` which can be called anytime before `Span.finish()`)

I think my vote – on balance – is to make the `operationName` here `@Nullable`, though it's a tradeoff.

michaelsembwever: If it is to be @Nullable then i'd rather see it go into the builder, ie `withOperationName(..)`.

The builder is entitled to document which methods are expected to be called (even in different combinations) and for `start()` to throw an AssertionError if still unset when it should be.

bensigelman: @michaelsembwever so is that idiomatic in modern Java? If so, SGTM.

bensigelman: ... oh, but there is a catch: namely that there are circumstances where no operation name is available at `join()` time (hence `Span.setOperationName()`). So an assertion won't work, or at least not that particular assertion. Another possibility would be to add something like

```
  SpanBuilder withoutInitialOperationName();
```

(or similar) which would force the user to specifically opt out of an operation name at `start()`-time.

michaelsembwever: > So an assertion won't work, or at least not that particular assertion.

Yeah i picked up on that after I wrote the comment. See the lastest code.
Mandatory arguments are used when first obtaining a SpanBuilder, ie tracer.buildSpan(name) and tracer.join(carrier).

Otherwise the builder can throw exceptions is the same setter is called twice.

opentracing/src/main/java/opentracing/Tracer.java
```diff
@@ -51,9 +50,44 @@
    *    the operation name for the Span it’s about to create, and
```
bensigelman: we'll need to update this comment

michaelsembwever: my bad. had it – forgot to add it.

bensigelman: ... and it would be wise for the comment to point out that the operation name should be set on the builder if it's available at `join`-time.

I am still hedging about whether the operation name should be a nullable param here, but it's cleaner to leave it off for now, I think. (I'm also having some doubts about what we've done in other languages, esp python where `operation_name` could be an optional named parameter... @yurishkuro thoughts?)

yurishkuro:
@bensigelman 
> I'm also having some doubts about what we've done in other languages, esp python where operation_name could be an optional named parameter... @yurishkuro thoughts?)

I'm ok with making it `operation_name=None`

michaelsembwever:
Any objections with the code as it is now?
If not i'll merge it in ~12hrs.

bensigelman:
LGTM

PR #13, opened 2016-03-02
Initial opentracing-java API

michaelsembwever:
Last chance to review before all this work hits master.

This is a combination of https://github.com/opentracing/opentracing-java/pull/4 , https://github.com/opentracing/opentracing-java/pull/10 , https://github.com/opentracing/opentracing-java/pull/12 , https://github.com/opentracing/opentracing-java/pull/8 , and https://github.com/opentracing/opentracing-java/pull/9 coming into master.

The initial implementation behind this sits in https://github.com/opentracing/opentracing-java/pull/11

ping @adriancole @bensigelman @yurishkuro


opentracing/src/main/java/opentracing/Tracer.java
```diff
@@ -0,0 +1,94 @@
...
+                      .withTag("user_agent", req.UserAgent)
+                      .withTag("lucky_number", 42)
+                      .start();
+    }</pre>
+   */
+  SpanBuilder buildSpan(String operationName);
+
+  /** Takes two arguments:
+   *    a Span instance, and
+   *    a “carrier” object in which to inject that Span for cross-process propagation.
```
adriancole: the word "carrier" is very vague when in practical terms it is very often http headers or some other side-channel (like RPC envelope). might be best to make an example of a carrier.

"For example. tracer.inject(span, httpRequestHeaders);"

opentracing/src/main/java/opentracing/Tracer.java
```diff
@@ -0,0 +1,94 @@
...
+      SpanBuilder withTag(String key, String value);
+
+      /** Same as {@link Span#setTag(String, String)}, but for the span being built. */
+      SpanBuilder withTag(String key, boolean value);
+
+      /** Same as {@link Span#setTag(String, String)}, but for the span being built. */
+      SpanBuilder withTag(String key, Number value);
+
+      /** Returns the started Span. */
+      Span start();
```
adriancole: prefer overloading start(long microseconds) vs withTimestamp().start()

opentracing/src/main/java/opentracing/Span.java
```diff
@@ -0,0 +1,80 @@
...
+  /**
+   * Add a new log event to the Span, accepting an event name string and an optional structured payload argument.
+   * If specified, the payload argument may be of any type and arbitrary size,
+   * though implementations are not required to retain all payload arguments
+   * (or even all parts of all payload arguments).
+   *
+   * The timestamp is specified manually here to represent a past log event.
+   * The timestamp in microseconds in UTC time.
+   **/
+  Span log(long instantMicroseconds, String eventName, /* @Nullable */ Object payload);
```
adriancole: prefer timestampMicroseconds to instantMicroseconds as the former word is used in the javadoc

opentracing/src/main/java/opentracing/Span.java
```diff
@@ -0,0 +1,80 @@
...
+   * Set a Baggage item, represented as a simple string:string pair.
+   *
+   * Note that newly-set Baggage items are only guaranteed to propagate to future children of the given Span.
+   */
+  Span setBaggageItem(String key, String value);
+
+  /** Get a Baggage item by key. */
+  String getBaggageItem(String key);
+
+  /**   *
```
adriancole: there's some things like this, which should be easy enough to scrub.

opentracing/src/main/java/opentracing/Span.java
```diff
@@ -0,0 +1,80 @@
...
+
+  /**
+   * Set a Baggage item, represented as a simple string:string pair.
+   *
+   * Note that newly-set Baggage items are only guaranteed to propagate to future children of the given Span.
+   */
+  Span setBaggageItem(String key, String value);
+
+  /** Get a Baggage item by key. */
+  String getBaggageItem(String key);
```
adriancole: returns null if baggage isn't supported or an item with this key isn't present?

opentracing/src/main/java/opentracing/Span.java
```diff
@@ -0,0 +1,80 @@
...
+   *
+   * <p>This should be the last call made to any span instance, and to do otherwise leads to
+   * undefined behavior.
+   */
+  void finish();
+
+  /**
+   * Set a key:value tag on the Span.
+   *
+   * <p>Tag values can be of arbitrary types, however the treatment of complex types is dependent on
```
adriancole: looks like complex types aren't possible.. do we need this doc?

opentracing/src/main/java/opentracing/Span.java
```diff
@@ -0,0 +1,80 @@
...
+  /**
+   * Sets the end timestamp and records the span.
+   *
+   * <p>This should be the last call made to any span instance, and to do otherwise leads to
+   * undefined behavior.
+   */
+  void finish();
+
+  /**
+   * Set a key:value tag on the Span.
```
adriancole: the behavior of multiple calls with the same key is implementation specific

opentracing/src/main/java/opentracing/Span.java
```diff
@@ -0,0 +1,80 @@
...
+ * is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express
+ * or implied. See the License for the specific language governing permissions and limitations under
+ * the License.
+ */
+package opentracing;
+
+/**
+ * Span represents an active, un-finished span in the opentracing system.
+ *
+ * <p>Spans are created by the {@link Tracer} interface.
```
adriancole: nit: more specific? ex `{@link Tracer#buildSpan}`

opentracing/src/main/java/opentracing/Span.java
```diff
@@ -0,0 +1,80 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing, software distributed under the License
+ * is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express
+ * or implied. See the License for the specific language governing permissions and limitations under
+ * the License.
+ */
+package opentracing;
+
+/**
+ * Span represents an active, un-finished span in the opentracing system.
```
adriancole: ignore if unhelpful.. I prefer "in-flight" to "active, un-finished"

adriancole:
Looks like an accurate, minimal specification of OT as defined. :shipit: when ready

opentracing/src/main/java/opentracing/Tracer.java
michaelsembwever: fixed

opentracing/src/main/java/opentracing/Tracer.java
michaelsembwever: fixed

opentracing/src/main/java/opentracing/Span.java
michaelsembwever: fixed

opentracing/src/main/java/opentracing/Span.java
michaelsembwever: fixed

opentracing/src/main/java/opentracing/Span.java
michaelsembwever: fixed

opentracing/src/main/java/opentracing/Span.java
michaelsembwever: fixed

opentracing/src/main/java/opentracing/Span.java
michaelsembwever: fixed

opentracing/src/main/java/opentracing/Span.java
michaelsembwever: fixed

opentracing/src/main/java/opentracing/Span.java
michaelsembwever: fixed

opentracing/src/main/java/opentracing/Tracer.java
yurishkuro: that seems to defeat the purpose of the Builder pattern. Starting a child span is by far the most frequent usage, so would we add `start(parent)` too?

michaelsembwever: > so would we add start(parent) too?

No. I believe @adriancole suggested this, and why it made sense to me, is because start and timestamp are close concepts. having a choice between calling `start()` or `start(timestamp)` intuitively makes sense. ie start now or start at this timestamp. that's part of designing a fluent api.

> Starting a child span is by far the most frequent usage …

It's not what's frequent, but what intuitively makes sense or reads well, ie fluent api.

That said, it could be argued that `start(timestamp)` isn't intuitive enough, and i've no strong opinion to that. `withTimestamp(..)` is slightly more intuitive if you ask me. so i'd be happy to return the code if this was the argument raised.

bensigelman: IMO, overriding the timestamp is a corner case, and thus we should choose the consistent API (i.e., `withTimestamp(...)`) rather than the shorthand API (i.e., overloading `start(timestamp)`). No strong feelings, though.

opentracing/src/main/java/opentracing/Span.java
```diff
@@ -0,0 +1,78 @@
...
+
+/**
+ * Represents an in-flight span in the opentracing system.
+ *
+ * <p>Spans are created by the {@link Tracer#buildSpan} interface.
+ */
+public interface Span {
+
+  /**
+   * Sets the end timestamp and records the span.
```
bensigelman: Per the `start()` debate, we do need some way to override the finish timestamp in order to do after-the-fact translation of span-like logs, etc. So, `void finish(long timestamp)`?

adriancole: >
> Per the start() debate, we do need some way to override the finish
> timestamp in order to do after-the-fact translation of span-like logs, etc.
> So, void finish(long timestamp)?
>
Brave, which is based on duration not endTs, as this in the span-scoped
tracer:
/**
* Completes the span, which took {@code duration} microseconds.
*/
public void finishSpan(long duration)

bensigelman: I am imagining – perhaps incorrectly – that the caller of the explicit-timestamp version has a raw timestamp rather than a duration on hand (i.e., that they are basically doing log processing / translation). FWIW, I think that tracing *impls* are better off tracking duration since it's not subject to clock-skew cruelty.

opentracing/src/main/java/opentracing/Span.java
```diff
@@ -0,0 +1,78 @@
...
+   *
+   * <p>This should be the last call made to any span instance, and to do otherwise leads to
+   * undefined behavior.
+   */
+  void finish();
+
+  /**
+   * Set a key:value tag on the Span.
+   *
+   * <p>If there is a pre-existing tag set for {@code key}, it is overwritten.
```
bensigelman: Minor point: this behavior (same key, new value) is formally undefined in some other impls. No strong feelings but wanted to point out the discrepancy.

opentracing/src/main/java/opentracing/Span.java
```diff
@@ -0,0 +1,78 @@
...
+  Span setTag(String key, String value);
+
+  /** Same as {@link #setTag(String, String)}, but for boolean values. */
+  Span setTag(String key, boolean value);
+
+  /** Same as {@link #setTag(String, String)}, but for numeric values. */
+  Span setTag(String key, Number value);
+
+  /**
+   * Set a Baggage item, represented as a simple string:string pair.
```
bensigelman: I see you have left out the unpalatable HTTP-header-inspired restrictions on the `key` here... intentional? It is a deviation from the spec in any case.

opentracing/src/main/java/opentracing/Tracer.java
```diff
@@ -0,0 +1,95 @@
...
+                      .start();
+    }</pre>
+   */
+  SpanBuilder buildSpan(String operationName);
+
+  /** Takes two arguments:
+   *    a Span instance, and
+   *    a “carrier” object in which to inject that Span for cross-process propagation.
+   *
+   * A “carrier” object is some sort of http or rpc envelope, for example HeaderGroup (from Apache HttpComponents).
```
bensigelman: We need to add the OT-required carriers to Java at some point, though FWIW I would like to make those required carriers clearer. https://github.com/opentracing/opentracing.github.io/issues/73

If the importance of "required carriers" is not obvious, I can explain...

opentracing/src/main/java/opentracing/Tracer.java
```diff
@@ -0,0 +1,95 @@
...
+   *
+   * The low-level format carriers Map&lt;String,String> and ByteBuffer are guaranteed to be supported,
```
bensigelman: nit: `&gt;`?

opentracing/src/main/java/opentracing/Tracer.java
bensigelman: And to be clear: `Map<String, String>` and `ByteBuffer` are not interoperable with the required carriers in the other platforms.

opentracing/src/main/java/opentracing/Tracer.java
```diff
@@ -0,0 +1,95 @@
...
+   * otherwise only carriers that have been registered are supported.
+   */
+  <T> void inject(Span span, T carrier);
```
bensigelman: How does an impl signal that a carrier is unsupported? (I.e., should we define some standard exception types?)

opentracing/src/main/java/opentracing/Tracer.java
```diff
@@ -0,0 +1,95 @@
...
+   *
+   * If the carrier object has no such span stored within it, a new Span is created.
+   *
+   * Unless there’s an error, it returns a SpanBuilder.
+   * The Span generated from the builder can be used in the host process like any other.
+   *
+   * (Note that some OpenTracing implementations consider the Spans on either side of an RPC to have the same identity,
+   * and others consider the caller to be the parent and the receiver to be the child)
+   */
+  <T> SpanBuilder join(T carrier);
```
bensigelman: On the exception front, there are a few possibilities for join()... here's what happened in Python: https://github.com/opentracing/opentracing-python/blob/master/opentracing/propagation.py#L24

> ```
> class UnsupportedFormatException(Exception):
>     """UnsupportedFormatException should be used when the provided format
>     value is unknown or disallowed by the Tracer.
>     See Tracer.inject() and Tracer.join().
>     """
>     pass
> 
> 
> class InvalidCarrierException(Exception):
>     """InvalidCarrierException should be used when the provided carrier
>     instance does not match what the `format` argument requires.
>     See Tracer.inject() and Tracer.join().
>     """
>     pass
> 
> 
> class TraceCorruptedException(Exception):
>     """TraceCorruptedException should be used when the underlynig trace state
>     is seemingly present but not well-formed.
>     See Tracer.inject() and Tracer.join().
>     """
>     pass
> ```

bensigelman: (InvalidCarrierException is N/A here given the lack of a format argument)

bensigelman:
I made a bunch of comments... anything that requires *new* code (as opposed to changes) I am fine doing in a follow-on; it would be helpful to have something in `master` at this point.

opentracing/src/main/java/opentracing/Span.java
adriancole: >
> I am imagining – perhaps incorrectly – that the caller of the explicit-timestamp version has a raw timestamp rather than a duration on hand (i.e., that they are basically doing log processing / translation). FWIW, I think that tracing *impls* are better off tracking duration since it's not subject to clock-skew cruelty.
>
> This isn't just for log scraping. this is for any case when OpenTracing
isn't used to create events. Many events are structured and have both start
and endTs, but duration is often recorded separately for more precision.

The problem is that timestamps in java are measured in millis accuracy. To
accurately align nanos ticks with current time is non-trivial and rarely
implemented. duration practice is to measure with nanos ticks. That's the
rationale anyway, and why zipkin/brave has duration.

Anyway it is a different topic, in OT status quo, endTs sounds like the
right choice?

adriancole: also, I think it may have been you that mentioned that the risk of
time moving backwards is an issue when doing timestamp based intervals
vs absolute + duration offset (back in the day)

I don't want this conversation to hold up mick's work as it is a
different topic which spun out of me enumerating how this is done
today in Brave. If you feel this is important enough to continue,
probably a top-level issue is best.

bensigelman: I definitely don't feel strongly about finish-time vs duration... I'm making an educated guess about what's most convenient most of the time, but if others feel strongly, I do not :)

michaelsembwever: i'm leaving the addition of `finish(timestamp)` as outside the scope of this PR.

adriancole: Fair enough

opentracing/src/main/java/opentracing/Tracer.java
michaelsembwever: These required carriers are in https://github.com/opentracing/opentracing-java/pull/11

The non-interoperable i presume you mean is at the "format" level.
I don't see this is a problem as long as it's clearly documented with java that these are the required carriers and are otherwise interoperable.

opentracing/src/main/java/opentracing/Tracer.java
michaelsembwever: fixed

opentracing/src/main/java/opentracing/Tracer.java
michaelsembwever: IllegalStateException. Since this is a programmatic error.

opentracing/src/main/java/opentracing/Tracer.java
michaelsembwever: InvalidCarrierException --> IllegalStateexception 
TraceCorruptedException --> IllegalArgumentException

The latter can be argued that it should be a checked exception.
I'd like to wait and see what experience shows first.

will add to apidoc.

opentracing/src/main/java/opentracing/Span.java
michaelsembwever: removed.

opentracing/src/main/java/opentracing/Tracer.java
michaelsembwever: ok, have changed code to `withStartTimestamp(..)`

michaelsembwever:
All points addressed.
If no further objections will merge to master in 24hrs.

opentracing/src/main/java/opentracing/Tracer.java
bensigelman: I'd suggest leaving out the comment about required carriers fro now... if the required carriers in Java don't have the same structural shape as the required carriers in other languages, I don't see how an OT implementation that is intended to work for both Java *and* other OT platforms can `inject()` in Java using `Map<String, String>` and `join()` in some other platform that uses `SplitTextCarrier` (which has not one but two maps within).

I'd rather not try to spend our time debating this over github, though, as https://github.com/opentracing/opentracing.github.io/issues/73 might make one or both of our respective points of view moot (probably mine, actually, as I'm basically arguing for the removal of the "Split" part of "Split*Carrier").

bensigelman:
@michaelsembwever I would prefer that we drop the comments about the required carriers before merging into `master`; but since [the required carriers are up in the air right now](https://github.com/opentracing/opentracing.github.io/issues/73), we'll have to revisit them (the comments) soon anyway.

opentracing/src/main/java/opentracing/Tracer.java
michaelsembwever: Good point.

Both Text and Binary Carrier need a defined serialisation specification.
I've raised this before in the need for an Internal Specification, that implementors need to adhere to.

I'm not in favour of the two maps approach, it just seems clumsy and presumptuous that the underlying rpc/http framework can put maps inside of maps.

The solution of state and baggage prefixes used on keys all within the one map makes more sense to me. It seems to be a solution that will work across languages and across rpc/http frameworks best.

I've removed the comment on required carriers.

bensigelman:
(LGTM!)

PR #16, opened 2016-03-17
add noop, and rename package

oibe:
- rename packages to be io.opentracing
- add noop Tracer, Span, and SpanBuilder

### DIFF
--- a/opentracing/src/main/java/io/opentracing/noop/NoopSpan.java
+++ b/opentracing/src/main/java/io/opentracing/noop/NoopSpan.java
@@ -1,0 +1,62 @@
+/**
+ * Copyright 2016 The OpenTracing Authors
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License"); you may not use this file except
+ * in compliance with the License. You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software distributed under the License
+ * is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express
+ * or implied. See the License for the specific language governing permissions and limitations under
+ * the License.
+ */
+package io.opentracing.noop;
+
+import io.opentracing.Span;
+
+public class NoopSpan implements Span {
+
+    private NoopSpan defaultSpan = new NoopSpan();
+    private String emptyString = "";
+
+    @Override
+    public void finish() {
+
+    }
+
+    @Override
+    public Span setTag(String key, String value) {
+        return defaultSpan;
+    }
+
+    @Override
+    public Span setTag(String key, boolean value) {
+        return defaultSpan;
+    }
+
+    @Override
+    public Span setTag(String key, Number value) {
+        return defaultSpan;
+    }
+
+    @Override
+    public Span setBaggageItem(String key, String value) {
+        return defaultSpan;
+    }
+
+    @Override
+    public String getBaggageItem(String key) {
+        return emptyString;
+    }
+
+    @Override
+    public Span log(String eventName, Object payload) {
+        return defaultSpan;
+    }
+
+    @Override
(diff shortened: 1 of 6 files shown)


yurishkuro:
LGTM

@bensigelman @michaelsembwever

bensigelman:
Sorry, I thought I saw this get merged but was apparently incorrect!

@michaelsembwever I assume this is uncontroversial...

opentracing/src/main/java/io/opentracing/noop/NoopSpan.java
```diff
@@ -0,0 +1,62 @@
...
+ * Unless required by applicable law or agreed to in writing, software distributed under the License
+ * is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express
+ * or implied. See the License for the specific language governing permissions and limitations under
+ * the License.
+ */
+package io.opentracing.noop;
+
+import io.opentracing.Span;
+
+public class NoopSpan implements Span {
```
michaelsembwever: [trivial] why is this public? and why isn't it final?

opentracing/src/main/java/io/opentracing/noop/NoopSpanBuilder.java
```diff
@@ -0,0 +1,63 @@
...
+ * or implied. See the License for the specific language governing permissions and limitations under
+ * the License.
+ */
+package io.opentracing.noop;
+
+
+import io.opentracing.Span;
+import io.opentracing.Tracer;
+
+public class NoopSpanBuilder implements Tracer.SpanBuilder {
```
michaelsembwever: [trivial] same question. why public? why not final?

opentracing/src/main/java/io/opentracing/noop/NoopSpan.java
```diff
@@ -0,0 +1,62 @@
...
+
+    private NoopSpan defaultSpan = new NoopSpan();
```
michaelsembwever: why not make this `static final` ?

opentracing/src/main/java/io/opentracing/noop/NoopSpanBuilder.java
```diff
@@ -0,0 +1,63 @@
...
+    private Tracer.SpanBuilder defaultSpanBuilder = new NoopSpanBuilder();
+    private Span defaultSpan = new NoopSpan();
```
michaelsembwever: It is odd that there's two NoopSpan instances created in the JVM. 

A method on the NoopSpan will return a different instance to that returned from the builder.
I'd rather see NoopSpan.DEFAULT_SPAN make package-protected, and it used in the builder here.

opentracing/src/main/java/io/opentracing/noop/NoopTracer.java
```diff
@@ -0,0 +1,35 @@
...
+ * the License.
+ */
+package io.opentracing.noop;
+
+import io.opentracing.Span;
+import io.opentracing.Tracer;
+
+public class NoopTracer implements Tracer {
+
+    private SpanBuilder defaultSpanBuilder = new NoopSpanBuilder();
```
michaelsembwever: again why not `static final` ?

opentracing/src/main/java/io/opentracing/noop/NoopTracer.java
```diff
@@ -0,0 +1,35 @@
...
+ * is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express
+ * or implied. See the License for the specific language governing permissions and limitations under
+ * the License.
+ */
+package io.opentracing.noop;
+
+import io.opentracing.Span;
+import io.opentracing.Tracer;
+
+public class NoopTracer implements Tracer {
```
michaelsembwever: should be a `final` class.

yurishkuro:
@michaelsembwever good points all around. 

@oibe let's address them.

PR #18, opened 2016-03-23
Add tracer function to span interface

oibe:
- added a tracer function to span interface
- added a tracer function to NoopSpan
- added a private constructor to NoopTracer

### DIFF
--- a/opentracing/src/main/java/io/opentracing/Span.java
+++ b/opentracing/src/main/java/io/opentracing/Span.java
@@ -73,4 +73,9 @@ public interface Span {
    * The timestamp in microseconds in UTC time.
    **/
   Span log(long timestampMicroseconds, String eventName, /* @Nullable */ Object payload);
+
+  /**
+   *  A method to get a reference to a tracer stored inside a span.
+   **/
+  Tracer tracer();
 }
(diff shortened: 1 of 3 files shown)


michaelsembwever:
I would rather not add this. It adds an unneeded complexity IMHO.

Can you provide examples are why this is needed @oibe ?

oibe:
It might be useful in the case where you want to create a child span, but are deeper in the stack.  If i'm inside of a function that hooks into some framework, and I want to use a span to measure the execution time of a nested call then I need to create a child span.  I have access to the span through thread locals, but I don't have the ability to create a child span because I need the tracer to do that.

michaelsembwever:
> It might be useful …

We really don't want to add complexity to the code prematurely because of might-be's.

The Tracer object will be part of the application control layer, and will be injected or available via lookup (singleton pattern) in most java applications. It's also thread-safe so does not need to be concerned with thread-locals as so forth. For this reason I'd rather hold off on such an addition until the demand for it is clearer.

bensigelman:
I think we will either need `Span.tracer()` or we will basically force OT users to use a `Tracer` singleton. Getting all of the DI stuff to line up properly is possible in theory but will be quite difficult in practice.

I am ok waiting for someone to need `Span.tracer()`, though... @oibe it sounds like this is still in hypothetical territory for Java specifically (I realize that we have this functionality in Go/python already)?

PR #20, opened 2016-03-24
Example helping implementation of the top-level API

michaelsembwever:
( this is the successor to https://github.com/opentracing/opentracing-java/pull/11 )

This code provides a basic implementation meeting the requirements that the specification imposes without adding additional API.

The code also provides that basic implementation in an non-tracer specific way.


michaelsembwever:
@oibe are you keen to take a look over this?

The pull request will look like a lot has changed, because files have been moved around (so to move them into separate subprojects: there's an api subproject, an basic java7 impl subproject, and then a java8 impl subproject), but you'll find your noop files there, and a new implementation to the specification.

yurishkuro:
@michaelsembwever is there any particular reason making the implementation non-java7?

opentracing-impl-java8/src/main/java/io/opentracing/propagation/AbstractTracer.java
```diff
@@ -0,0 +1,175 @@
...
+            // match second on interfaces
+            for (Class<?> iface : carrierType.getInterfaces()) {
+                if (injectors.containsKey(c)) {
+                    return injectors.get(c);
+                }
+            }
+            throw new AssertionError("no registered injector for " + carrierType.getName());
+        }
+
+        public <T> Extractor<T> getExtractor(Class<T> carrierType) {
```
yurishkuro: impl is identical to `getInjector`, aside from return type, could merge into one `lookup(map, carrierType)`

opentracing-impl-java8/src/main/java/io/opentracing/propagation/AbstractTracer.java
```diff
@@ -0,0 +1,175 @@
...
+    }
+
+    public <T> Extractor<T> register(Class<T> carrierType, Extractor<T> extractor) {
+        return registry.register(carrierType, extractor);
+    }
+
+    private static class PropagationRegistry {
+
+        private final ConcurrentMap<Class, Injector> injectors = new ConcurrentHashMap<>();
+        private final ConcurrentMap<Class, Extractor> extractors = new ConcurrentHashMap<>();
```
yurishkuro: I would expect registration to happen at initialization time, so concurrent map is unnecessary locking overhead.

opentracing-impl-java8/src/main/java/io/opentracing/propagation/AbstractTracer.java
```diff
@@ -0,0 +1,175 @@
...
+            return extractors.putIfAbsent(carrierType, extractor);
+        }
+    }
+
+    static class SpanBuilderImpl implements SpanBuilder {
+
+        SpanBuilderImpl() {}
+
+        final Map<String,String> traceState = new HashMap<>();
+        final Map<String,String> traceAttributes = new HashMap<>();
```
yurishkuro: rename "baggage", to be consistent?

opentracing-impl-java8/src/main/java/io/opentracing/propagation/AbstractTracer.java
```diff
@@ -0,0 +1,175 @@
...
+        public <T> Extractor<T> register(Class<T> carrierType, Extractor<T> extractor) {
+            return extractors.putIfAbsent(carrierType, extractor);
+        }
+    }
+
+    static class SpanBuilderImpl implements SpanBuilder {
+
+        SpanBuilderImpl() {}
+
+        final Map<String,String> traceState = new HashMap<>();
```
yurishkuro: how about making this into an abstract type instead?  Forcing everyone to convert trace state into strings seems like unnecessary overhead.

opentracing-impl-java8/src/main/java/io/opentracing/propagation/SpanImpl.java
```diff
@@ -0,0 +1,116 @@
...
+import io.opentracing.Span;
+
+final class SpanImpl implements Span {
+
+    private final String operationName;
+    private final Optional<Span> parent;
+    private final Instant start;
+    private Duration duration;
+    final Map<String,String> traceState = new HashMap<>();
+    final Map<String,String> traceAttributes = new HashMap<>();
```
yurishkuro: rename "baggage"

opentracing-impl-java8/src/main/java/io/opentracing/propagation/AbstractTracer.java
```diff
@@ -0,0 +1,175 @@
...
+        final Map<String,String> traceAttributes = new HashMap<>();
+
+        private String operationName = null;
+        private Span parent = null;
+        private Instant start = Instant.now();
+        private final Map<String,String> stringTags = new HashMap<>();
+        private final Map<String,Boolean> booleanTags = new HashMap<>();
+        private final Map<String,Number> numberTags = new HashMap<>();
```
yurishkuro: why keep these separate if they are in a single map in the `SpanImpl`?

opentracing-impl-java8/src/main/java/io/opentracing/propagation/AbstractTracer.java
```diff
@@ -0,0 +1,175 @@
...
+            return this;
+        }
+
+        @Override
+        public Span start() {
+            Span span = new SpanImpl(operationName, Optional.ofNullable(parent), start);
+            stringTags.entrySet().stream().forEach((entry) -> { span.setTag(entry.getKey(), entry.getValue()); });
+            booleanTags.entrySet().stream().forEach((entry) -> { span.setTag(entry.getKey(), entry.getValue()); });
+            numberTags.entrySet().stream().forEach((entry) -> { span.setTag(entry.getKey(), entry.getValue()); });
+            traceState.entrySet().stream().forEach((entry) -> { span.setBaggageItem(entry.getKey(), entry.getValue()); });
```
yurishkuro: why not just pass already constructed maps for tags and baggage to span constructor?

opentracing-impl/src/main/java/io/opentracing/propagation/SplitTextFormatInjector.java
```diff
@@ -0,0 +1,24 @@
...
+package io.opentracing.propagation;
+
+import java.util.Map;
+
+/**
+ *
+ */
+interface SplitTextFormatInjector extends Injector<Map> {
+    static final String TRACE_IDENTIFIERS = "Open-Tracing-Context-Id-";
+    static final String TRACE_ATTRIBUTES = "Open-Tracing-Trace-Tags-";
```
yurishkuro: these constants need to die. We finally got rid of them in Go implementation, let's not introduce them again, they make no sense. We are not dictating which prefix the implementations use for headers. And we've dropped the "split" requirement, so these are not needed for separating state vs. baggage either.

opentracing-impl/src/main/java/io/opentracing/propagation/SplitBinaryFormatExtractor.java
```diff
@@ -0,0 +1,23 @@
...
+ * the License.
+ */
+package io.opentracing.propagation;
+
+import java.nio.ByteBuffer;
+
+/**
+ *
+ */
+interface SplitBinaryFormatExtractor extends Extractor<ByteBuffer> {
```
yurishkuro: "split" concept has been decommissioned.

opentracing-impl/src/main/java/io/opentracing/NoopTracer.java
```diff
@@ -11,16 +11,13 @@
...
-import io.opentracing.Span;
-import io.opentracing.Tracer;
-
-public class NoopTracer implements Tracer {
+public final class NoopTracer implements Tracer {
 
     @Override
     public SpanBuilder buildSpan(String operationName) {
-        return NoopSpanBuilder.Instance;
+        return NoopSpanBuilder.INSTANCE;
```
yurishkuro: I was actually thinking we can get rid of these singletons and use normal member refs:

```java
public NoopTracer() {
    this.span = new NoopSpan(this);
    this.builder = new NoopSpanBuilder(this.span);
}
```

oibe:
@michaelsembwever taking a look now.

opentracing-impl-java8/src/main/java/io/opentracing/propagation/AbstractTracer.java
yurishkuro: In fact, I would prefer to have a Builder pattern for tracer, so that the only way to register serializers is via builder, and in tracer these maps become immutable.

opentracing-impl/src/main/java/io/opentracing/NoopTracer.java
michaelsembwever: These are not singletons so much as a fly-weight implementation.

opentracing-impl-java8/src/main/java/io/opentracing/propagation/AbstractTracer.java
michaelsembwever: I'm not keen to force when or how initialization happens.

I'd be happy to re-visit this with profiling data indicating that the use of ConcurrentMap is hurting performance.

The builder pattern i'm not sold on as that would force a style of configuration. I feel that this side of the code is still very young and I would like to see evolve along with the use-cases.

michaelsembwever:
@yurishkuro 
> is there any particular reason making the implementation non-java7?
the initial code I wrote was Java8.
there was a number of small things, like the use of Instant, that made keep it that way.
but because the issue of supporting Android (Java7) was raised I've created a project structure that can support both.

opentracing-impl-java8/src/main/java/io/opentracing/propagation/AbstractTracer.java
michaelsembwever: Just reusing what SpanImpl already provides.

opentracing-impl-java8/src/main/java/io/opentracing/propagation/AbstractTracer.java
michaelsembwever: Simplifies the way `span.withTags(..)` is subsequently called.

I suspect this code will get redone after some performance profiling. But my initial look through AbstractMap's constructor and maybe it won't. Again I'll be happy to come back to it.

opentracing-impl-java8/src/main/java/io/opentracing/propagation/AbstractTracer.java
michaelsembwever: good spot.

opentracing-impl-java8/src/main/java/io/opentracing/propagation/SpanImpl.java
michaelsembwever: good spot.

opentracing-impl-java8/src/main/java/io/opentracing/propagation/AbstractTracer.java
michaelsembwever: good spot.

opentracing-impl-java8/src/main/java/io/opentracing/propagation/AbstractTracer.java
michaelsembwever: Having them separate keeps the code clearer.

opentracing-impl/src/main/java/io/opentracing/propagation/SplitBinaryFormatExtractor.java
michaelsembwever: done

opentracing-impl/src/main/java/io/opentracing/propagation/SplitTextFormatInjector.java
michaelsembwever: Thanks. I haven't kept up on that front.

How does the Extractor now know which of the headers belong to the Span?

yurishkuro:
> How does the Extractor now know which of the headers belong to the Span?

it may still do it via headers it should know about, but the actual header names must be provided by the concrete implementation of the tracer, since they are going to be different across tracing impls. There are actually many ways of encoding span in the headers:
  * trace state may be encoded in one header (e.g. `uber-trace-id=1:2:3:1`), or in multiple headers (e.g. `B3-Trace-***` in Finagle)
  * N baggage items may be encoded in N headers with a fixed prefix (e.g. `uberctx-`), or in a single header with both keys/values encoded in the value (I recall basictracer-go did)

michaelsembwever:
> it may still do it via headers it should know about, but the actual header names must be provided by the concrete implementation of the tracer,

yes, i took this approach by leaving the decision be contextual so that each concrete tracer implementation injects/configures it. see code https://github.com/opentracing/opentracing-java/pull/20/files#diff-ec94bc6d9b32af57ad8009b28bfbb28eR40

opentracing-impl-java8/src/main/java/io/opentracing/propagation/TextFormatExtractorImpl.java
```diff
@@ -0,0 +1,49 @@
...
+
+final class TextFormatExtractorImpl implements TextFormatExtractor {
+
+    /** How the trace state is serialised and can be identified separately to baggage during deserialisation,
+     * for the tracing implementation being used, is contextual to the extraction process.
+     */
+    interface TraceState {
+        /** Returns true if the key+value belongs in traceState, otherwise it is presumed to be baggage. */
+        boolean isTraceState(String key, Object value);
+    }
```
michaelsembwever: I'm not happy about the "TraceState" interface name. But haven't landed on anything better yet.

opentracing-impl-java8/src/main/java/io/opentracing/propagation/TextFormatExtractorImpl.java
```diff
@@ -0,0 +1,49 @@
...
+        this.traceState = traceState;
+    }
+
+    @Override
+    public Tracer.SpanBuilder join(Map carrier) {
+        AbstractTracer.SpanBuilderImpl builder = new AbstractTracer.SpanBuilderImpl();
+        for (Map.Entry<String, String> entry : ((Map<String, String>) carrier).entrySet()) {
+            if (traceState.isTraceState(entry.getKey(), entry.getValue())) {
+                builder.traceState.put(entry.getKey(), entry.getValue());
+            } else {
```
yurishkuro: it needs to do a similar check for `isBaggage`, because the carrier is allowed to contain more headers than what the sending tracer injected (essentially Map is an abstraction of all HTTP headers from the request, not just tracing related)

michaelsembwever: Can we not impose that only traceState or baggage attributes exist in the carrier supplied?

I don't know if this is within the scope of the OT Specification, but it would work given that TextFormatExtractorImpl is going to be delegated to from a specific Extractor implementation.

It also make sense in that filtering makes more sense as part of the delegation chain, rather than supplied via context into TextFormatExtractorImpl and run there.

michaelsembwever: I've added `isBaggage(..)`, does seem to be the simplest approach.

yurishkuro:
> Can we not impose that only traceState or baggage attributes exist in the carrier supplied?

That would break the current separation of responsibilities. We are assuming that the instrumentor provides the carrier merely as an abstraction of the underlying RPC request, and does not know which "keys" the tracer will insert into that abstraction. It's the tracer's responsibility to know which keys are "theirs". If a tracer does not want to deal with that, it can encode the whole set (state + baggage) into a single key, for example.

Go version has docs about that: https://github.com/opentracing/opentracing-go/blob/master/propagation.go#L94

opentracing-impl/src/main/java/io/opentracing/propagation/BinaryFormatExtractor.java
```diff
@@ -0,0 +1,23 @@
...
+ * the License.
+ */
+package io.opentracing.propagation;
+
+import java.nio.ByteBuffer;
+
+/**
+ *
+ */
+interface BinaryFormatExtractor extends Extractor<ByteBuffer> {
```
oibe: Shouldn't these interfaces be public?

opentracing-api/pom.xml
```diff
@@ -25,8 +25,8 @@
         <version>0.1.0-SNAPSHOT</version>
     </parent>
 
-    <artifactId>opentracing</artifactId>
-    <name>OpenTracing</name>
+    <artifactId>opentracing-api</artifactId>
```
bensigelman: sorry that I'm so ignorant about this -- but can you very briefly educate me about the impact of this change as far as Maven is concerned? I'm guessing that people who depend on the Java OT API (as callers) will list `"opentracing-api"` rather than `"opentracing"`, correct?

bensigelman:
@michaelsembwever is Java8 so much better that it's worth the complexity cost and cognitive overhead of separate impls? (I guess you can see where I'm leaning here :)

opentracing-impl-java8/src/main/java/io/opentracing/propagation/AbstractTracer.java
bensigelman: @michaelsembwever that's a bit of a tautology :)

I don't have strong feelings about it, but FWIW: these are non-trivial methods, and if there's an easy way to factor the non-trivialness into a single method that seems like a win to my brain. YMMV, I guess.

opentracing-impl-java8/src/main/java/io/opentracing/propagation/SpanImpl.java
```diff
@@ -0,0 +1,117 @@
...
+
+    protected Span setTraceState(String key, String value) {
+        traceState.put(key, value);
+        return this;
+    }
+
+    final class LogData {
+            private final Instant time;
+            private final String message;
+            private final Object paylod;
```
bensigelman: ("paylod")

opentracing-impl-java8/src/main/java/io/opentracing/propagation/SpanImpl.java
```diff
@@ -0,0 +1,117 @@
...
+import java.util.List;
+import java.util.Map;
+import java.util.Optional;
+import java.util.concurrent.TimeUnit;
+
+import io.opentracing.Span;
+
+final class SpanImpl implements Span {
+
+    final Map<String,Object> traceState = new HashMap<>();
```
bensigelman: I like a lot of this about this PR, but this is not one of them... I'd rather that we either (a) really commit to a "trace state" representation and make this a less-general PR that introduces `basictracer-java` (i.e., where traceState is explicitly `<trace_id, span_id, sampled>`), OR (b) we pull back the scope a bit and focus on things like the injector/extractor registry.

bensigelman:
@michaelsembwever this comment sums up my thoughts at the moment: https://github.com/opentracing/opentracing-java/pull/20/files#r57466386

I would favor something more like a `basictracer-java` artifact... that's something we want to build anyway, and makes it less likely that we'll try to build something too general too early.

All of that said, it's great to see that all of this can be accomplished without touching the core `opentracing` Java API.

Happy to chat via hangouts or whatever if that's better.

opentracing-api/pom.xml
yurishkuro: yes, but only in their dependencies management file, the package name remains `opentracing` for the API classes. Having this artifact name allows other artifacts downloadable separately, like `opentracing-basictracer`.

yurishkuro:
I share this sentiment. This implementation is too opinionated about a number of things, tracer state representation being one of them, so for example I don't see our in-house tracer using it as an abstract base. If we call it a `basictracer`, it can be a concrete implementation, ideally compatible with the other basictracer imps we have in other languages. And given the fact that the other implementations are already representing tracer state in a (Zipkin-compatible struct)[https://github.com/opentracing/basictracer-go/blob/master/context.go], we have a good opportunity to easily turn basic tracers into Zipkin-compatible implementations (by implementing the corresponding reporters/recorders).

opentracing-impl-java8/src/main/java/io/opentracing/propagation/SpanImpl.java
michaelsembwever: I agree. This code was written in a very "unknown" state.

My aim with this PR is the latter that you describe, (c).
The intention was to put down some initial code, it doesn't need to be final nor complete, and then let it evolve as we implement Brave and other bridges.

So… my thoughts here are to make SpanImpl into AbstractSpan, leaving traceState out as the missing piece. As traceState is the missing piece that we can't presume anything about at this level.

michaelsembwever:
> is Java8 so much better that it's worth the complexity cost and cognitive overhead of separate impls?

Yes, IMHO. While leaving the door open to support Android (if that becomes a requirement).

opentracing-impl/src/main/java/io/opentracing/propagation/BinaryFormatExtractor.java
michaelsembwever: Good spot.

So long as we're settled this is one of the "required formats".

opentracing-impl/src/main/java/io/opentracing/propagation/TextFormatInjector.java
```diff
@@ -0,0 +1,22 @@
...
+ * the License.
+ */
+package io.opentracing.propagation;
+
+import java.util.Map;
+
+/**
+ *
+ */
+interface TextFormatInjector extends Injector<Map> {
```
michaelsembwever: @bensigelman, @yurishkuro,  we had a discussion recently about `Map` vs `TextCarrier` and the benefits of typing explicitly the required formats. Where did we land on that? (where was that discussion?)

bensigelman:
@michaelsembwever supporting Android is definitely a requirement for OT and `basictracer`, wherever that ends up living.

opentracing-impl-java8/src/main/java/io/opentracing/propagation/SpanImpl.java
michaelsembwever: I've pushed the update that 
 - makes SpanBuilder and SpanImpl abstract, so to leave the traceState to be implemented,
 - cleans up the contextual functionality to TextFormatExtractorImpl and TextFormatInjectorImpl, but defining that functionality as abstract in AbstractSpanBuilder and AbstractTracer respectively.

opentracing-api/pom.xml
bensigelman: Yes yes, I just meant in the dep management file...

Given that, I would suggest we keep this as `opentracing` since that is the by-far-most-important-case. Fine with me if we eventually have `opentracing-basictracer` or `opentracing-spi` (etc etc) in addition.

Not a strongly-held opinion in any case. @michaelsembwever thoughts?

michaelsembwever: What I've got here is a common way of naming the artifacts.

Using the `-api` suffix here means the jar file is named opentracing-api.jar.
Users will also want to define the api as a compile dependency and then the chosen implementation, `opentracing-brave`, as a provided dependency. That the api suffix is explicit makes it obvious to why the api artifact is a visibly exposed dependency while the implementation and bridge are not.

bensigelman: @michaelsembwever got it... I have been thinking forward (maybe the word is "fantasizing") about `brave` eventually just supporting OT out of the box, so rather than `opentracing-brave`, users will just list `brave` as their dependency... at that point, plain old `opentracing` makes more sense.

Anyway, let's go with what you have. Just wanted to understand the rationale a bit better.

opentracing-impl/src/main/java/io/opentracing/propagation/BinaryFormatExtractor.java
```diff
@@ -0,0 +1,23 @@
...
+ * the License.
+ */
+package io.opentracing.propagation;
+
+import java.nio.ByteBuffer;
+
+/**
+ *
+ */
+public interface BinaryFormatExtractor extends Extractor<ByteBuffer> {
```
bensigelman: IMO it's going to be hard to factor out the binary format in a meaningful/useful way... i.e., I'd suggest that we just leave it to implementations to register a `Injector<ByteBuffer>` and `Extractor<ByteBuffer>` themselves. Otherwise this layer is just going to be a (useless) passthrough intermediary.

Make sense?

bensigelman:
@michaelsembwever TL;DR In my view, what you have here is at a reasonable stopping/merging point. I am primarily interested in getting `basictracer-java` going... that will have more immediate use (I would personally benefit from it, FWIW) and building it will teach us more about using the abstractions here. Does that seem like a reasonable direction? (To be clear, you're not signing up for the `basictracer-java` work by saying yes :)

Thanks...

opentracing-impl/src/main/java/io/opentracing/propagation/BinaryFormatExtractor.java
michaelsembwever: Unless we want a marker interface around a specific type rather than ByteBuffer...?

michaelsembwever:
> Does that seem like a reasonable direction?

Yes. I've always intended to land just the minimal increment.
Take note that the lastest push changed a bit by removing SpanImpl.traceState and TextFormatExtractorImpl.TraceState.
Also there remains the questions of a) using Map and ByteByffer types, or something specific to the required formats, and b) keeping/removing BinaryFormat* and TextFormat* interfaces.

I'm not entirely sure what `basictracer-java` entails, as opposed for example to `opentracing-brave`. The latter I'd like to take a look at.

bensigelman:
Removing `SpanImpl.traceState` / `TextFormatExtractorImpl.TraceState` is helpful, thanks.

Remaining questions:

**Using Map and ByteBuffer types vs something else:** another possibility would be to mimic what ultimately happened in Go and define separate, as-narrow-as-possible interfaces for the Inject()/Join() (i.e., Write/Read) uses of the given format. E.g., something like this: (cc @yurishkuro)

```java
public class TextMap {
    public interface Injector {
        public void put(String key, String value);
    }
    public interface Joiner {
        // ...
    }
}
```

It is of course trivial/easy to write a wrapper around `Map<String, String>` given the above, which would save people boilerplate like the following:

```java
        final Map<String, String> m = new HashMap<String, String>();
        TextMap.Injector i = new TextMap.Injector() {
            public void put(String key, String val) {
                m.put(key, val);
            }
        };
```

Personally, I like the idea of pursuing something like the above... it gives OT users the option of avoiding unnecessary copies and keeps the Inject/Join interface as narrow as possible. For those who want something easier to program against, they can use whatever adapter we write around `Map<String, String>`. (Of course there is an entirely analogous set of changes for the binary carrier)

All of this said, if we want to pursue this direction it will of course need to happen at the `opentracing-java` layer.

Thoughts??

**BasicTracer vs OT-Brave**: if you want to work on Brave integration, that would be excellent... Maybe we can frontload a quick conversation with kristofa as a heads-up. Someone who knows the `basictracer-go` code should be able to create a Java impl in parallel.

michaelsembwever:
I've pushed a commit that uses a `TextFormat` class instead of a Map.
If this is better I'll squash it in, otherwise I will remove it.

> … define separate, as-narrow-as-possible interfaces for the Inject()/Join()

I don't yet see the benefit of this. My first attempt did it that way but it just means more boilerplate code in java.

michaelsembwever:
My vote is to remove the additional commit and stick with just the Map and ByteBuffer types.
(Also removing the marker interfaces in io.opentracing.propagation).

I feel there's still many things up in the air and it's better that this PR keeps things as simple as possible.

opentracing-impl/src/main/java/io/opentracing/propagation/TextFormat.java
```diff
@@ -0,0 +1,22 @@
...
+ * Unless required by applicable law or agreed to in writing, software distributed under the License
+ * is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express
+ * or implied. See the License for the specific language governing permissions and limitations under
+ * the License.
+ */
+package io.opentracing.propagation;
+
+import java.util.Map;
+
+public interface TextFormat {
```
bensigelman: Personally I like this, especially since Java generics can't cleanly distinguish between `Map<String, String>` and `Map<String, Whatever>` at runtime. That said, for it to be useful to OpenTracing users as a ["required carrier"](http://opentracing.io/propagation/#required-carriers), it should live [somewhere in the core OT-java API](https://github.com/opentracing/opentracing-java/tree/master/opentracing/src/main/java/io/opentracing).

bensigelman: Also, for consistency with other OT platforms, this would be called `TextMap` or perhaps `TextMapCarrier`. The "format" identifies the type of carrier: a const string (or similar) for the other platforms.

bensigelman:
Since we don't have explicit format parameters in Java, I prefer the clarity and self-documentation of `TextFormat` (or `TextMap` per my comment on the commit itself). Some other tidbits are [in my comment](https://github.com/opentracing/opentracing-java/pull/20/commits/1b637122984c3c5ad60a68133d8ec965308bf647#r57611149).

Thanks...

opentracing-impl/src/main/java/io/opentracing/propagation/TextFormat.java
yurishkuro: I would also split it into two interfaces, reader and writer, same as in Go, to reduce interface surface area.

michaelsembwever: > That said, for it to be useful to OpenTracing users as a "required carrier", it should live somewhere in the core OT-java API.

Yes, but the notion of required carriers are not exposed in the public API, despite being exposed in the public Specification.

michaelsembwever:
comments address, TextMapReader and TextMapWriter added, commits squashed to one.

opentracing-impl/src/main/java/io/opentracing/propagation/TextFormatExtractor.java
```diff
@@ -0,0 +1,21 @@
...
+ * is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express
+ * or implied. See the License for the specific language governing permissions and limitations under
+ * the License.
+ */
+package io.opentracing.propagation;
+
+/**
+ *
+ */
+public interface TextFormatExtractor extends Extractor<TextMapReader> {
```
yurishkuro: these specializations don't provide much value imho

yurishkuro:
+1

opentracing-impl/src/main/java/io/opentracing/propagation/TextFormatExtractor.java
michaelsembwever: removed.

bensigelman:
@michaelsembwever 

I previously proposed moving `TextMapWriter/Reader` to the core `opentracing-api` artifact... you responded thusly though I can't find the comment in github (perhaps due to squashed commits or something):

>> That said, for it to be useful to OpenTracing users as a "required carrier", it should live somewhere in the core OT-java API.
>
> Yes, but the notion of required carriers are not exposed in the public API, despite being exposed in the public Specification.

I acknowledge that the `opentracing-api` artifact doesn't presently expose the required carriers, but that is basically a bug/issue... until it does, I can't actually use OT-Java without binding myself to the carrier in a particular impl (like the one here). Compare against [Go](https://github.com/opentracing/opentracing-go/blob/master/propagation.go#L73), [Python](https://github.com/opentracing/opentracing-python/blob/master/opentracing/propagation.py#L78), [Javascript](https://github.com/opentracing/opentracing-javascript/blob/master/src/constants.js#L24). (Analogous things can be found in the PHP and C++ ports which are in-progress)

Anyway, if you want to merge this PR *as an intermediate step* and then move `TextMapWriter`/`TextMapReader` into the `opentracing-api` artifact, I guess that's fine by me, but it seems indirect... I would rather just add the carrier interfaces to `opentracing-api` and rebase this PR on top of them.

Everything else LGTM!

bensigelman:
Ah -- nevermind. When I looked at this earlier I think (??) the file paths were different. It appears that the TextMap stuff is in the `opentracing-api` artifact: my mistake!

:ship: (I might add a nitpick or two in a sec, though)

bensigelman:
Yeah, my only nits would be that:

1. we point to `TextMap` from the `Tracer.inject(...)` comment here: https://github.com/opentracing/opentracing-java/pull/20/files#diff-9f484babc6e66bd62198ce009d9247dcL51
1. we add a TODO somewhere (or a github issue) about a binary counterpart to TextMap(Writer|Reader)

But this looks great.

michaelsembwever:
> 1. we point to TextMap from the Tracer.inject(...) comment here: https://github.com/opentracing/opentracing-java/pull/20/files#diff-9f484babc6e66bd62198ce009d9247dcL51
> 2.  we add a TODO somewhere (or a github issue) about a binary counterpart to TextMap(Writer|Reader)

Both points addressed.

opentracing-api/src/main/java/io/opentracing/Tracer.java
```diff
@@ -47,6 +47,8 @@
    *
    * Attempting to inject to a carrier that has been registered/configured to this Tracer will result in a
    * IllegalStateException.
+   *
+   * All implementations support at minimum the required carriers BinaryWriter and TextMapWriter.
```
bensigelman: great, thank you

bensigelman:
LGTM

michaelsembwever:
If no further objections I'll merge this tomorrow.

opentracing-impl-java8/src/main/java/io/opentracing/propagation/AbstractTracer.java
```diff
@@ -0,0 +1,112 @@
...
+            // match first on concrete classes
+            do {
+                if (injectors.containsKey(c)) {
+                    return injectors.get(c);
+                }
+                c = c.getSuperclass();
+            } while (c != null);
+            // match second on interfaces
+            for (Class<?> iface : carrierType.getInterfaces()) {
+                if (injectors.containsKey(c)) {
```
toaler: Should be:

if (injectors.containsKey(iface)){
  return injectors.get(iface);
}

michaelsembwever: Thanks!

Reminded me that I'd had always been to write unit tests once general agreement on the direction of the code had been reached.

Someone (anyone) really should have pulled me up for trying to submit code without tests :-)

A unit test against AbstractTracer now exists.

opentracing-impl-java8/src/main/java/io/opentracing/propagation/TextFormatInjectorImpl.java
```diff
@@ -0,0 +1,50 @@
...
+    private final AbstractTracer tracer;
+    private boolean baggageEnabled = AbstractTracer.BAGGAGE_ENABLED;
+
+    TextFormatInjectorImpl(AbstractTracer tracer) {
+        this.tracer = tracer;
+    }
+
+    @Override
+    public void inject(Span span, TextMapWriter carrier) {
+        assert span instanceof AbstractSpan;
```
toaler: Don't we want Injector's to be generic enough to work with any Span implementation?  By casting Span to a specific implementation it seems like we are limiting the usability of this Injector...

michaelsembwever: The abstract classes in this package are intended to be used together.

But… the reason for the assert constraint no longer exists, so I've removed it.

PR #21, opened 2016-03-25
Make Span Autocloseable

toaler:
Updated interface to make it possible for Span to be used in try-with-resource blocks to reduce some of the standard boilerplate code that's required to ensure that the finish method is called.  Treating Span as a resource allows for close (previously finish) to be closed in an automated way.

### DIFF
--- a/opentracing/src/main/java/io/opentracing/Span.java
+++ b/opentracing/src/main/java/io/opentracing/Span.java
@@ -18,7 +18,7 @@ package io.opentracing;
  *
  * <p>Spans are created by the {@link Tracer#buildSpan} interface.
  */
-public interface Span {
+public interface Span extends AutoCloseable {
 
   /**
    * Sets the end timestamp and records the span.
@@ -26,7 +26,8 @@ public interface Span {
    * <p>This should be the last call made to any span instance, and to do otherwise leads to
    * undefined behavior.
    */
-  void finish();
+  @Override
+  void close();
 
   /**
    * Set a key:value tag on the Span.
--- a/opentracing/src/main/java/io/opentracing/Span.java
+++ b/opentracing/src/main/java/io/opentracing/Span.java
@@ -28,6 +28,8 @@ public interface Span extends AutoCloseable {
    * <p>This should be the last call made to any span instance, and to do otherwise leads to
    * undefined behavior.
    */
+  void finish();
+  
   @Override
   void close();
 
--- a/opentracing/src/main/java/io/opentracing/Span.java
+++ b/opentracing/src/main/java/io/opentracing/Span.java
@@ -13,6 +13,8 @@
  */
 package io.opentracing;
 
+import java.util.Set;
+
 /**
  * Represents an in-flight span in the opentracing system.
  *
@@ -53,6 +55,14 @@ public interface Span extends AutoCloseable {
    * Returns null if no entry found, or baggage is not supported in the current implementation.
    */
   String getBaggageItem(String key);
+  
+  /**
+   * Useful for acquiring the set of baggage keys so that clients of the Span have the ability
+   * to access the corresponding baggage value.
+   * 
+   * @return {@link Set} of baggage keys
+   */
+  Set<String> getBaggageKeys();
 
   /**
    * Add a new log event to the Span, accepting an event name string and an optional structured payload argument.
--- a/opentracing/src/main/java/io/opentracing/Tracer.java
+++ b/opentracing/src/main/java/io/opentracing/Tracer.java
@@ -93,10 +93,10 @@ public interface Tracer {
       /** Same as {@link Span#setTag(String, Number)}, but for the span being built. */
       SpanBuilder withTag(String key, Number value);
 
-      /** Same as {@link Span#setBaggageItem(String, String)}, but for the span being built. */
+      /** Specify a timestamp of when the Span was started, represented in microseconds since epoch. */
       SpanBuilder withStartTimestamp(long microseconds);
 
-      /** Specify a timestamp of when the Span was started, represented in microseconds since epoch. */
+      /** Same as {@link Span#setBaggageItem(String, String)}, but for the span being built. */
       SpanBuilder withBaggage(String key, String value);
 
       /** Returns the started Span. */
--- a/opentracing/src/main/java/io/opentracing/Tracer.java
+++ b/opentracing/src/main/java/io/opentracing/Tracer.java
@@ -87,14 +87,17 @@ public interface Tracer {
       /** Same as {@link Span#setTag(String, String)}, but for the span being built. */
       SpanBuilder withTag(String key, String value);
 
-      /** Same as {@link Span#setTag(String, String)}, but for the span being built. */
+      /** Same as {@link Span#setTag(String, boolean)}, but for the span being built. */
       SpanBuilder withTag(String key, boolean value);
 
-      /** Same as {@link Span#setTag(String, String)}, but for the span being built. */
+      /** Same as {@link Span#setTag(String, Number)}, but for the span being built. */
       SpanBuilder withTag(String key, Number value);
 
-      /** Specify a timestamp of when the Span was started, represented in microseconds since epoch. */
+      /** Same as {@link Span#setBaggageItem(String, String)}, but for the span being built. */
       SpanBuilder withStartTimestamp(long microseconds);
+
+      /** Specify a timestamp of when the Span was started, represented in microseconds since epoch. */
+      SpanBuilder withBaggage(String key, String value);
 
       /** Returns the started Span. */
       Span start();
--- a/opentracing/src/main/java/io/opentracing/noop/NoopSpan.java
+++ b/opentracing/src/main/java/io/opentracing/noop/NoopSpan.java
@@ -13,6 +13,9 @@
  */
 package io.opentracing.noop;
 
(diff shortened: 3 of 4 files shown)


opentracing/src/main/java/io/opentracing/Span.java
```diff
@@ -26,7 +28,8 @@
    * <p>This should be the last call made to any span instance, and to do otherwise leads to
    * undefined behavior.
    */
-  void finish();
```
michaelsembwever: You can't remove this method as it implements the [Specification](http://opentracing.io/spec/#the-span-interface).

`close()` can delegate to it though.

toaler: No problem will add back in.

opentracing/src/main/java/io/opentracing/Tracer.java
```diff
@@ -87,15 +87,18 @@
...
-      /** Same as {@link Span#setTag(String, String)}, but for the span being built. */
+      /** Same as {@link Span#setTag(String, boolean)}, but for the span being built. */
       SpanBuilder withTag(String key, boolean value);
 
-      /** Same as {@link Span#setTag(String, String)}, but for the span being built. */
+      /** Same as {@link Span#setTag(String, Number)}, but for the span being built. */
       SpanBuilder withTag(String key, Number value);
 
-      /** Specify a timestamp of when the Span was started, represented in microseconds since epoch. */
+      /** Same as {@link Span#setBaggageItem(String, String)}, but for the span being built. */
```
michaelsembwever: the original apidoc made more sense, and held more information.

opentracing/src/main/java/io/opentracing/noop/NoopSpan.java
```diff
@@ -60,4 +63,9 @@ public Span log(String eventName, Object payload) {
     public Span log(long timestampMicroseconds, String eventName, Object payload) {
         return this;
     }
+
+	@Override
+	public Set<String> getBaggageKeys() {
+		return Collections.emptySet();
+	}
```
michaelsembwever: Do you have an example of when and why this would be used?
I'm unsure of a situation where you'd need to *explore* what baggage existed.

opentracing/src/main/java/io/opentracing/Tracer.java
toaler: Line 96 needs to be swapped with 99, will resubmit.

Line 90, 93 were incorrect.

opentracing/src/main/java/io/opentracing/Tracer.java
```diff
@@ -87,15 +87,18 @@
...
       SpanBuilder withStartTimestamp(long microseconds);
 
+      /** Specify a timestamp of when the Span was started, represented in microseconds since epoch. */
+      SpanBuilder withBaggage(String key, String value);
```
michaelsembwever: It would be `withBaggageItem(..)`.

Do we need this as part of the builder? The baggage state will never influence the creation.

opentracing/src/main/java/io/opentracing/Span.java
```diff
@@ -52,6 +57,14 @@
...
    */
   String getBaggageItem(String key);
+  
+  /**
+   * Useful for acquiring the set of baggage keys so that clients of the Span have the ability
+   * to access the corresponding baggage value.
+   * 
+   * @return {@link Set} of baggage keys
+   */
+  Set<String> getBaggageKeys();
```
michaelsembwever: Do you have an example of when and why this would be used?
I'm unsure of a situation where you'd need to explore what baggage existed.

michaelsembwever:
Using `AutoClosable` is a good idea.

Do you mind if we wait until https://github.com/opentracing/opentracing-java/pull/20 is agreed upon before we finally merge this? #20 contains project refactoring that easily throws up conflicts when rebasing it.

opentracing/src/main/java/io/opentracing/Span.java
toaler: In Tracer.inject, I want to enumerate through the existing key/value pairs and push them into the corresponding carrier without access internal state of the Span implementation.  Want to avoid this:

span.baggage.entrySet()

for (String key : span.getBaggageKeys()) {
  carrier.put(key, span.getBaggageItem(value));
}

I want to keep inject code as generic as possible and not code against a span implementation.

toaler:
I can wait for #20 I have added it in my local implementation which I'm coding my implementation of Span against.

opentracing/src/main/java/io/opentracing/Span.java
```diff
@@ -27,6 +29,9 @@
    * undefined behavior.
    */
   void finish();
```
toaler: Added finish back

opentracing/src/main/java/io/opentracing/Tracer.java
```diff
@@ -87,15 +87,18 @@
...
-      /** Same as {@link Span#setTag(String, String)}, but for the span being built. */
+      /** Same as {@link Span#setTag(String, boolean)}, but for the span being built. */
       SpanBuilder withTag(String key, boolean value);
 
-      /** Same as {@link Span#setTag(String, String)}, but for the span being built. */
+      /** Same as {@link Span#setTag(String, Number)}, but for the span being built. */
       SpanBuilder withTag(String key, Number value);
 
       /** Specify a timestamp of when the Span was started, represented in microseconds since epoch. */
       SpanBuilder withStartTimestamp(long microseconds);
```
toaler: Corrected comments

opentracing/src/main/java/io/opentracing/Span.java
michaelsembwever: In `Tracer.inject(..)` you're then working with the API at a different level.

It is expected to work more with the internal state.

This internal state is still new, available in https://github.com/opentracing/opentracing-java/pull/20, and it would be appreciated if your efforts could be used in testing and evolving it.

toaler: Ok I'll hang tight until #20 is available.  When do you think that will be?

opentracing/src/main/java/io/opentracing/Tracer.java
```diff
@@ -87,15 +87,18 @@
...
 
+      /** Same as {@link Span#setBaggageItem(String, String)}, but for the span being built. */
+      SpanBuilder withBaggage(String key, String value);
```
michaelsembwever: Do we need this as part of the builder? The baggage state will never influence the creation.

toaler: I thought the whole point of baggage was to be able to propagate state across RPC's.  So my assumption was baggage needs to be read from the RPC handler on the receiving service and added to the new span being created, so that on the next RPC baggage can continue to be passed along.

michaelsembwever: Yes that is correct.

I've done that [here](https://github.com/opentracing/opentracing-java/pull/20/files#diff-ec94bc6d9b32af57ad8009b28bfbb28eR45)

yurishkuro: > So my assumption was baggage needs to be read from the RPC handler

I don't think this is correct. Yes, the baggage needs to be read, but by the **Tracer** (or its Extractor implementing `join` method), not the the end user of the API.

So `-1` to this method, it widens the API unnecessarily. If this is a new baggage the instrumenter is adding to the span, then it can be added with `span.setBaggageItem()` after the span has been constructed.

yurishkuro: BTW, the reason the same argument does not apply to `withTag` is because tags are used as a mechanism to influence span behavior (e.g. to increase sampling priority), so sometimes it's necessary to be able to set the tags *before* the span is constructed. We have not extended the same concept to baggage.

opentracing/src/main/java/io/opentracing/Span.java
bensigelman: @toaler @michaelsembwever 

An *implementation* of `Tracer.inject` will be free to cast the `Span` instance to the actual non-interface type, at which point gaining access to the baggage is trivial. I.e., this shouldn't be needed in the `Span` interface, regardless of #20.

opentracing/src/main/java/io/opentracing/Tracer.java
bensigelman: btw, I have it on my list (we'll see if I get to it) to write down everything a person could want to know or understand about inject/join... from the perspective of a caller (in an RPC/IPC situation, that is), from the perspective of an implementor, and why various things are the way they are. Right now it's not written down and thus there is always confusion as people spin up.

More specific to this particular thread, though: @yurishkuro is correct that the `Tracer` will handle baggage parsing.

bensigelman: @toaler, I just wrote a more detailed document about cross-process propagation (i.e., Inject and Join) here: http://opentracing.io/propagation/

I think that should hopefully be an incremental improvement in terms of understanding, although there is still more work to be done there (in particular, there ought to be more detail about *implementing* Inject and Join).

michaelsembwever:
implemented in https://github.com/opentracing/opentracing-java/commit/01025fc8c0beab93484420c333e43f014da60d2b

PR #23, opened 2016-03-28
Work In Progress – OpenTracing Brave Bridge

michaelsembwever:

@adriancole @kristofa

### DIFF
--- a/opentracing-brave/pom.xml
+++ b/opentracing-brave/pom.xml
@@ -1,0 +1,102 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!--
+
+    Copyright 2016 The OpenTracing Authors
+
+    Licensed under the Apache License, Version 2.0 (the "License"); you may not use this file except
+    in compliance with the License. You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+    Unless required by applicable law or agreed to in writing, software distributed under the License
+    is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express
+    or implied. See the License for the specific language governing permissions and limitations under
+    the License.
+
+-->
+<project xmlns="http://maven.apache.org/POM/4.0.0"
+        xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+        xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
+    <modelVersion>4.0.0</modelVersion>
+
+    <parent>
+        <groupId>io.opentracing</groupId>
+        <artifactId>parent</artifactId>
+        <version>0.1.0-SNAPSHOT</version>
+    </parent>
+
+    <artifactId>opentracing-brave</artifactId>
+    <name>OpenTracing-brave</name>
+    <description>OpenTracing Brave bridge</description>
+
+    <properties>
+        <main.basedir>${project.basedir}/..</main.basedir>
+        <animal-sniffer-maven-plugin.version>1.14</animal-sniffer-maven-plugin.version>
+    </properties>
+
+    <dependencies>
+        <dependency>
+            <groupId>io.opentracing</groupId>
+            <artifactId>opentracing-impl-java8</artifactId>
+            <version>0.1.0-SNAPSHOT</version>
+        </dependency>
+        <dependency>
+            <groupId>com.github.kristofa</groupId>
+            <artifactId>brave-core</artifactId>
+            <version>3.5.0</version>
+        </dependency>
+        <dependency>
+            <groupId>com.github.kristofa</groupId>
+            <artifactId>brave-http</artifactId>
+            <version>3.5.0</version>
+        </dependency>
+    </dependencies>
+
+    <build>
+        <plugins>
+            <!-- Ensure main source tree compiles to Java 7 bytecode. This allows higher
+                 reuse, including android, which doesn't support Java 8 features. -->
+            <plugin>
+                <inherited>true</inherited>
+                <artifactId>maven-compiler-plugin</artifactId>
+                <configuration>
+                    <source>1.8</source>
+                    <target>1.8</target>
+                </configuration>
+                <executions>
+                    <execution>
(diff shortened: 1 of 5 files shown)


opentracing-brave/pom.xml
```diff
@@ -0,0 +1,98 @@
...
+                </configuration>
+                <executions>
+                    <execution>
+                        <id>default-compile</id>
+                        <phase>compile</phase>
+                        <goals>
+                            <goal>compile</goal>
+                        </goals>
+                        <configuration>
+                            <source>1.8</source>
```
adriancole: :) this doesn't seem to agree with the comment

opentracing-brave/src/main/java/io/opentracing/propagation/BraveSpanImpl.java
```diff
@@ -0,0 +1,75 @@
...
+            Optional<Span> parent,
+            Instant start,
+            Optional<ServerTracer> serverTracer) {
+
+        super(operationName, parent, start);
+        this.brave = brave;
+        this.serverTracer = serverTracer;
+
+        this.spanId = brave.localTracer().startNewSpan(
+                "???",
```
adriancole: empty string is the latest unknown :P

opentracing-brave/src/main/java/io/opentracing/propagation/BraveSpanImpl.java
```diff
@@ -0,0 +1,75 @@
...
+
+final class BraveSpanImpl extends AbstractSpan {
+
+    final SpanId spanId;
+    final Optional<ServerTracer> serverTracer;
+    private final Brave brave;
+    private Optional<ClientTracer> clientTracer = Optional.empty();
+
+    public BraveSpanImpl(
+            Brave brave,
```
adriancole: unused?

opentracing-brave/src/main/java/io/opentracing/propagation/BraveSpanImpl.java
adriancole: oh wait.. not sure which field this is.. if component, then binary annotation "lc"'s value is the component name

opentracing-brave/src/main/java/io/opentracing/propagation/BraveTracerImpl.java
```diff
@@ -0,0 +1,59 @@
...
+    private final Brave brave = new Brave.Builder().build();
+
+    @Override
+    BraveSpanBuilderImpl createSpanBuilder() {
+        return new BraveSpanBuilderImpl(brave);
+    }
+
+    @Override
+    Map<String, String> getTraceState(Span span) {
+        return new HashMap<String,String>() {{
```
adriancole: there's actually 4 possible fields to propagate, including parentId. the sampled flag is an interesting one here

https://github.com/openzipkin/brave/blob/master/brave-http/src/main/java/com/github/kristofa/brave/http/BraveHttpHeaders.java

opentracing-brave/src/main/java/io/opentracing/propagation/BraveTracerImpl.java
```diff
@@ -0,0 +1,59 @@
...
+    }
+
+    @Override
+    Map<String, String> getBaggage(Span span) {
+        return ((BraveSpanImpl)span).baggage;
+    }
+
+    @Override
+    public <T> void inject(Span span, T carrier) {
+        brave.clientTracer().setClientSent();
```
adriancole: interesting.. I didn't consider that inject had side-effects on the span itself.

bensigelman:
@michaelsembwever @adriancole @kristofa just wanted to mention that I have great interest in this PR but know far less about Brave than any of you. That said, please pull me in about the OpenTracing side if it would be helpful.

opentracing-brave/pom.xml
michaelsembwever: fixed

opentracing-brave/src/main/java/io/opentracing/propagation/BraveSpanImpl.java
michaelsembwever: sry, what should the component value be?

opentracing-brave/src/main/java/io/opentracing/propagation/BraveSpanImpl.java
michaelsembwever: fixed. forgot to call `localTracer.finish()` in `finish()`

opentracing-brave/src/main/java/io/opentracing/propagation/BraveSpanImpl.java
adriancole: https://github.com/openzipkin/zipkin-java/blob/master/zipkin/src/main/java/zipkin/Constants.java#L138

ex "nginx" There's no implicit component in brave's local tracer. local
tracer was made for explicit use where the caller knows the component they
are invoking. maybe leave local tracer out, if there's no argument for
component we can hijack?

opentracing-brave/src/main/java/io/opentracing/propagation/BraveTracerImpl.java
michaelsembwever: nice. used BraveHttpHeaders in a few places.

opentracing-brave/src/main/java/io/opentracing/propagation/BraveTracerImpl.java
michaelsembwever: Yeah. I've stuck with using `Brave.localTracer()` by default, and hooking into `Brave.clientTracer()` in `inject(..)` and `Brave.serverTracer()` in `join(..)`.

I'm not sure if this approach works, it's just the simplest solution that comes to mind.

opentracing-brave/src/main/java/io/opentracing/propagation/BraveSpanImpl.java
michaelsembwever: am just using "jvm" for now.

kristofa:
@michaelsembwever Nice, but I'll have to do some reading on the OpenTracing api :)

michaelsembwever:
re-opening this fresh as the destination is now master

PR #25, opened 2016-04-01
OpenTracing Brave Bridge

michaelsembwever:

Continuation of https://github.com/opentracing/opentracing-java/pull/23

@adriancole @kristofa @bensigelman


yurishkuro:
> All spans are created with the local tracer. CS+CR and SR+SS annotations are done around the inject and join methods. Not sure yet whether that approach will work.

FWIW, that's similar to what our tracers are doing, although we do it even later - the send/receive annotations are actually added by the reporter (recorder) before sending the span out of process.

opentracing-brave/src/main/java/io/opentracing/propagation/BraveTracerImpl.java
```diff
@@ -0,0 +1,69 @@
...
+    }
+
+    @Override
+    Map<String, String> getBaggage(Span span) {
+        return ((BraveSpanImpl)span).baggage;
+    }
+
+    @Override
+    public <T> void inject(Span span, T carrier) {
+        brave.clientTracer().setClientSent();
```
kristofa: When you call `brave.clientTracer().setClientSent()` there should have been a call to `brave.clientTracer().startNewSpan(name)` for it to work.

The explicit split-up between client / server / local in brave makes it harder to integrate with OpenTracing.

michaelsembwever: fixed

adriancole:
One thing that became very more obvious to me now than earlier is that while Brave implicitly does in-process propagation, the opentracing api does not.

There may be a way to address this bridge much easier, ex by cherry-picking components from Brave, like the span collector and sampler, vs the Brave instance (which has limitations due to its need to deal with in-process propagation).

maxwindiff:
@michaelsembwever just curious, are you still working on this?

michaelsembwever:
> @michaelsembwever just curious, are you still working on this?

yes. sorry for the gap.

michaelsembwever:
We're back on for review.
 (I still need to do a round of writing unit tests)

michaelsembwever:
Tests are complete. Commit message is final.

opentracing-api/src/main/java/io/opentracing/Tracer.java
```diff
@@ -78,12 +78,14 @@
...
-   * @returns the SpanContext instance extracted from the carrier
+   *
+   * @return the SpanBuilder instance holding context to create a Span.
+   * A noop builder is returned if the carrier held no trace information.
    *
    * @see io.opentracing.propagation.Format
    * @see io.opentracing.propagation.Format.Builtin
    */
-  <C> SpanContext extract(Format<C> format, C carrier);
+  <C> SpanBuilder extract(Format<C> format, C carrier);
```
bensigelman: @michaelsembwever I'm not in favor of this change... all other languages inject and extract SpanContexts... this makes things like multiple-parent specification uniform whether SpanContexts came from the wire or within the process, etc.

yurishkuro: I agree with @bensigelman. Extracting `SpanContext` from a serialized representation does not have to be mixed with span creation. I actually have [use cases](https://github.com/uber/tchannel-go/blob/dev/tracing.go#L227) where doing so is in fact harmful (well, at least wasteful), because I only need the SpanContext. The link is to Go example, but I'll have the exact same pattern in Java.

michaelsembwever: What happens next to the SpanContext?
It is always, to my understanding, in some form or another, toward the creation of a Span.
Hence I made this change because I presumed that the only subsequent action was to then build a span.
Thanks for the use-case @yurishkuro, that was ofc my next question :-)

The counter use-case I have for this is creating the fluent API.
The extract method should return something that fluently follows for instrumentation code.
In this case it returns a NoopSpanBuider.

I'm not sure how a SpanContext alone can do that for you.
To solve @yurishkuro's use-case the SpanBuilder can extend SpanContext.
This prevents the waste that is described, as baggage can be taken from the builder without the creation of a Span. At the same time it provides the fluent api. 

These changes have been pushed.

bensigelman:
@michaelsembwever thanks for the adjustment via subclassing with SpanBuilder. It’s still not the way I would personally do it, but the stakes are 99% lower when we’re not changing the API that OpenTracing callers see and you are the one willing to do the work. :)

As for the larger purpose of this PR, I don’t feel sufficiently familiar with Brave to offer a review you could count on... I'm mainly interested in helping out and/or answering questions about OpenTracing-specific API questions/concerns (per the above).

That said, I did want to mention one thing: there is now an https://github.com/opentracing-contrib github organization that's intended to provide a home for OpenTracing code that doesn't affect the core APIs, and IMO the Brave bridge fits that description perfectly. Would you be open to moving the Brave-specific aspects of this over there before merging? (I realize that some of the changes affect the `-impl` artifacts, and that you'd presumably like to develop all of this in a single branch until the review converges).

michaelsembwever:
> That said, I did want to mention one thing: there is now an https://github.com/opentracing-contrib github organization …

Thanks for raising that.
Yes I will move the brave component out to that once we land this PR.

michaelsembwever:
rebased off master.

opentracing-brave/src/main/java/io/opentracing/BraveSpanImpl.java
```diff
@@ -0,0 +1,78 @@
...
+import com.github.kristofa.brave.ClientTracer;
+import com.github.kristofa.brave.ServerTracer;
+import com.github.kristofa.brave.SpanId;
+
+
+final class BraveSpanImpl extends AbstractSpan {
+
+    final SpanId spanId;
+    final Optional<Span> parent;
+    final Optional<ServerTracer> serverTracer;
```
adriancole: main thought here (following my comment) is that open tracing span's aren't implicitly propagated, while Brave's are. This is why you need to pass around a ServerTracer or ClientTracer (because the span thread-locals are in there). Knowing that anyone using OT has to do propagation themselves, whatever Brave propagation exists is redundant to that.

Long story short, I think the easiest way is to just yank Brave's Sampler and SpanCollector, manually creating spans, as opposed to passing around the server, client or local tracer.

michaelsembwever: The serverTracer and clientTracer are only ever inserted here if this span was involved in or created by calls to `tracer.inject(..)` and `tracer.extract(..)`.

When you call `tracer.inject(..)` you annotate a CS and then when `span.finish()` is called the span must annotate a CR, hence in this situation the `tracer.inject(..)` process gives the span the clientTracer reference.

A similare thing with `tracer.extract(..)`, the SR and SS annotations, and the serverTracer reference here.

I don't yet see how manually creating spans will simplify the code. Maybe an example?

adriancole: sure. I was just looking at this impl vs the one in golang, for example.

What I noticed was that the act of injecting or extracting didn't seem to
have any side-effects on the span (like adding an "cs" annotation, for
example)
https://github.com/openzipkin/zipkin-go-opentracing/blob/master/propagation_ot.go

Most of the zipkin-specific logic ended up in a recorder (which calls a
collector incidentally very similar to brave's)
https://github.com/openzipkin/zipkin-go-opentracing

This is also similar to Uber's (eventhough Uber's doesn't use zipkin
propagation or transports).

Ex. the injector converts to their format (as opposed to B3), without
modifying the span.
https://github.com/uber/jaeger-client-java/blob/master/jaeger-jaxrs2/src/main/java/com/uber/jaeger/propagation/JerseyClientRequestInjector.java#L41

And the reporter converts the span to a zipkin struct (without side-effects
on the input)

https://github.com/uber/jaeger-client-java/blob/194476bdd7caa403db83ac83a1bc08e389d1217c/jaeger-core/src/main/java/com/uber/jaeger/reporters/RemoteReporter.java#L83

In Uber's they use opentracing tags as indicators if something is
client/server etc (you'll notice conversions internal in their span type)
https://github.com/uber/jaeger-client-java/blob/master/jaeger-core/src/main/java/com/uber/jaeger/Span.java

Long story short, I was mainly curious how we wanted to go. The most
opentracing way seems to be isolating zipkin specifics to the propagation
and recorder components, and expecting users to use opentracing tags like
"peer.ipv4" which we'd convert internally to zipkin state.

If still unclear I suppose you are right that writing code would be the
next step towards clarity.

michaelsembwever: > What I noticed was that the act of injecting or extracting didn't seem to have any side-effects on the span …

To be honest I haven't thought about using the brave collector directly.

I've implemented this in mind with the platform being traced is being traced by Zipkin.
Therefore it makes sense to use local/client/server spans as is expected for Zipkin.
For example the platform might not be consistently using OT, but only using it in those components that are also deployed to other platforms that have different tracing backends.

I'm not sure that really contributes much though to the decision of using Brave's porcelain or plumbing api here. Otherwise AFAIK we're sticking with OT idioms here, while implementing Zipkin behaviour.

And I can't see how the idea of using the plumbing (the sampler and collector) isn't going to save any code or complexity here.

@adriancole

adriancole: I think we aren't on the same page, but that shouldn't stop the experiment
from continuing. To show exactly the gaps would require writing behavior
tests and the implementation I mentioned: I don't have time to do that
right now. Besides, there's more than one way to skin the cat, and of
course the whole thing can be iterated.

In other words, carry on. This code is best put in users' hands and all
things don't have to be addressed up front.

michaelsembwever:
With the final added unit test in BraveTracerImplTest i'm happy with this patch and hoping to merge it in 24-48hrs. ( I don't want to have to rebase it yet again :-/ )

Although it is dependant on https://github.com/openzipkin/brave/pull/216 and a subsequent release of Brave.

adriancole:
LGTM I can help extract this into another repo post-merge if folks would find that helpful

michaelsembwever:
This patch is no longer dependant on https://github.com/openzipkin/brave/pull/216 thanks to Adrian explaining a thing or two about Brave and its limitations and future.

It is understood that by implementing against brave's upper APIs (its tracers) that there's an impedance mismatch between OT's explicit state of parent to child span relationships to Brave's implicit state of such relationships.

opentracing-api/src/main/java/io/opentracing/Tracer.java
```diff
@@ -78,15 +78,17 @@
    * @param format the Format of the carrier
    * @param carrier the carrier for the SpanContext state. All Tracer.extract() implementations must support
    *                io.opentracing.propagation.TextMap and java.nio.ByteBuffer.
-   * @returns the SpanContext instance extracted from the carrier
+   *
+   * @return the SpanContext instance holding context to create a Span.
+   * A noop builder is returned if the carrier held no trace information.
```
yurishkuro: This comment does not match the signature.

michaelsembwever: fixed.

Issue #31, opened 2016-07-12
Introduce carrier format identifiers for parity with other platforms


yurishkuro:
This is driven by several concerns:
 1. The implementation to locate the appropriate injectors/extractors based on runtime information of the passed-in carrier type is ugly and inefficient. Having a unique identifier for a format, like `opentracing.TextMap` in Go, makes the implementation as simple as a map lookup, and remove ambiguity of dealing with potentially multiple superclasses and interfaces implemented by a carrier
 1. It is currently impossible to reuse the same carrier for different formats. For example, the `TextMapReader/Writer` interfaces in Go can be used for both plain TextMap format as well as for upcoming HTTPHeader format. It is much easier to define another constant for the format than to duplicate the reader/writer interfaces.
 1. One earlier concern was that having both a format and the carrier creates a possibility of type mismatches. The proposal below addresses that at compile time.

```java
interface TextMapReader {}
interface TextMapWriter {}

interface Format<R, W> {}

public class Formats<R, W> implements Format<R, W> {
    public final static Format<TextMapReader, TextMapWriter> TEXT_MAP = new Formats<>();
    public final static Format<TextMapReader, TextMapWriter> HTTP_HEADER = new Formats<>();
    public final static Format<ByteBuffer, ByteBuffer> BINARY= new Formats<>();
}

interface Span {}
interface Tracer {
    <W> void inject(Span span, Format<?, W> format, W carrier);
    <R> Span extract(Format<R, ?> format, R carrier);
}

public void test(Tracer tracer) {
    Span span = tracer.extract(Formats.TEXT_MAP, new TextMapReader(){});
    tracer.inject(span, Formats.TEXT_MAP, new TextMapWriter(){});
}
```

bensigelman:
(Note that this will be `s/Span/SpanContext/g` soon)

bensigelman:
FWIW, I really like this direction. I'm planning to do an RFC for a `SpanContext` change in Java today and will try to incorporate this direction in the PR.

bensigelman:
Alright, I sent out https://github.com/opentracing/opentracing-java/pull/32 ... comments welcome.

michaelsembwever:
I can see the discussion and project has evolved to SpanContext, and i'm a lot more comfortable with it now.

I'm not sold on the ideas of Formats though (not yet at least).

I disagree with (1) being either ugly and inefficient.
Do we have implementing code that demonstrates this?
Can this be just inexperience with coding such things in java?

[...]

PR #32, opened 2016-07-13
An RFC for SpanContext, References, as well as changes to carriers

bensigelman:
I would love feedback about the API here... once we converge, we can decide what to do about the rest of this (github) repo (the impl helpers, etc).

Please look at https://github.com/opentracing/opentracing-java/issues/31 and https://github.com/opentracing/opentracing.github.io/issues/99 for background and motivation.

I'll add a few PR comments inline as well.


opentracing-api/src/main/java/io/opentracing/BuiltinFormats.java
```diff
@@ -0,0 +1,44 @@
...
+import io.opentracing.propagation.TextMapWriter;
+
+/**
+ * BuiltinFormats encapsulates the inject()/extract() carrier formats that all Tracer implementations must support.
+ *
+ * @see Format
+ * @see io.opentracing.Tracer#inject(SpanContext, Format, Object)
+ * @see io.opentracing.Tracer#extract(Format, Object)
+ */
+public class BuiltinFormats<R, W> implements Format<R, W> {
```
bensigelman: @yurishkuro I went with `BuiltinFormats` instead of `Formats` -- seemed to better emphasize that non-builtin extensions were feasible.

opentracing-api/src/main/java/io/opentracing/BuiltinFormats.java
```diff
@@ -0,0 +1,44 @@
...
+    /**
+     * TEXT_MAP carriers support arbitrary string:string mapping with no special escaping needed on either keys or values.
+     *
+     * The backing store may (or may not) be shared with other non-OpenTracing writers and readers, so it's important for Tracer implementations to use a prefix to identify their own entries in the text map.
```
bensigelman: what do we think about this for non-HTTP TextMap?

opentracing-api/src/main/java/io/opentracing/Format.java
```diff
@@ -11,9 +11,16 @@
...
+/**
+ * Format identifies a reader (R) and Writer (W) pair used by Tracer.inject() and Tracer.extract() to encode/decode SpanContext instances.
+ *
+ * Nearly all OpenTracing use cases are satisfied by the BuiltinFormats, but Format provides extensibility when custom carriers are warranted.
+ *
+ * @see BuiltinFormats
+ * @see io.opentracing.Tracer#inject(SpanContext, Format, Object)
+ * @see io.opentracing.Tracer#extract(Format, Object)
+ */
+public interface Format<R, W> {
```
bensigelman: I guess the builtin formats could also just be public static members inside here...

```
public interface Format<R, W> {
    public final static Format<String, String> TextMap = new Builtins<>();

    class Builtins<R, W> implements Format<R, W> {
    }
}
```

Thoughts?

opentracing-api/src/main/java/io/opentracing/Reference.java
```diff
@@ -0,0 +1,65 @@
...
+ *
+ * References are used by Tracer.buildSpan() to describe the relationships
+ between Spans.
+ */
+public class Reference {
+    private Type type;
+    private SpanContext referee;
+
+    public enum Type {
+        CHILD_OF,
```
bensigelman: I guess we could try harder to use the type system and compiler here... I'm fine with this, though.

opentracing-api/src/main/java/io/opentracing/Span.java
```diff
@@ -19,60 +19,53 @@
...
+    SpanContext context();
 
-  /**
-   * Sets the end timestamp and records the span.
-   *
-   * <p>This should be the last call made to any span instance, and to do otherwise leads to
-   * undefined behavior.
-   */
-  void finish();
+    /**
```
bensigelman: (indenting was off throughout this file)

opentracing-api/src/main/java/io/opentracing/Span.java
```diff
@@ -19,60 +19,53 @@
...
-  Span setTag(String key, Number value);
+    /** Same as {@link #setTag(String, String)}, but for numeric values. */
+    Span setTag(String key, Number value);
 
-  /**
-   * Set a Baggage item, represented as a simple string:string pair.
-   *
-   * Note that newly-set Baggage items are only guaranteed to propagate to future children of the given Span.
-   */
-  Span setBaggageItem(String key, String value);
```
bensigelman: (removed / moved to SpanContext)

opentracing-api/src/main/java/io/opentracing/propagation/BinaryWriter.java
```diff
@@ -1,19 +0,0 @@
...
- *
- * Unless required by applicable law or agreed to in writing, software distributed under the License
- * is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express
- * or implied. See the License for the specific language governing permissions and limitations under
- * the License.
- */
-package io.opentracing.propagation;
-
-/** work in progress */
-public interface BinaryWriter {
```
bensigelman: no longer needed, correct? (if we rely on `ByteBuffer`)

opentracing-api/src/main/java/io/opentracing/BuiltinFormats.java
```diff
@@ -0,0 +1,44 @@
...
+ */
+package io.opentracing;
+
+import java.nio.ByteBuffer;
+import io.opentracing.propagation.TextMapReader;
+import io.opentracing.propagation.TextMapWriter;
+
+/**
+ * BuiltinFormats encapsulates the inject()/extract() carrier formats that all Tracer implementations must support.
+ *
```
adriancole: probably worth it to add docs for these generic params.. like..

```java
 * @param <R> foo used by {@link foo#bar} to bar things
 * @param <W> baz used by {@link baz#qux} to qux things
```

opentracing-api/src/main/java/io/opentracing/BuiltinFormats.java
```diff
@@ -0,0 +1,44 @@
...
+
+    /**
+     * HTTP_HEADER is much like TEXT_MAP, except that keys must be suitable for use as HTTP headers (i.e., must have a limited character set and must be retrieved in a case-insensitive manner) and values should be URL-escaped.
+     */
+    public final static Format<TextMapReader, TextMapWriter> HTTP_HEADER = new BuiltinFormats<>();
+
+    /**
+     * BINARY carriers represent SpanContext state as an opaque byte array controlled entirely by the Tracer implementation.
+     */
+    public final static Format<ByteBuffer, ByteBuffer> BINARY = new BuiltinFormats<>();
```
adriancole: bytebuffer is a fairly crap api (requires book-keeping, flip etc). since these are likely to be small, seems more straight-forward to use `byte[]`

opentracing-api/src/main/java/io/opentracing/Format.java
adriancole: the Builtins thing seems awkward to me. why not just inline the three constants here?

opentracing-api/src/main/java/io/opentracing/Reference.java
```diff
@@ -0,0 +1,65 @@
...
+package io.opentracing;
+
+/**
+ * A Reference pairs a reference type with a SpanContext referee.
+ *
+ * References are used by Tracer.buildSpan() to describe the relationships
+ between Spans.
+ */
+public class Reference {
+    private Type type;
```
adriancole: nit: final classes and field references unless something is intended to be extended, mutated

ex.

```java
public final class Foo {
  private final Type type;

```

opentracing-api/src/main/java/io/opentracing/Reference.java
```diff
@@ -0,0 +1,65 @@
...
+
+    /**
+     * @return the Reference.Type for this reference.
+     */
+    public Type type() { return this.type; }
+
+    /**
+     * @return the SpanContext referred to by this reference.
+     */
+    public SpanContext referee() { return this.referee; }
```
adriancole: referee is a distracting term. it is usually used in sports

opentracing-api/src/main/java/io/opentracing/Tracer.java
```diff
@@ -20,73 +20,80 @@
...
    *
-   * Attempting to inject to a carrier that has been registered/configured to this Tracer will result in a
-   * IllegalStateException.
+   * All implementations must support, at minimum, the BuiltinFormats.
    *
-   * All implementations support at minimum the required carriers BinaryWriter and TextMapWriter.
+   * @see BuiltinFormats
    */
-  <T> void inject(Span span, T carrier);
+  <W> void inject(SpanContext spanContext, Format<?, W> format, W carrier);
```
adriancole: this Format type doesn't seem to be buying us much, as it is only used one-sided. The result looks more awkward than before to me.

opentracing-api/src/main/java/io/opentracing/Tracer.java
```diff
@@ -20,73 +20,80 @@
...
 
   interface SpanBuilder {
 
-      /** Specify the operationName.
-       *
-       * If the operationName has already been set (implicitly or explicitly) an IllegalStateException will be thrown.
+      /**
+       * A shorthand for withReference(Reference.childOf(parent)).
        */
-      SpanBuilder withOperationName(String operationName);
```
adriancole: bye operation name.. you will be missed :P

opentracing-api/src/main/java/io/opentracing/Reference.java
```diff
@@ -0,0 +1,65 @@
...
+ */
+package io.opentracing;
+
+/**
+ * A Reference pairs a reference type with a SpanContext referee.
+ *
+ * References are used by Tracer.buildSpan() to describe the relationships
+ between Spans.
+ */
+public class Reference {
```
adriancole: I don't quite understand why we are pulling out this type. Seems simpler to rename Reference.Type to ReferenceType and squash the rest of this class. We could make one of the referenceTypes default.

If there's some compelling reason for making this object, it is likely a good thing for the doc section.

opentracing-api/src/main/java/io/opentracing/Tracer.java
```diff
@@ -20,73 +20,80 @@
...
     Tracer tracer = ...
 
-    Span feed = tracer.buildSpan("GetFeed")
-                      .start();
+    Span parentSpan = tracer.buildSpan("DoWork")
+                            .start();
 
     Span http = tracer.buildSpan("HandleHTTPRequest")
-                      .withParent(feed)
+                      .withChildOf(parentSpan.context())
```
adriancole: with prefix isn't something I'm in love with. I'd optimize for this reading nicer. ex `asChildOf()` or `withContext(context, referenceType)`

opentracing-api/src/main/java/io/opentracing/Tracer.java
```diff
@@ -20,73 +20,80 @@
...
+      SpanBuilder withChildOf(SpanContext parent);
 
-      /** Specify the parent span
-       *
-       * If the parent has already been set an IllegalStateException will be thrown.
+      /**
+       * Add a reference from the Span being built to a distinct (usually parent) Span. May be called multiple times to represent multiple such References.
        */
-      SpanBuilder withParent(Span parent);
+      SpanBuilder withReference(Reference ref);
```
adriancole: if this can be called multiple times, we should use `add` prefix to make it more obvious

opentracing-api/src/main/java/io/opentracing/Tracer.java
adriancole: the goal seems to be to top-level the difference between writing strings for use in http vs not http. Seems a penalty to have the standard allow both as it forces us to have an api to express a qualifier of a type.

Since `TextMapReader` is a contrived type anyway, how about `HttpHeaderReader`? Then you don't need to special-case and then add this Format thing, which would unravel some of the awkwardness.

opentracing-api/src/main/java/io/opentracing/BuiltinFormats.java
michaelsembwever: I'd really like to see example (s)of a ByteBuffer carrier in play.

My first attempt at implementing one raised a number of questions as to what usefulness any such defined Format and abstract Carrier can actually provide. (My efforts weren't very exhaustive so I won't be surprised to be proven wrong.)

Otherwise I would rather see ByteBuffer used than byte[]. (Sorry @adriancole )

opentracing-api/src/main/java/io/opentracing/Reference.java
michaelsembwever: +1

opentracing-api/src/main/java/io/opentracing/Reference.java
michaelsembwever: +1

opentracing-api/src/main/java/io/opentracing/Tracer.java
michaelsembwever: `asChildOf(..)` is nice.

michaelsembwever:
Without the Format stuff from https://github.com/opentracing/opentracing-java/issues/31 i'm +1 on merging.

With the Format stuff it's currently a -1 from me, unless further discussion helps me see the light.

opentracing-api/src/main/java/io/opentracing/BuiltinFormats.java
adriancole: >
> Otherwise I would rather see ByteBuffer used than byte[]. (Sorry
> @adriancole <https://github.com/adriancole> )
>
ByteBuffer is a crap api, but it is in-the-box and a few libraries use it
natively. Plus you can wrap byte arrays in with ByteBuffer etc. If we have
a compelling case that needed to represent bytes, I'm ok with ByteBuffer
(whine noted)

opentracing-api/src/main/java/io/opentracing/Reference.java
```diff
@@ -0,0 +1,65 @@
...
+    private Type type;
+    private SpanContext referee;
+
+    public enum Type {
```
yurishkuro: I am ok with this being an enum, as a list of _build-in_ reference types, but the `type` field in the `Reference` class needs to be of comparable type, not this enum. Otherwise this is not extensible without OT API release.

opentracing-api/src/main/java/io/opentracing/Tracer.java
yurishkuro: `HttpHeaderReader` is also a qualifier of type, only (a) we need two of those, reader and writer, instead of just one new Format, and (b) it mixes two things, _how_ the data is given to the carrier, and _what_ data can be given.

Cf. https://github.com/opentracing/opentracing-java/issues/31#issuecomment-232394469

opentracing-api/src/main/java/io/opentracing/Format.java
```diff
@@ -11,9 +11,16 @@
...
 }
```
yurishkuro: I suggest adding `bool equals(Format<?, ?> other)` to the interface to force implementations to provide proper `equals` method, since `==` should not be used for formats.

bensigelman:
Meta-comment: I'm going to try and split the (controversial) `Format` stuff out of this PR inasmuch as that's possible. We can then talk about the SpanContext and Reference aspects in isolation. I'll try to make those changes in the next 4 hours or so.

opentracing-api/src/main/java/io/opentracing/BuiltinFormats.java
bensigelman: I didn't see any great options here but don't feel strongly... I ended up going with `ByteBuffer` :-/


In other languages we have learned that it's often beneficial for the carrier to be an interface describing the reading/writing of data rather than being an actual data structure like ByteBuffer or byte[]. This leads me to the only other defensible (and in-the-box) choice: OutputStream/InputStream. Thoughts on that from people who use Java more than I do? (i.e., anyone :)

opentracing-api/src/main/java/io/opentracing/Reference.java
bensigelman: this is a good point, @adriancole... it made sense in other languages that aren't using the builder pattern, but for Java we only need the reference type. I'll push my version of this in a moment.

opentracing-api/src/main/java/io/opentracing/Reference.java
bensigelman: thank you... I <3 nits and would like to be a better Java programmer :)

opentracing-api/src/main/java/io/opentracing/Reference.java
bensigelman: we talked about this briefly off github -- I've gone with `Comparable` where reference types are needed and am using Strings for the builtin ones.

opentracing-api/src/main/java/io/opentracing/Reference.java
bensigelman: what's wrong with sports??? 🏈  🏉 

I changed this. I agree it's a little too cute / wonky.

bensigelman:
Thanks so much for the comments, all, PTAL.

opentracing-api/src/main/java/io/opentracing/Tracer.java
bensigelman: still there, of course, when creating the builder... but another nice thing about the move towards SpanContext (and extract instead of join) is that you basically never need to create a span when you don't already know the operation name.

opentracing-api/src/main/java/io/opentracing/Tracer.java
bensigelman: @adriancole re TextMapReader and HttpHeaderReader: the pattern that's emerging is that some systems (e.g., gRPC) support unrestricted, sane, key:value string metadata maps. And then HTTP is a total mess from a key-mangling standpoint... yet it's obviously vitally important.

So, the gradual consensus has been that HTTP is "special" in that OpenTracing will recognize it with its own carrier (or pair of reader/writer carriers, whatever). But that we should also require Tracers to implement something less annoying like TextMap.

PS: none of this is at all interesting or even relevant if OT users aren't using Baggage.

opentracing-api/src/main/java/io/opentracing/BuiltinFormats.java
yurishkuro: > OutputStream/InputStream

Not a bad choice, but leaves certain questions like "do I have to call `close()` on it when done?"

nio channels have quite minimal API surface, but ironically they do take ByteBuffer(), e.g. [WriteableByteChannel](https://docs.oracle.com/javase/7/docs/api/java/nio/channels/WritableByteChannel.html). And same problem with `close()`.

My Java experience has always been much higher in the stack, so I don't have a strong opinion.

opentracing-api/src/main/java/io/opentracing/Tracer.java
adriancole: The Format thing, which accepts any input and uses `?` as one of the parameters is feigning type safety. There's nothing that binds R and W together when `?` is used, right?

I'm most concerned about how this api is designed, not what it is trying to solve.

I think if you posted to a larger audience, they would agree that this api is tedious, almost scala in java:

```java
<W> void inject(SpanContext spanContext, Format<?, W> format, W carrier);
```

If you need to express a qualifier, you can append it as a parameter, but I wouldn't use a dual-parameter qualifier key knowing one would be ignored. If the only use-case is http, and the only case is baggage, I'd at least consider finding another way.

Most importantly, when designing a new api we should cite other similar designs, especially when they are full of types.

For example, there's a similar concern in dependency injection. They `Key` type below is made to both hold a qualifier and also the type of the instance passed in, (in this case `W`). It has more responsibility than what we are trying to do, yet has a less complex signature. Let's try to simplify!

https://google.github.io/guice/api-docs/latest/javadoc/index.html?com/google/inject/Key.html

opentracing-api/src/main/java/io/opentracing/Tracer.java
adriancole: cool thx

adriancole:
my non-binding opinion is that the type complexity of `Format<R, W>` isn't worth putting into the bottom api. It would end up being wrapped I think.

Complex apis in java usually breed libraries that hide their complexity. If we want to avoid some side-project called OpenTracing-Lite, I'd suggest highly that we simplify as much as possible the front-door of this api.

opentracing-api/src/main/java/io/opentracing/Tracer.java
adriancole: later comment comparing with Key in dependency injection (which store the
type of W and allow it to be qualified by a string, annotation type etc).
basically, my biggest complaint is with the extra types in Format which are
ignored. Also the lack of a basic api (ex that whatever format becomes is a
mandatory parameter).

adriancole:
good idea

opentracing-api/src/main/java/io/opentracing/Tracer.java
michaelsembwever: In addition to @adriancole's concerns around the suitability of this api for java, i'm wondering if we should be adding an inject method rather than replacing the inject method.

That is to have both methods:

    <T> void inject(SpanContext span, T carrier);
    <W> void inject(SpanContext spanContext, …something more complex…);

Maintaining the simple approach for those use-cases where multiple mappings on carriers is not needed.

adriancole: +1 wires crossed, but definitely agree on keeping the old one.

Most apis that i enjoy using only require advanced apis when you are doing
something advanced.

http://square.github.io/moshi/1.x/moshi/

yurishkuro: Keeping both would've been a good compromise if not for the proposed `Format.HTTPHeader` that actually requires the 2-argument version, and is _the_ prevalent use case.

adriancole: depends on how we document it, right?

keeping both could mean defer the decision to impl, or in the case of text
say it will use Format.HTTPHeader, etc

yurishkuro: not sure if documenting will help. The point is that for http header we still want to use the same `TextMapReader/Writer` interface as for unrestricted maps, so we either differentiate by a Format constant, or by duplicating `TextMapReader/Writer`. Duplicating the interfaces just to force a different branch in the tracer feels an anti-pattern to me, quite worse than passing a second Format.X constant as differentiator.

adriancole: I don't expect that you would choose to use the other method, yuri. I'm
speaking about users who are farther away from internals than you are, who
may prefer simpler choices.

yurishkuro: Users, exactly. Right now the users see this method:

```java
void <T> inject(s SpanContext, T carrier)
``` 

Since `T` is unconstrained in any way, it's equivalent to 

```java
void inject(s SpanContext, Object carrier)
``` 

As a user, what am I supposed to pass as `carrier`?

Now consider the case with Format:

```java
void <W> inject(s SpanContext, Format<?, W> format, W carrier)
``` 

What would the user do? They will look at Format class and see a constant: 

```java
Format<.., TextMapWriter> TextMap = ...
``

Now they know what to pass as the `format` parameter, and it explicitly constraints what they should pass as a `carrier`:

```java
tracer.inject(span.getContext(), Format.TextMap, new TextMapWriter(){...})
```

opentracing-api/src/main/java/io/opentracing/Tracer.java
```diff
@@ -15,78 +15,86 @@
...
+       * Add a reference from the Span being built to a distinct (usually parent) Span. May be called multiple times to represent multiple such References.
+       *
+       * @param referenceType the reference type, typically one of the constants defined in References
+       * @param referencedContext the SpanContext being referenced; e.g., for a References.CHILD_OF referenceType, the referencedContext is the parent
        *
-       * If the parent has already been set an IllegalStateException will be thrown.
+       * @see io.opentracing.References
        */
-      SpanBuilder withParent(Span parent);
+      SpanBuilder addReference(Comparable referenceType, SpanContext referencedContext);
```
adriancole: if using Comparable, we should provide rationale for the choice, for example, why ordering matters. It is distracting otherwise.

A guess is that we are expecting an implementation to have multiple parents and need to know the parent vs adopted vs step-parent? I suppose we are expecting the implementation to retain the order?

Another note, Comparable has a type parameter `<T>` which would lead to this method having a compiler warning.

opentracing-api/src/main/java/io/opentracing/propagation/HttpHeaderReader.java
```diff
@@ -0,0 +1,39 @@
...
+ *
+ * @see io.opentracing.Tracer#extract(Object)
+ */
+public interface HttpHeaderReader {
+    /**
+     * Gets HTTP headers from the implementations backing store.
+     *
+     * Note that these headers will often be a superset of whatever was injected via an HttpHeaderWriter in the peer. As such, Tracers should use a unique prefix or substring to identify their header map entries.
+     *
+     * @param key a key suitable for use in an HTTP header (i.e., case-insensitive, no special characters, etc)
```
adriancole: should highlight that keys can repeat (since headers can have many values)

opentracing-api/src/main/java/io/opentracing/propagation/HttpHeaderReader.java
```diff
@@ -0,0 +1,39 @@
...
+     * @param value a value suitable for use in an HTTP header (i.e., URL-escaped)
+     *
+     * @return all entries in the HTTP header map
+     *
+     * @see io.opentracing.Tracer#extract(Object)
+     */
+    Iterable<Map.Entry<String,String>> getEntries();
```
adriancole: there's another choice, which is to return a multimap. Very often the impl will need to copy-out anyway..

Ex from okhttp:

```java
  public Map<String, List<String>> toMultimap() {
    Map<String, List<String>> result = new TreeMap<>(String.CASE_INSENSITIVE_ORDER);
    for (int i = 0, size = size(); i < size; i++) {
      String name = name(i).toLowerCase(Locale.US);
      List<String> values = result.get(name);
      if (values == null) {
        values = new ArrayList<>(2);
        result.put(name, values);
      }
      values.add(value(i));
    }
    return result;
  }
```

opentracing-api/src/main/java/io/opentracing/propagation/HttpHeaderWriter.java
```diff
@@ -0,0 +1,35 @@
...
+     * Puts a key:value pair into the HTTP header map.
+     *
+     * Note that headers added via put() will often share the HTTP header map with other application data. As such, Tracers should use a unique prefix or substring to identify their header map entries.
+     *
+     * @param key a key suitable for use in an HTTP header (i.e., case-insensitive, no special characters, etc)
+     * @param value a value suitable for use in an HTTP header (i.e., URL-escaped)
+     *
+     * @see io.opentracing.Tracer#inject(io.opentracing.SpanContext, Object)
+     */
+    void put(String key, String value);
```
adriancole: since http headers can repeat, should we define what should occur when the key exists?

opentracing-api/src/main/java/io/opentracing/Tracer.java
bensigelman: Yeah, I'll just change to `Object`. I was responding to an earlier comment about the downside of using the (now obliterated) `Reference.Type` enum, as enums by definition are not externally extensible. I don't think the use case for a global ordering nearly compelling enough... we really just want equality comparisons.

opentracing-api/src/main/java/io/opentracing/propagation/HttpHeaderReader.java
bensigelman: I prefer the `Iterable` here since it's more general and allows for fewer copies in some circumstances... good point about the comment above, though, I'll change that.

opentracing-api/src/main/java/io/opentracing/propagation/HttpHeaderReader.java
bensigelman: also, this comment was copy-paste detritus... there should only be a `@return` here

opentracing-api/src/main/java/io/opentracing/propagation/HttpHeaderReader.java
adriancole: unless the user needs to rewind, Iterator is slightly less hassle to
implement.

adriancole: s/rewind/replay

opentracing-api/src/main/java/io/opentracing/Tracer.java
adriancole: :thumbsup:

opentracing-api/src/main/java/io/opentracing/propagation/HttpHeaderWriter.java
bensigelman: this is just me, but I'd rather not. HTTP header support is so inconsistent across a large system, and IMO it's bad practice for a tracing system to rely on repeated headers anyway...

opentracing-api/src/main/java/io/opentracing/propagation/HttpHeaderReader.java
bensigelman: ah, good call -- changed.

opentracing-api/src/main/java/io/opentracing/propagation/HttpHeaderWriter.java
adriancole: I agree I don't think we should support duplicate values, just curious if
we should ask implementors to prevent this somehow. Not documenting is fine
as we could later if it became common enough concern.

bensigelman: 👍 

In practice there will be O(1) headers for things like trace/span ids, and then zero or more headers for baggage. Since the baggage doesn't allow for duplicate keys, this is unlikely to become an issue at this layer (or so I hope).

opentracing-api/src/main/java/io/opentracing/Tracer.java
michaelsembwever: It's a good argument but I think it's simplifying things a bit.

To use the inject method you need to understand what you're doing in relation to the wire and tracing headers. For example that such headers go into the httpHeaders becomes base knowledge.

The use-case of an application having to support multiple formats that use the same carrier i would guess is more rare. As applications become smaller in a microservices world. When each application specifically only registering/configuring the carriers it needs to use simplifies this further.

opentracing-api/src/main/java/io/opentracing/Tracer.java
```diff
@@ -15,78 +15,86 @@
...
+       * Add a reference from the Span being built to a distinct (usually parent) Span. May be called multiple times to represent multiple such References.
+       *
+       * @param referenceType the reference type, typically one of the constants defined in References
+       * @param referencedContext the SpanContext being referenced; e.g., for a References.CHILD_OF referenceType, the referencedContext is the parent
        *
-       * If the parent has already been set an IllegalStateException will be thrown.
+       * @see io.opentracing.References
        */
-      SpanBuilder withParent(Span parent);
+      SpanBuilder addReference(Object referenceType, SpanContext referencedContext);
```
tinkerware: I understand reference types need to be extensible, but is there a need to make them arbitrary types? Why not `String`?

bensigelman: If all we need is `equals()`, I guess I would ask *why* String?

bensigelman:
@adriancole @michaelsembwever @yurishkuro @tinkerware

Since the (controversial) `Format` stuff has been split off from this PR, are there any objections to the remaining content? It seemed like we'd converged on the SpanContext / Reference / extract() aspects presented here.

I'm presently fixing the rest of the (github) repo to work along with these changes... it does require some surgery here and there but most of the diff will be uncontroversial. It's also far less important since OT callers don't depend on any of the impl helpers directly, of course.

Setting expectations: I'd like to merge this PR in 24-36h or so unless substantial new issues are raised.

opentracing-api/src/main/java/io/opentracing/Tracer.java
tinkerware: If we had a `Symbol` type in Java, it would be that. Failing that, isn't it better that we at least pick something that  can `switch`'ed on?

opentracing-api/src/main/java/io/opentracing/Tracer.java
```diff
@@ -15,78 +15,86 @@
...
-   * IllegalStateException.
+   * @param <C> the carrier type: all Tracer.inject() implementations must support io.opentracing.propagation.TextMapWriter, io.opentracing.propagation.HttpHeaderWriter, and java.nio.ByteBuffer
+   * @param spanContext the SpanContext instance to inject into the carrier
+   * @param carrier the carrier for the SpanContext state; when inject() returns, the Tracer implementation will have represented the SpanContext within `carrier`
    *
-   * All implementations support at minimum the required carriers BinaryWriter and TextMapWriter.
+   * @see io.opentracing.propagation
    */
-  <T> void inject(Span span, T carrier);
+  <C> void inject(SpanContext spanContext, C carrier);
```
tinkerware: What's the point to the `C` type parameter when it does not constrain anything? If we are saying that a more elaborate format API is later work, we can just leave the carrier type `Object` for now, though it seems to me that there could at least be a marker interface for carriers.

tinkerware:
This seems okay to me, assuming that carrier/format API  is not in its final form.

opentracing-api/src/main/java/io/opentracing/Tracer.java
```diff
@@ -15,78 +15,86 @@
...
     Tracer tracer = ...
 
-    Span feed = tracer.buildSpan("GetFeed")
-                      .start();
+    Span parentSpan = tracer.buildSpan("DoWork")
+                            .start();
 
     Span http = tracer.buildSpan("HandleHTTPRequest")
-                      .withParent(feed)
+                      .asChildOf(parentSpan.context())
```
yurishkuro: @bensigelman similar to Node/Python, we can overload this to `asChildOf(parentSpan)`

opentracing-api/src/main/java/io/opentracing/Tracer.java
yurishkuro: I agree, this `C` makes it look like there's more to the `carrier` type than just an `Object`. +1 to drop it.

opentracing-api/src/main/java/io/opentracing/References.java
```diff
@@ -0,0 +1,35 @@
...
+ * the License.
+ */
+package io.opentracing;
+
+/**
+ * References is essentially a namespace for the official OpenTracing reference types.
+ *
+ * References are used by Tracer.buildSpan() to describe the relationships between Spans.
+ *
+ * @see io.opentracing.Tracer.SpanBuilder#addReference(Comparable, SpanContext)
```
yurishkuro: rm Comparable (a couple places)

yurishkuro:
lgtm

opentracing-api/src/main/java/io/opentracing/Tracer.java
bensigelman: This was inherited from the code at master...

https://github.com/opentracing/opentracing-java/blob/master/opentracing-api/src/main/java/io/opentracing/Tracer.java#L53

Aside from making the documentation more explicit (`@param <C> ...`), I can't provide any particular justification. I think this is code @michaelsembwever wrote, so I might be missing part of the rationale. (The documentation aspect is not trivial, though...)

michaelsembwever: It helps implementations. See AbstractTracer for example.
It also creates a stub to explain Carrier types.

No biggie either way for me.

michaelsembwever:
:+1:

opentracing-api/src/main/java/io/opentracing/SpanContext.java
```diff
@@ -0,0 +1,39 @@
...
+     *
+     * @return this SpanContext instance, for chaining
+     */
+    SpanContext setBaggageItem(String key, String value);
+
+    /**
+     * @return the value of the baggage item identified by the given key, or null if no such item could be found
+     */
+    String getBaggageItem(String key);
+}
```
dawallin: In go a method to iterate through all baggage items were introduced. Should you have it here as well ?

opentracing-api/src/main/java/io/opentracing/SpanContext.java
```diff
@@ -0,0 +1,39 @@
...
+     *
+     * Baggage enables powerful distributed context propagation functionality where arbitrary application data can be carried along the full path of request execution throughout the system.
+     *
+     * Note 1: Baggage is only propagated to the future (recursive) children of this SpanContext.
+     *
+     * Note 2: Baggage is sent in-band with every subsequent local and remote calls, so this feature must be used with care.
+     *
+     * @return this SpanContext instance, for chaining
+     */
+    SpanContext setBaggageItem(String key, String value);
```
dawallin: I don't like the setBaggageItem method on the span context. It forces span context to be mutable and opened for side effects. I rather like to leave it on the aggregate root (the Span) which can forward the baggage item to the context internally. That would make it possible to have the spanContext immutable.

yurishkuro: @dawallin if the SpanContext does not contain the baggage, then `tracer.Extract` cannot return just the SpanContext, and we're back to the problem with `Join` which was both extracting the context _and_ starting the span

dawallin: The SpanContext should definitely contain the baggage, it is only the set method that I don't like since it makes the SpanContext mutable. The Extract method creates the SpanContext and can pass the baggage as a list in the spanContext constructor, it does not need the set method.

opentracing-api/src/main/java/io/opentracing/Tracer.java
adriancole: good point. I think that parameter doesn't make sense here since it doesn't
bind any other parameters or the return type.

opentracing-api/src/main/java/io/opentracing/Tracer.java
adriancole: good point on switching. enum couldn't be used bc enums aren't extensible.
for switch, we're left with ints or strings, strings require java 7+ (which
I think this library does anyway)

opentracing-api/src/main/java/io/opentracing/SpanContext.java
bensigelman: I would like this to happen in a followup... there are some subtleties that @tinkerware brought up re immutability that I think are relevant here (namely, if we know that SpanContexts are immutable, the interface can be zero-copy *and* return an iterator *and* be threadsafe... otherwise I'd rather do something like what we did in Go and have the SpanContext impl invoke a callback repeatedly).

opentracing-api/src/main/java/io/opentracing/SpanContext.java
bensigelman: @dawallin I see your point that the baggage could be set when the SpanContext impl is created. My best argument for keeping setBaggageItem on SpanContext is just symmetry: having the setter and getter on different objects will confuse people. There's also the question of whether baggage should be settable after Span `finish` time... if not, that would be yet another argument in favor of what you're proposing.

All of this said, I would like to keep the setter and getter where they are for this PR. Your arguments apply across languages and we should be discussing them in the right place, not here on ot-java. I've created https://github.com/opentracing/opentracing.github.io/issues/106 .

bensigelman:
Okay -- I changed a bunch of stuff and fixed the build and tests in the rest of this repo (including NoopTracer, which is probably the only part anyone actually uses at this point).

opentracing-api/src/main/java/io/opentracing/Span.java
```diff
@@ -19,60 +19,53 @@
  * <p>Spans are created by the {@link Tracer#buildSpan} interface.
  */
 public interface Span extends AutoCloseable {
+    /**
+     * Retrieve the associated SpanContext.
+     *
+     * @return the SpanContext that encapsulates Span state that should propagate across process boundaries.
+     */
+    SpanContext getContext();
```
bensigelman: I changed this to `getContext()` since this is Java :)

(I'd internalized Go's no-get getter convention, I guess)

opentracing-api/src/main/java/io/opentracing/Tracer.java
bensigelman: this is a String now.

bensigelman:
(and if there are no further concerns, I will merge this sometime tomorrow)

opentracing-api/src/main/java/io/opentracing/Span.java
adriancole: going without a getX is not abnormal when there is no setter. A lot of
libraries follow this, including zipkin. I prefer is as context()

adriancole: ps the history of getX setX is JavaBeans. this isn't a java bean, so
doesn't need to behold to that. A lot of the api has more to do with
builders than beans
http://www.informit.com/articles/article.aspx?p=1216151&seqNum=2

bensigelman:
@adriancole I dropped the `getContext()` commit. Thanks for the history lesson, too. :)

bensigelman:
Okay, as promised, I'm merging this now. Thanks for all of the feedback.

https://github.com/opentracing/opentracing-java/issues/31 and https://github.com/opentracing/opentracing.github.io/issues/106 are still open and could impact this API going forward.

adriancole:
Thanks @bensigelman!

PR #33, opened 2016-07-24
introduce a Format type in Java

bensigelman:
TL;DR, this is option 3 described at https://github.com/opentracing/opentracing-java/issues/31#issuecomment-234567476

I will make a few comments inline to focus attention on the important bits.


opentracing-api/src/main/java/io/opentracing/Tracer.java
```diff
@@ -45,36 +46,42 @@
...
-   * @param carrier the carrier for the SpanContext state; when inject() returns, the Tracer implementation will have represented the SpanContext within `carrier`. All Tracer.inject() implementations must support io.opentracing.propagation.TextMapWriter, io.opentracing.propagation.HttpHeaderWriter, and java.nio.ByteBuffer.
+   * @param format the Format of the carrier
+   * @param carrier the carrier for the SpanContext state. All Tracer.inject() implementations must support io.opentracing.propagation.TextMap and java.nio.ByteBuffer.
    *
-   * @see io.opentracing.propagation
+   * @see io.opentracing.propagation.Format
+   * @see io.opentracing.propagation.Format.Builtin
    */
-  void inject(SpanContext spanContext, Object carrier);
+  <C> void inject(SpanContext spanContext, Format<C> format, C carrier);
```
bensigelman: obviously the changes to `inject` and `extract` are the main events here.

opentracing-api/src/main/java/io/opentracing/propagation/HttpHeaderReader.java
```diff
@@ -1,37 +0,0 @@
...
-import java.util.Map;
-
-/**
- * HttpHeaderReader is a built-in carrier for Tracer.inject().
- *
- * HttpHeaderReader implementations allows Tracers to write key:value String pairs into arbitrary HTTP header representations. In this way, HttpHeaderReader prevents a tight coupling between OpenTracing and any particular Java HTTP Header representation.
- *
- * @see io.opentracing.Tracer#extract(Object)
- */
-public interface HttpHeaderReader {
```
bensigelman: as @yurishkuro has aptly pointed out, introducing a Format type allows us to delete HttpHeaderReader/Writer entirely since the TextMap carrier is a perfectly capable interface... what we are really trying to communicate to the `Tracer` implementation is the *allowable keys and values*, and that is more explicitly (and readably) accomplished via a format parameter than via `instanceof`.

opentracing-api/src/main/java/io/opentracing/propagation/TextMap.java
```diff
@@ -0,0 +1,50 @@
...
+import java.util.Iterator;
+import java.util.Map;
+
+/**
+ * TextMap is a built-in carrier for Tracer.inject() and Tracer.extract(). TextMap implementations allows Tracers to read and write key:value String pairs from arbitrary underlying sources of data.
+ *
+ * @see io.opentracing.Tracer#inject(SpanContext, Format, Object)
+ * @see io.opentracing.Tracer#extract(Format, Object)
+ */
+public interface TextMap {
```
bensigelman: If we went with `Format<R, W>` as @yurishkuro has suggested, this would be split into separate interfaces for TextMap reading and writing.

opentracing-impl-java8/src/main/java/io/opentracing/AbstractSpanBuilder.java
```diff
@@ -11,7 +11,7 @@
  * or implied. See the License for the specific language governing permissions and limitations under
  * the License.
  */
-package io.opentracing.propagation;
+package io.opentracing;
```
bensigelman: (somehow UnIntelliJ moved a bunch of stuff in my previous PR; these moves back are not related to the PR but should happen anyway.)

opentracing-impl-java8/src/main/java/io/opentracing/AbstractTracer.java
```diff
@@ -36,20 +34,20 @@ public SpanBuilder buildSpan(String operationName){
     }
 
     @Override
-    public void inject(SpanContext spanContext, Object carrier) {
-        registry.getInjector(carrier.getClass()).inject(spanContext, carrier);
+    public <C> void inject(SpanContext spanContext, Format<C> format, C carrier) {
```
bensigelman: all of this generics stuff can come back now... it was lost in #32. It *does* serve a useful purpose; the compiler/IDE is able to enforce constraints about particular carrier formats.

opentracing-impl-java8/src/test/java/io/opentracing/propagation/TestTextMapExtractorImpl.java
```diff
@@ -13,17 +13,16 @@
...
+
 import java.util.Iterator;
 import java.util.Map;
 
-import io.opentracing.propagation.TextMapReader;
-import io.opentracing.SpanContext;
-
-public class TestTextMapExtractorImpl implements Extractor {
-    public SpanContext extract(Object carrier) {
-        TextMapReader textMapCarrier = (TextMapReader)carrier;
```
bensigelman: per https://github.com/opentracing/opentracing-java/pull/33/files#r71980119 , note how the right side of the diff can just declare the `extract(TextMap carrier)` directly, eliminating the need for this cast.

bensigelman: (admittedly this is just on the Tracer impl side which is far less important than the user code)

opentracing-impl-java8/src/test/java/io/opentracing/AbstractTracerTest.java
```diff
@@ -64,8 +58,8 @@ public SpanContext context() {
             }
         };
         Map<String,String> map = new HashMap<>();
-        TextMapWriter carrier = new TextMapImpl(map);
-        instance.inject(span.context(), carrier);
+        TextMap carrier = new TextMapImpl(map);
+        instance.inject(span.context(), Format.Builtin.TEXT_MAP, carrier);
```
bensigelman: This is approximately how vanilla OpenTracing user code would look after this change. I personally prefer how clear/explicit this is, as inject/extract can be confusing and the in-place documentation of TEXT_MAP is useful. It also allows the compiler/IDE to force the format and carrier to match from a type-safety perspective (unlike `master` with its Object params).

opentracing-impl/src/main/java/io/opentracing/propagation/TextMapImpl.java
```diff
@@ -16,7 +16,7 @@
 import java.util.Iterator;
 import java.util.Map;
 
-public final class TextMapImpl implements TextMapWriter, TextMapReader {
+public final class TextMapImpl implements TextMap {
```
yurishkuro: I would rather keep TextMapWriter and TextMapReader (instead of Impl) as concrete implementations, and let them throw NotImplementedException from one of the methods.

It's also problematic that this class is defined in the `-impl` jar, because it is intended for the end user of the API, and the end users have no need to depend on the `-impl` module.

yurishkuro:
I personally prefer `Format<R, W>` as a more strongly typed approach, but introducing a format in the first place is more important, so +1

opentracing-impl/src/main/java/io/opentracing/propagation/TextMapImpl.java
bensigelman: In other languages we haven't provided text map implementations as part of the core API. Not that it's needed by `-impl`, either: the only caller of TextMapImpl is test code at the moment.

As for splitting it into reader/writer pieces: I can see it either way. If we're going to provide helpers for the `Map<String, String>` case, though, I think we can make the intention clearer through naming. I'll give it a shot.

bensigelman: Okay, done. I went with `TextMapInjectAdapter` and `TextMapExtractAdapter` which is clearer and more self-documenting (at least in my brain).

yurishkuro: > In other languages we haven't provided text map implementations as part of the core API. 

We do have `TextMapCarrier` in [Go](https://github.com/opentracing/opentracing-go/blob/master/propagation.go#L110). That one is actually R/W, but it's ok because in Go the accepted carrier interfaces are separated to R and W.

opentracing-api/src/main/java/io/opentracing/propagation/Format.java
```diff
@@ -0,0 +1,70 @@
...
+ */
+package io.opentracing.propagation;
+
+import io.opentracing.SpanContext;
+import io.opentracing.Tracer;
+
+import java.nio.ByteBuffer;
+
+/**
+ * Format instances control the behavier of Tracer.inject and Tracer.extract (and also constrain the type of the carrier parameter to same).
```
tinkerware: `behavier` → `behavior`

bensigelman: (thanks, fixed)

bensigelman:
@adriancole @michaelsembwever curious if either of you has an opinion about this. IMO the self-documentation benefits make this worthwhile independent of anything else, though there are also some wins for implementations from a compile-time type-checking standpoint.

I realize it's been a contentious (and long-standing) issue, so at this point I'm looking less for fanatical consensus and more for "this seems reasonable". Of course fanatical consensus is welcome, too. :)

michaelsembwever:
> the self-documentation benefits make this worthwhile independent of anything else, though there are also some wins for implementations from a compile-time type-checking standpoint.

This sums it up well. The second point I feel is moot. The first not so simple, because adding constructs can also be seen as a burden for a concept that the user has to understand anyway. Try and make an API extra simple but adding more and more types and by underestimating rather basic intelligence of the user you end up achieving the opposite. Whether that's true in this situation I really need to read more examples to make a decision.

For me, I was never against the introduction of the Format parameter, I just saw it as potentially superfluous and wanted to see time prove that true or not. For that it is (3) in Yuri's comment https://github.com/opentracing/opentracing-java/issues/31#issuecomment-232505628 that provides enough rationale to go ahead with this change.

I appreciate the patience and explanations that have gone into this.

bensigelman:
@michaelsembwever thank you kindly for giving this a look. You wrote:

> The first not so simple, because adding constructs can also be seen as a burden for a concept that the user has to understand anyway.

Absolutely agree. Without an explicit `Format` parameter, the user is quite literally controlling the *semantic* behavior of the underlying `Tracer` impl via `instanceof`... I think it's an undue burden on the *reader* of inject/extract code to not just know the type but the *supertype(s)* of actual parameters being passed around a codebase. The `format` param is a little blunt, but it clarifies intentions.

Re patience/explanations: same to you.

michaelsembwever:
> I think it's an undue burden on the reader of inject/extract code to not just know the type but the supertype(s) of actual parameters being passed around a codebase.

Yes this is true.

bensigelman:
Well... I am going to merge this in ~24h or so if there are no further objections.

PR #39, opened 2016-08-12
Expose public singleton instance of NoopTracer

yurishkuro:
All other classes in this package expose singletons, any reason not to do the same for NoopTracer?


opentracing-impl/src/main/java/io/opentracing/NoopTracer.java
```diff
@@ -17,6 +17,8 @@
 
 public final class NoopTracer implements Tracer {
 
+    public final static NoopTracer INSTANCE = new NoopTracer();
```
bensigelman: LGTM!

michaelsembwever:
@yurishkuro 
nit: static goes before final.

michaelsembwever:
Also where was the 24hr turn around on this?

PR #43, opened 2016-08-16
Add UnsupportedFormatException and clarify expectations for inject/extract methods

### LINKED ISSUE #42: Clarify inject/extract exceptions
The most common outcome of inject/extract methods that instrumentation needs to deal with are
1. success
2. unsupported format
3. empty carrier (extract only)
4. other exceptions (including invalid carrier type, corrupted carrier, etc.)

In Python API we specified exactly how tracer must react to these, specifically by throwing known exception for (2) and returning None for (3). The case (4) is usually a result of misconfiguration, so the exact exception is not that important.

I would like to have a well-defined (unchecked) exception type for (2), and specify that `null` must be returned for (3).

Side note: inject() documentation refers to carrier types instead of formats: `@param carrier the carrier for the SpanContext state. All Tracer.inject() implementations must support io.opentracing.propagation.TextMap and java.nio.ByteBuffer.`

@bensigelman @michaelsembwever

yurishkuro:
Fixes #42

@bensigelman @michaelsembwever

### DIFF
--- a/opentracing-api/src/main/java/io/opentracing/Tracer.java
+++ b/opentracing-api/src/main/java/io/opentracing/Tracer.java
@@ -93,9 +93,9 @@ public interface Tracer {
    * IllegalArgumentException.
    *
    * @param <C> the carrier type, which also parametrizes the Format.
-   * @param format the Format of the carrier
-   * @param carrier the carrier for the SpanContext state. All Tracer.extract() implementations must support
-   *                io.opentracing.propagation.TextMap and java.nio.ByteBuffer.
+   * @param format the Format of the carrier. All Tracer.inject() implementations
+   *               must support formats defined in io.opentracing.propagation.Format.Builtin
+   * @param carrier the carrier for the SpanContext state.
    *
    * @return the SpanContext instance extracted from the carrier holding context to create a Span or null if it cannot be extracted.
    *
--- a/opentracing-api/src/main/java/io/opentracing/Tracer.java
+++ b/opentracing-api/src/main/java/io/opentracing/Tracer.java
@@ -1,4 +1,4 @@
-/**
+/*
  * Copyright 2016 The OpenTracing Authors
  *
  * Licensed under the Apache License, Version 2.0 (the "License"); you may not use this file except
@@ -42,6 +42,10 @@ public interface Tracer {
   /**
    * Inject a SpanContext into a `carrier` of a given type, presumably for propagation across process boundaries.
    *
+   * If tracer does not support given format, it must throw
+   * io.opentracing.propagation.UnsupportedFormatException so that instrumentation
+   * can try another format.
+   *
    * <p>Example:
    * <pre>{@code
    * Tracer tracer = ...
@@ -52,23 +56,37 @@ public interface Tracer {
    *
    * @param <C> the carrier type, which also parametrizes the Format.
    * @param spanContext the SpanContext instance to inject into the carrier
-   * @param format the Format of the carrier
-   * @param carrier the carrier for the SpanContext state. All Tracer.inject() implementations must support io.opentracing.propagation.TextMap and java.nio.ByteBuffer.
+   * @param format the Format of the carrier. All Tracer.inject() implementations
+   *               must support formats defined in io.opentracing.propagation.Format.Builtin
+   * @param carrier the carrier for the SpanContext state.
+   *
+   * @throws io.opentracing.propagation.UnsupportedFormatException when tracer does not support given format
    *
    * @see io.opentracing.propagation.Format
    * @see io.opentracing.propagation.Format.Builtin
+   * @see io.opentracing.propagation.UnsupportedFormatException
    */
   <C> void inject(SpanContext spanContext, Format<C> format, C carrier);
 
   /**
    * Extract a SpanContext from a `carrier` of a given type, presumably after propagation across a process boundary.
    *
+   * If tracer does not support given format, it must throw
+   * io.opentracing.propagation.UnsupportedFormatException so that instrumentation
+   * can try another format.
+   *
+   * If given carrier does not contain information to construct SpanContext,
+   * the tracer must return `null`. If SpanContext cannot be extracted because
+   * the data in the carrier is malformed, the tracer should return null, rather
+   * than throw an exception. The tracer may handle the exceptions internally
+   * by logging them or incrementing an error counter.
+   *
    * <p>Example:
    * <pre>{@code
    * Tracer tracer = ...
    * TextMap httpHeadersCarrier = new AnHttpHeaderCarrier(httpRequest);
    * SpanContext spanCtx = tracer.extract(Format.Builtin.HTTP_HEADERS, httpHeadersCarrier);
-   * tracer.buildSpan('...').withChildOf(spanCtx).start();
+   * tracer.buildSpan('...').asChildOf(spanCtx).start();
    * }</pre>
    *
    * If the span serialized state is invalid (corrupt, wrong version, etc) inside the carrier this will result in an
@@ -79,10 +97,13 @@ public interface Tracer {
    * @param carrier the carrier for the SpanContext state. All Tracer.extract() implementations must support
    *                io.opentracing.propagation.TextMap and java.nio.ByteBuffer.
    *
-   * @return the SpanContext instance holding context to create a Span.
+   * @return the SpanContext instance extracted from the carrier holding context to create a Span or null if it cannot be extracted.
    *
+   * @throws io.opentracing.propagation.UnsupportedFormatException when tracer does not support given format
+
    * @see io.opentracing.propagation.Format
    * @see io.opentracing.propagation.Format.Builtin
+   * @see io.opentracing.propagation.UnsupportedFormatException
    */
   <C> SpanContext extract(Format<C> format, C carrier);
 
--- a/opentracing-api/src/main/java/io/opentracing/Tracer.java
+++ b/opentracing-api/src/main/java/io/opentracing/Tracer.java
@@ -1,4 +1,4 @@
-/*
+/**
  * Copyright 2016 The OpenTracing Authors
  *
  * Licensed under the Apache License, Version 2.0 (the "License"); you may not use this file except
--- a/opentracing-api/src/main/java/io/opentracing/Tracer.java
+++ b/opentracing-api/src/main/java/io/opentracing/Tracer.java
@@ -93,7 +93,7 @@ public interface Tracer {
    * IllegalArgumentException.
    *
    * @param <C> the carrier type, which also parametrizes the Format.
-   * @param format the Format of the carrier. All Tracer.inject() implementations
+   * @param format the Format of the carrier. All Tracer.extract() implementations
    *               must support formats defined in io.opentracing.propagation.Format.Builtin
    * @param carrier the carrier for the SpanContext state.
    *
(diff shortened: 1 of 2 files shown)


bensigelman:
LGTM

opentracing-api/src/main/java/io/opentracing/Tracer.java
```diff
@@ -52,37 +56,54 @@
...
    * @param <C> the carrier type, which also parametrizes the Format.
-   * @param format the Format of the carrier
-   * @param carrier the carrier for the SpanContext state. All Tracer.extract() implementations must support
-   *                io.opentracing.propagation.TextMap and java.nio.ByteBuffer.
+   * @param format the Format of the carrier. All Tracer.extract() implementations
+   *               must support formats defined in io.opentracing.propagation.Format.Builtin
+   * @param carrier the carrier for the SpanContext state.
    *
-   * @return the SpanContext instance holding context to create a Span.
+   * @return the SpanContext instance extracted from the carrier holding context to create a Span or null if it cannot be extracted.
```
michaelsembwever: please don't return `null`

opentracing-api/src/main/java/io/opentracing/propagation/UnsupportedFormatException.java
```diff
@@ -0,0 +1,24 @@
...
+ * or implied. See the License for the specific language governing permissions and limitations under
+ * the License.
+ */
+package io.opentracing.propagation;
+
+/**
+ * This exception must be thrown by Tracer's inject/extract methods
+ * when the tracer does not support provided format.
+ */
+public class UnsupportedFormatException extends RuntimeException {
```
michaelsembwever: This should be either a checked exception, or we should be using `IllegalArgumentException`.

If it's unchecked, then it's a programmatic fault, the programmer should have implicitly or explicitly known that this format was not supported, and in that case the proper result is an `IllegalArgumentException`.

Note there is no way of explicitly checking whether the format is supported. Implicitly one knows through the configuration (registration) of the injectors and extractors. Note that calls to `extract(..)` written in libraries or frameworks are unaware of the contextual configuration to the application.

Without adding a method that can check whether a format is supported, and without making a statement that `extract(..)` called from within libraries or frameworks is not supported, I think the case is clear that this needs to be a checked exception.

michaelsembwever:
It's worth noting that the code already returns an unchecked exception.

    throw new AssertionError("no registered extractor for " + format);

But the issue and the PR seems to start off on the wrong foot.

yurishkuro:
> It's worth noting that the code already returns an unchecked exception.

You're probably referring to the -impl code, which I am not interested in. The goal was to set expectations on the API. Also, generic exceptions like IllegalArgument or Assertion are not useful for the explicit purpose of identifying the condition as unsupported format.

So two open questions:
  1. Unchecked or checked exception or introspection methods. I don't have a strong opinion on either of the three, slight preference for introspection methods, which are better than catching exceptions.
  1. Returning null or Option or noop span context. I am against the latter, I think it will create more problems than it solves. We can add our own Option class, I don't think it necessarily increases the API surface because it's a utility with a well known meaning & usage patterns. But in terms of fluid syntax, returning an Option actually makes for more awkward code, because the common usage pattern for extracted span context is to pass it as Reference, and tracers can easily deal with References where referenced context is null. I agree that returning null is less pretty.

michaelsembwever:
> You're probably referring to the -impl code, which I am not interested in. 

I don't think it's appropriate to raise PRs that ignore the -impl code in the same repository.

> Also, generic exceptions like IllegalArgument or Assertion are not useful for the explicit purpose of identifying the condition as unsupported format.

That is an incorrect statement. For unchecked exceptions re-using IllegalArgumentException would be the normal Java precedent. There is rarely a need to create a new unchecked exception type in any java codebase.

michaelsembwever:
> Unchecked or checked exception or introspection methods. I don't have a strong opinion on either of the three, slight preference for introspection methods, which are better than catching exceptions.

I do have a strong opinion, based on my analysis in https://github.com/opentracing/opentracing-java/pull/43#discussion_r75054612

Whether you want to catch exceptions or not has *nothing* to do with it.

> Returning null or Option or noop span context.

I'm against null and against a custom additional Option type..

> But in terms of fluid syntax, returning an Option actually makes for more awkward code, because the common usage pattern for extracted span context is to pass it as Reference, and tracers can easily deal with References where referenced context is null.

A reference can just as easily handle an empty SpanContext as it can a null.
Given that Java precedence is to avoid `null` it makes more sense that a reference of an empty spanContext is preferred to than a reference of a null.

yurishkuro:
> I don't think it's appropriate to raise PRs that ignore the -impl code in the same repository.

-impl module is just it, _an_ implementation. It's not a spec, not a reference implementation, it is not even used as a base by the two actual production implementations of the API that I know of. If we want to standardize on a certain tracer behavior, it must be documented in the API.

tinkerware:
>  It's not a spec, not a reference implementation, it is not even used as a base by the two actual production implementations of the API that I know of. If we want to standardize on a certain tracer behavior, it must be documented in the API.

Then why is it in the repo? If the intention is to provide guidance to implementors, ignoring its existence in our own PR's is counterproductive.

opentracing-api/src/main/java/io/opentracing/Tracer.java
tinkerware: I have to second this. There's no good reason to introduce `null` return values to the API when we can easily return an empty context.

yurishkuro:
> Then why is it in the repo? If the intention is to provide guidance to implementors, ignoring its existence in our own PR's is counterproductive.

Once again, just because there is code in the -impl-java8 that throws AssertionException (which in no way solves the issue at hand, btw), it does not mean that it is a requirement of the API that all implementations do so. And the question of why the impl code is in this repo in the first place has been repeatedly raised in other tickets. In other OT languages the reference implementations were always done in separate repos called `basictracer-{lang}`. `-impl-java8` is not a basic tracer because it's not a full implementation --> #44.

opentracing-api/src/main/java/io/opentracing/Tracer.java
yurishkuro: I am not in favor of making an API that lies to the users. Knowing if there was or wasn't a tracing context in the incoming request may be important to some consumers. By returning an empty context we're loosing that bit of information, because the context interface has no properties to identify itself as "empty".

tinkerware: I'd be in favor of adding something like this:

```java
public interface SpanContext {
  SpanContext NONE = NullSpanContext.INSTANCE;
  
   Iterable<Map.Entry<String, String>> baggageItems();
}

/*package-private*/ enum NullSpanContext implements SpanContext {
  INSTANCE;

  @Override public Iterable<Map.Entry<String, String>> baggageItems() {
    return Collections.emptySet();
  }
}
```

With this API, you can specify that implementations disregard any attempt to add a reference that uses `SpanContext.NONE` is no-op; the following code works without change whether a context was present or not:

```java
SpanContext spanCtx = tracer.extract(Format.Builtin.HTTP_HEADERS, httpHeadersCarrier);
tracer.buildSpan('...').asChildOf(spanCtx).start();
```

API users that do want to know if the incoming context actually had a tracing context or not can still check for it by reference equality:

```java
SpanContext spanCtx = tracer.extract(Format.Builtin.HTTP_HEADERS, httpHeadersCarrier);
if (spanCtx == SpanContext.NONE) {
 // handle missing context
}
```

The `enum` construct is there to prevent multiple instances of a null context from being created in some edge cases.

yurishkuro: That's what I meant by introducing a new concept. This NOOP constant now becomes defined at the API level and affects everyone. Tracer implementations need to check for it every time they are passed a span context. Users may also be affected. For example, I have some users who insist on knowing the trace ID, for logging and other purposes, and they are ok with direct dependency on our `com.uber.jaerger.SpanContext` implementation class to get access to that trace ID by down-casting.

So just to preserve the yes/no bit about the trace context in the request we'd be introducing this new concept with much greater impact.

If Java supported multiple return values as Go does, this would've been simple:
```go
if spanContext, ok := tracer.extract(...); !ok {
   // act when not found
}
```
In Scala, this would've been `Option[SpanContext]` with no arguments whatsoever. Both of these are examples of good APIs where the impact of returning yes/no bit is localized to the method in question.

tinkerware: > If Java supported multiple return values as Go does, this would've been simple 

The idiomatic way this happens in Java is either through a checked exception to indicate a recoverable error, a null object like the one I suggested, or a separate result object such as `ExtractResult` (assuming we are not targeting Java 7 and above; using `Optional` is a choice in that case). I'm leaning towards the null object pattern because it is cleanest in terms up API usage. If consensus is such that existing apps that downcast to implementations of a pre-1.0 spec can not be asked to change, then you can introduce `ExtractResult` or throw a checked exception. _Either_ of those options would be better than returning `null`.

> Tracer implementations need to check for it every time they are passed a span context. Users may also be affected.

I think requiring no-op behavior for null contexts from implementations is much preferable if it means orders of magnitude more code written by clients of the OT API have a simpler means of dealing with missing tracing contexts.

> So just to preserve the yes/no bit about the trace context in the request we'd be introducing this new concept with much greater impact.

You may say this is just semantics, but I think the concept is not new. It was always there but the API did not recognize it: _tracing is optional_; lack of a distributed tracing context should not cause failure in an app. I think the question is how we deal with it. 

As for apps that have a direct dependency on a particular tracing implementation; why do they care about the OT API in this case? Can you not keep using Jaeger API to extract tracing contexts? Alternatively, Jaeger is always free to return its own null context implementation from `extract` and check for it in addition to the one in the OT API. This limits the cost of having to preserve `SpanContext` downcasts to just Jaeger users.

michaelsembwever:
@yurishkuro would you be interested in closing this PR and opening two separate atomic PRs. One to introduce the IllegalArgumentException along with the `has...(..)` methods as described in https://github.com/opentracing/opentracing-java/pull/43#discussion_r75054612 . And another to discuss empty SpanContexts and how they can be implemented without changing the API.

yurishkuro:
@michaelsembwever yes, I was thinking the same, but before spending time on that I'd like to see if @bensigelman has objections to the `has...(..)` methods:
```
boolean hasInjectorFor(Format<C> format);
boolean hasExtractorFor(Format<C> format);
```

bensigelman:
@michaelsembwever @yurishkuro @tinkerware 

I don't like using exceptions for ordinary control flow, so the `has*For` methods seem sensible, sure. If it were me, I would force Tracers to be symmetric about their support for various formats and call it `supportsFormat(Format<C> format)` (just in the way that we ended up with `Format<C>` rather than `Format<R, W>`).

bensigelman:
Where are we with this PR? I like `supportsFormat<C>` and could live with the injector- and extractor-specific variants...

yurishkuro:
This PR is dead as neither of the two changes are going through. I found that I don't need the "supports format" for the use case I was working on, so de-prioritized this. FWIW, I also prefer a single `supportsFormat()` method.

Also FWIW, even though my comment addition about `null` did not go through, the current wording does not say that null cannot be returned, which means it can, which is what our tracer implementation is doing when span context is not found in the carrier. I'd be curious to know what other real implementations are doing, and if they also return null, I think we should document it as the current behavior, until someone comes up with a sensible alternative (which I haven't seen so far).

PR #47, opened 2016-09-02
Key-value logging

bensigelman:
Per https://github.com/opentracing/opentracing.github.io/issues/96

This is only a sketch... once we decide on a specific API I will adjust tests and dependent artifacts.


opentracing-api/src/main/java/io/opentracing/Span.java
```diff
@@ -65,25 +65,18 @@
...
-     * required to retain all payload arguments (or even all parts of all payload arguments).
-     *
-     * The timestamp of this log event is the current time.
-     **/
-    Span log(String eventName, /* @Nullable */ Object payload);
-
+     * @param keyValues Alternating key Strings and value Objects. Values may be numeric types, bools, Strings, or
+     *                  arbitrary objects, though the treatment of arbitrary Objects varies across Tracer/Span
+     *                  implementations.
+     * @return the Span, for chaining
```
yurishkuro: usage example in the docs would be useful

bensigelman:
@yurishkuro I added a trivial usage example for clarity.

Curious to know what others thing (including C# people who have some analogous constraints)... @adriancole @michaelsembwever @dawallin @cwe1ss

Just so we're all on the same page: the idea here is to choose a narrow base logging API rather than trying to please all parties who want something more elaborate. Those sorts of elaborate interfaces can be projected onto this underlying Span API (and, in many cases, I would hope that adapters could be written that translate slf4j-style logging into Span key:value logging (though we need a standard way to access the active span... something I'd like to take on later this year).

cwe1ss:
Is this now just an object array? This would allow non-string keys and an uneven number of parameters. That seems too generic IMO. I'd prefer a map/dictionary or a list of tuple/custom type to at least limit the key-type to string.

yurishkuro:
@cwe1ss those can always be added, but they are significantly less efficient

cwe1ss:
@yurishkuro, I did a quick test in C# (I guess Java will be similar) and compared object array vs. custom type array vs KeyValuePair list vs. Dictionary:

```
Iterations: 10,000,000
Log(object[]): 633ms
Log(LogValue[]): 839ms
Log(KVP List): 1456ms
Log(Dictionary): 3956ms
```

As expected, Dictionary ("Map" in Java) is the slowest because it ensures unique keys. However, a custom KeyValue-type is just marginally slower than the object array but it at least ensures string keys and a correct number of arguments. It's a little bit more verbose though.

Using the object array has a very short syntax, but at least in my experience, using this for a pair of values is very uncommon in C#. I think I've never seen this before. Is this common in Java? I'm afraid this would be very confusing for users of this library?!

That's why I'd still prefer the custom type over the object-array.

```csharp
// That's the code from the test
public static void Main(string[] args)
{
    int iterations = 10000000;

    Console.WriteLine("Hello World! Iterations: " + iterations);

    // object array
    var watch = Stopwatch.StartNew();
    for (int i = 0; i < iterations; i++)
    {
        LogObjectArray("key1", "value1", "key2", 10, "key3", 4.3, "key4", true);
    }
    Console.WriteLine("Log(object[]): " + watch.ElapsedMilliseconds + "ms");

    // custom type array
    watch.Restart();
    for (int i = 0; i < iterations; i++)
    {
        LogCustomTypeArray(
            new LogValue("key1", "value1"),
            new LogValue("key2", 10),
            new LogValue("key3", 4.3),
            new LogValue("key4", true)
        );
    }
    Console.WriteLine("Log(LogValue[]): " + watch.ElapsedMilliseconds + "ms");

    // KeyValuePair List
    watch.Restart();
    for (int i = 0; i < iterations; i++)
    {
        LogList(new List<KeyValuePair<string, object>>
        {
            new KeyValuePair<string, object>("key1", "value1"),
            new KeyValuePair<string, object>("key2", 10),
            new KeyValuePair<string, object>("key3", 4.3),
            new KeyValuePair<string, object>("key4", true)
        });
    }
    Console.WriteLine("Log(KVP List): " + watch.ElapsedMilliseconds + "ms");

    // dictionary
    watch.Restart();
    for (int i = 0; i < iterations; i++)
    {
        LogDictionary(new Dictionary<string, object>
        {
            { "key1", "value1" },
            { "key2", 10 },
            { "key3", 4.3 },
            { "key4", true }
        });
    }
    Console.WriteLine("Log(Dictionary): " + watch.ElapsedMilliseconds + "ms");
}

public static void LogObjectArray(params object[] args) { }

public static void LogCustomTypeArray(params LogValue[] args) { }

public static void LogList(IList<KeyValuePair<string, object>> args) { }

public static void LogDictionary(IDictionary<string, object> args) { }

public struct LogValue
{
    public string Key { get; }
    public object Value { get; }

    public LogValue(string key, object value)
    {
        Key = key;
        Value = value;
    }
}
```

cwe1ss:
Having a custom type instead of a simple array would also allow us to add helper methods for known keys later. Example:

```csharp
// sorry for the continuous use of C# in this java repository :)
public struct LogValue
{
    public string Key { get; }
    public object Value { get; }

    public LogValue(string key, object value)
    {
        Key = key;
        Value = value;
    }

    // could be directly on the type or on a separate class
    public static LogValue Level(string level) => new LogValue("level", level);
    public static LogValue EventName(string eventName) => new LogValue("event", eventName);
}

// usage
span.Log(LogValue.Level("info"), new LogValue("key", "value"));
```

Inheritance (e.g. `new LogLevelValue("info")`) would also be possible, but this seems to have a higher impact on perf.

If I understand the go implementation correctly, it also does all kinds of strong typing and forces key/value usage. So I wonder why we wouldn't want to have this in Java/C# if there's only a rather insignificant perf hit.

yurishkuro:
One thing your benchmark does not take into account is the increased GC pressure from allocating all host LogValue objects. In Go those objects are passed by value and have no heap allocations. So in Go you need to use more verbose syntax in return for higher efficiency, while in Java a more verbose syntax leads to lower efficiency.

But I am not opposed to having both.

cwe1ss:
in C#, a `struct` indicates a value type which is also passed by value and usually doesn't go on the heap. You can't use inheritance on these types though.

Does Java have something similar?

cwe1ss:
I ran my test using the unoptimized DEBUG mode. In RELEASE mode, using the custom type is even faster than using the object array. :smile: (I don't know why - maybe because there's less boxing involved?!) I also ran a memory profiler (JetBrains dotMemory) and couldn't see any noticeable difference.

So, unless you want otherwise, I'd like to use the custom value type in C#. Maybe the situation is similar in Java?

Disclaimer: I only have very basic experience with detailed performance/memory tests, so I might have done/read it wrong.

```
// DEBUG
Iterations: 100,000,000
Log(object[]): 6348ms
Log(LogValue[]): 9112ms
Log(KVP List): 11996ms

// RELEASE
Iterations: 100,000,000
Log(object[]): 6132ms
Log(LogValue[]): 3704ms
Log(KVP List): 7525ms
```

bensigelman:
@cwe1ss @yurishkuro of course meeting a minimum bar for performance is a requirement... that said, I am more concerned with making the API unobtrusive and low-boilerplate. In Java-land that's swimming against the current, though, and perhaps it would be better to embrace the local style and do something heavier-weight.

I just pushed a different version of this PR/RFC that is a little more explicit... I think the benefit here is just what @cwe1ss was getting at with the future presence of helpers to do things like `LogField.Event(String val)` and so on... those would be harder to get right with the varargs `Object...` approach. The extra typing bums me out, but at least it would be easy to wrap this base layer in something that does more inference / requires fewer characters.

opentracing-api/src/main/java/io/opentracing/Span.java
```diff
@@ -65,25 +65,24 @@
...
-     * If specified, the payload argument may be of any type and arbitrary size, though implementations are not
-     * required to retain all payload arguments (or even all parts of all payload arguments).
-     *
-     * The timestamp is specified manually here to represent a past log event.
-     * The timestamp in microseconds in UTC time.
-     **/
-    Span log(long timestampMicroseconds, String eventName, /* @Nullable */ Object payload);
+     * Like log(LogField...), but with an explicit timestamp.
+     */
+    Span log(long timestampMicroseconds, LogField<?>... fields);
```
yurishkuro: In Go API the only way to log with explicit timestamp is to provide logs in bulk to the Finish method. What makes Java different that it requires an explicit timestamp. I do not recall ever seeing a Java logging API that explicitly accepted timestamps.

opentracing-api/src/main/java/io/opentracing/LogField.java
```diff
@@ -0,0 +1,20 @@
+package io.opentracing;
```
yurishkuro: suggest moving to log package

opentracing-api/src/main/java/io/opentracing/LogField.java
```diff
@@ -0,0 +1,20 @@
...
+
+/**
+ * LogField represents a single key:value pair in a Span.log() record.
+ *
+ * The key must always be a String. All Tracer implementations must support values bool, numeric, and String values;
+ * some may also support arbitrary Object values.
+ */
+public class LogField<V> {
```
yurishkuro: can be just Field

opentracing-api/src/main/java/io/opentracing/LogField.java
```diff
@@ -0,0 +1,20 @@
...
+    private final String key;
+    private final V value;
+
+    public LogField(String key, V value) {
```
yurishkuro: I think a static `Field.of(key, value)` method is preferable to using constructor.

dawallin:
I like this, using the custom key-value type array instead of object array makes it much clearer and more natural, at least in C#.

yurishkuro:
Well, shockingly, after 15min of googling I could not find an existing Java API for structured logging. `¯\(°_o)/¯`

adriancole:
not to be contrary, but googling structured logging java api should return
log4j and fluentd pretty quickly. If structured means map values, this
might be of interest:

https://logging.apache.org/log4j/2.0/log4j-api/apidocs/org/apache/logging/log4j/message/StructuredDataMessage.html

adriancole:
@rgoers I know this is an old issue, but seems like you were behind [structured data](https://tools.ietf.org/html/rfc5424#section-6.3) support in [log4j](https://issues.apache.org/jira/browse/LOG4J2-31) back in the day. This issue is about adding similar functionality to a distributed tracing version of a logging api. Do you have any insights looking back from now until then about the task, and maybe some modeling gotchas we might want to be aware of?

yurishkuro:
> googling structured logging java api should return log4j and fluentd pretty quickly. 

yes, but they explicitly use Maps, so I classified those into "not found" category.

adriancole:
I think we need to enumerate the qualities we think are performance-based non-starters with benchmarks. Then we focus in two areas.. what's the api we want (if we want it), and can we afford that api!

For example, there are several choices in this api design
* varargs with even indexing (ex "key1", "value1", "key2", "value2")
  * this implicitly allocates an array to wrap the pairs, and validation should at least check that it is an even number. there's no assurance someone hasn't logged the same key multiple times by api
  * read performance should be good, as this is an array: you don't need to allocate an iterator to cycle through
* same, except defining a field type (ex. Field.of("key1", "value1"),  Field.of("key2", "value2"))
  * also implicitly allocates an array to wrap the pairs. length validation not needed, but duplicates might occur. higher reference count due to the Field objects.
  * read performance should be good, as this is an array: you don't need to allocate an iterator to cycle through
* accept a map (ex. ImmutableMap.of("key1", "value1",  "key2", "value2"))
  * performance characteristics are punted to the implementation for the most part
    * the cost of allocating the map will vary wildly, ex key hashing etc, many might use guava to help with that.
  * read performance is also impl dependent. special-cased maps will be array-backed, but still require allocating an iterator.
  * easy to integrate with other languages like clojure which can convert kwargs to a map
* make a builder object that internalizes the details (ex. LogBuilder logWithFields(timestamp).put("key1, "value1")
  * allocates a builder object, and hides how to be efficient or not.

These types of concerns are so provable with JMH, it would seem best to use benchmarking instead of conjecture about what should be tossed on account of performance.

adriancole:
I guess there's one more design option, a Map-like interface (similar to text-map reader thing)

```java
interface/class Fields<K, V> extends Iterable<Map.Entry<K,V>> {
... internalize your favorite backing choice, likely an array or a delegate
  static <K, V> Fields forMap(Map<K, V> map) {
.. convert
  }
}
```

rgoers:
Since you asked... although I feel like I am walking into the middle of a discussion where I don't have all the background.

Yes, structured data (namely, support for RFC 5424) was one of the things that motivated me to start work on Log4j 2.  If you look at the Log4j 2 API you will see that it really logs Messages. The RFC 5424 Message is just one of the supported message types. The advantage of this is that each Message type can be tuned for the data it is manipulating.

The discussion around creating objects is interesting. We have run a lot of benchmarks that have shown that Java has significantly improved in dealing with garbage, especially when the objects are created and dereferenced quickly. That said, Log4j has users who require a garbage-free logging framework so efforts were made in our latest releases to make it possible to use Log4j in specific documented ways without creating any garbage. What is interesting is that in some cases the throughput is slightly lower when this is enabled. The Log4j web site has a few pages that document this.

I am not at all familiar with opentracing but I would be looking at the data format(s) you want to use for transmitting the data. I would also be looking at how the data gets passed to you. Creating an efficient data structure is great but not if the API to manipulate it sucks. For example, if you use an array does the user somehow have to know how big the array needs to before they start adding to it?

bensigelman:
@rgoers thanks for chiming in. I'm not terrible surprised to hear that there's a tradeoff between logging throughput and GC churn (which is really just another flavor of latency)...

@adriancole I like the idea of using benchmarks to *eliminate* certain options. You're not suggesting this, but I want to make it clear that I don't want to use benchmarks—esp microbenchmarks—to *select* the "best" option... It should also be noted that OpenTracing impls will often have some sort of sampling regimen in place; that would suggest a bias towards APIs that minimize the amount of work done before the function call actually takes place (e.g., it would be a bummer to construct a fully-formed Map in the caller just to have it dropped by the (unsampled) Span impl below).

I'll let others offer any thoughts they have before trying to move forward with this (probably tomorrow sometime).

adriancole:
Thx for the reply, Ralph! Lots to think about, just mentioning one thing I
think I misled

> For example, if you use an array does the user somehow have to know how
> big the array needs to before they start adding to it?
>
^^ I was just calling out that varargs is essentially an array. Likely the
intent wasn't to make people think in arrays.

adriancole:
>
> @adriancole <https://github.com/adriancole> I like the idea of using
> benchmarks to *eliminate* certain options. You're not suggesting this,
> but I want to make it clear that I don't want to use benchmarks—esp
> microbenchmarks—to *select* the "best" option..
>
^^ great point! the api should be the best we can afford (benchmarks can
rule out things that are unaffordable)

opentracing-api/src/main/java/io/opentracing/LogField.java
```diff
@@ -0,0 +1,20 @@
+package io.opentracing;
+
+/**
+ * LogField represents a single key:value pair in a Span.log() record.
+ *
+ * The key must always be a String. All Tracer implementations must support values bool, numeric, and String values;
```
cwe1ss: "support values bool, ... values" or just "support bool, ... values"?

opentracing-api/src/main/java/io/opentracing/LogField.java
cwe1ss: Should this throw if key is null or is checking this up to the trace implementation?

opentracing-api/src/main/java/io/opentracing/Span.java
cwe1ss: I like it because it's consistent with Finish etc. Also, the finish method doesn't accept logs right now.

bensigelman:
I thought this through and did some reading through similar sorts of (logging) libraries. I like the `Field.of` + `log(Field...)` approach more than the `log(Map)` approach, but only slightly/aesthetically. That's what I've gone with here.

The thing I *do* like about `log(Map)` is that the opentracing artifact itself doesn't need to get into the business of defining a type for the log parameters. In that way, we *could* start with `log(Map)` and maybe/maybe-not generalize to support `log(Field...)` in the future.

The best justification I can come up with for `log(Field...)` is consistency with other OT languages and arguably cleaner (or at least more explicit) calling code.

PTAL in any case. I fixed the non-api artifacts and tests to illustrate what the fallout looks like for tracers (not too bad IMO).

opentracing-impl-java8/src/main/java/io/opentracing/AbstractSpan.java
```diff
@@ -127,6 +125,18 @@ public String getBaggageItem(String key) {
     }
 
     @Override
+    public final Span log(Field... fields) {
+        return log(System.nanoTime() / 1000, fields);
```
yurishkuro: probably setting a bad example with this, from https://docs.oracle.com/javase/7/docs/api/java/lang/System.html#nanoTime():

> This method can only be used to measure elapsed time and is not related to any other notion of system or wall-clock time. The value returned represents nanoseconds since some fixed but arbitrary origin time (perhaps in the future, so values may be negative). The same origin is used by all invocations of this method in an instance of a Java virtual machine; other virtual machine instances are likely to use a different origin.

yurishkuro:
We haven't considered one obvious (in Java) approach of overloaded methods with 2, 4, 6, ... parameters, similar to Guava's `ImmutableMap.of`, which goes [up to 10 args](https://google.github.io/guava/releases/snapshot/api/docs/com/google/common/collect/ImmutableMap.html#of(K,%20V,%20K,%20V,%20K,%20V,%20K,%20V,%20K,%20V)):

```
of(K k1, V v1, K k2, V v2, K k3, V v3, K k4, V v4, K k5, V v5)
```
It's a zero-allocation approach, used by most logging frameworks. We could start with that plus `log(Map<String, Object>)` for higher dimensions and punt on `Field` and other flavours.

cwe1ss:
@yurishkuro Could you elaborate on this? Would this be used instead of the `Field<?>...` type on the `Tracer.Log` method or would this just be a static method on `Field` that returns a list/array of fields - which will then be passed to the current method? If it's the latter, how would it result in less allocations?

yurishkuro:
I was referring to overloaded methods on the Span

```
log(String k1, Object v1, ... up to 5 times)
```

cwe1ss:
ok thx! While being better than just an object array, I still feel that the possible perf/gc gain is not worth the more complex implementation. My main concern in this case is the usage for trace implementations that (while probably still being faster) would have to do pretty weird stuff to e.g. loop through all keys etc.

Isn't this something that should be introduced when there's a real (measured) performance problem with it? Or are allocations/garbage collections really that problematic in Java?

I hope I'm not being naive and not worrying about perf enough?!?

bensigelman:
@yurishkuro re your suggestion:

> `of(K k1, V v1, K k2, V v2, K k3, V v3, K k4, V v4, K k5, V v5)`

Do you really mean `V`? It doesn't seem useful to constrain all of the values to have the same type. So `V` would just become `Object`. As such, I'd rather we just did `Map<String, Object>` and leave the inlined/expanded versions for some later date where it seems like a meaningful improvement to fluency or performance.

Stating the obvious, the safest (or most cowardly?) thing would be to go with a `Map<String, Object>` and just add documentation about what `Object` is meant to be... this means that any performance issues are not OpenTracing's problem per se. There are plenty of fluent map-builders in the world that people could use if they so chose.

Strawman: start with `Map<String, Object>` and see how far it gets us. If we wanted something that involved less typing and more specificity, the current `log.Field` proposal seems like a good alternative, but it places more bets and in that respect is less appealing to me the more I think about it.

Someone please tell me if this is getting too conservative and if users will be annoyed to construct a `Map` every time they just want to log something.

cwe1ss:
I guess it would have to be V1, V2, V3, ...

Regarding `Map` - see my small (C#) benchmark from above - using a Map/Dictionary is *a lot* slower than a list/array in C# - not sure if it's the same in Java?
The main question here is whether we want to make sure that keys are unique?!

Is your main issue with the current `Field` proposal performance or the more verbose usage?

yurishkuro:
> Do you really mean `V`? 

I was simply quoting Guava, ours would be string,object,string,object

> Someone please tell me if this is getting too conservative and if users will be annoyed to construct a Map every time they just want to log something.

I probably would be, especially if people are not using Guava, constructing maps in Java is not fun.

bensigelman:
Sorry for the radio silence on this, I realized I've been waiting to see what people think about forcing logging users to construct a `Map` implementation inline... Obviously there are options on that front (Guava, etc), though I don't like the smell of implying that OT is unwieldy without bulky utility libraries.

The compromise approach would be to support both `Map` and something like the `Field` proposal in this PR... All things being equal, I would prefer to pick one or the other for now, though, as the smaller the API surface is the less risk we're taking going forward.

adriancole:
I'm trying to stay back on the "if we should do this" rather commenting on "what if we do this"

One point to call out is that it is likely that all JVM languages will end up using this api. Map literals are very easy to construct in other JVM languages. Ex keyword args in clojure directly interop as java maps.

If we use a non-map type, we make key-value logging harder in these languages, ex as difficult as we make it for java.

It should also be noted that guava isn't the only library that allows map literals in Java, even if it is the more popular one.

adriancole:
Another note: the difficulty of instantiating maps with standard libraries is not something undiscussed or in any way an opentracing specific concern. We can look at what's going on in the language as we have a goal to not have users thrash libraries moving forward.

Maps in Java 9 will be easier to declare due to factory methods defined in [JEP269](http://cr.openjdk.java.net/~smarks/reviews/jep269/api.20151123/)

Download java 9 today, and the following will compile.
```java
// similar to guava
Map<Integer, String> map1 = Map.of(1, "a", 2, "b");

// similar tools to assertj
Map<Integer, String> map2 = Map.ofEntries(
   Map.entry(1, "a"),
   Map.entry(2, "b")
);
```

If we wanted to support maps, and we are worried that OpenTracing users will not be able to construct maps, we can always add a trivial utility to construct cheap immutable maps (say up to 5 or less pairs). This isn't rocket science.

bensigelman:
> This isn't rocket science.

To be clear, @adriancole, I'm certainly not proposing that it's rocket science. I am wondering out loud whether it's too much of a pain in the a** for consumers of the API.

Seems that having a `Map<String, Object>` at the bottom of the API is the safest thing and probably the best place to start. I'll try to make that change shortly, hopefully I don't get interrupted :)

bensigelman:
(I got interrupted :)  But I will do this.)

adriancole:
> (I got interrupted :) But I will do this.)
I know the feeling!

Lemme know if you want a utility method to create maps (to make consumers
life easier when they aren't using any libraries to create maps).
I can whip one up fairly quickly.

adriancole:
Disadvantage of writing our own map factory impl is several hundred lines
of code and tests to maintain. This could end up with a patch release if a
bug is found and otherwise distract the issues list. Also some who care
about size might complain if they don't use it.

If we want to reduce maintenance of this option, we could make a map
factory type, but implement it with a literal copy of guava using shade. Ex
in io.opentracing.internal.guava This would be larger in size even if
minimised,  but likely without bugs.

Of course another option is to not host our own map factory or wait until
users ask for it.

opentracing-api/src/main/java/io/opentracing/Span.java
```diff
@@ -65,25 +67,34 @@
...
+     * <pre>{@code
+     span.log(
+         ImmutableMap.Builder<String, Object>()
+             .put("event", "soft error")
+             .put("type", "cache timeout")
+             .put("waited.millis", 1500)
+             .build());
+     }</pre>
+     *
+     * @param fields key:value log fields. Tracer implementations are expected to support String, numeric, and boolean
```
bensigelman: Should we specify anything about the immutability of the `fields` map?

opentracing-api/src/test/java/io/opentracing/PlaygroundTest.java
```diff
@@ -1,37 +0,0 @@
...
- */
-package io.opentracing;
-
-import java.util.Arrays;
-import java.util.concurrent.CompletableFuture;
-import org.junit.Test;
-
-import static org.assertj.core.api.Assertions.assertThat;
-
-public final class PlaygroundTest {
```
bensigelman: unrelated, but this file is a long way from compiling and (IMO) has no place in the repo.

opentracing-api/src/main/java/io/opentracing/Span.java
```diff
@@ -65,25 +67,34 @@
...
-     *
-     * The timestamp of this log event is the current time.
-     **/
-    Span log(String eventName, /* @Nullable */ Object payload);
+     * Log key:value pairs to the Span with the current walltime timestamp.
+     *
+     * <p>A contrived example (using Guava, which is not required):
+     * <pre>{@code
+     span.log(
+         ImmutableMap.Builder<String, Object>()
```
bensigelman: I was imagining that application code callers would already have Guava or something similar on hand...

opentracing-api/src/main/java/io/opentracing/log/Field.java
```diff
@@ -1,46 +0,0 @@
...
- */
-package io.opentracing.log;
-
-/**
- * Field represents a single key:value pair in a Span.log(...) call.
- *
- * The key must always be a String. All Tracer implementations must support bool, numeric, and String values; some may
- * also support arbitrary Object values.
- */
-public class Field<V> {
```
bensigelman: (and this last commit gets rid of the `Field` class entirely)

bensigelman:
If there are no objections, I'd like to merge this in 36h or so. If someone wants more time to look/think, let me know.

opentracing-api/src/main/java/io/opentracing/Span.java
```diff
@@ -111,4 +122,13 @@
      * @return this Span instance, for chaining
      */
     Span setOperationName(String operationName);
+
+    /**
+     * DEPRECATED
+     **/
+    Span log(String eventName, /* @Nullable */ Object payload);
```
adriancole: if deprecated, probably better to say what the expected replacement is. We can't ignore that most dapper systems have a plain single-arg event/annotation type. This leaves someone with a mysterious question we we now suggest logging with maps isn't only an option, but the preferred (knowing it is very rarely used in the wild)

opentracing-api/src/main/java/io/opentracing/Span.java
```diff
@@ -65,25 +67,34 @@
...
+             .put("type", "cache timeout")
+             .put("waited.millis", 1500)
+             .build());
+     }</pre>
+     *
+     * @param fields key:value log fields. Tracer implementations are expected to support String, numeric, and boolean
+     *               values; some may also support arbitrary Objects.
+     * @return the Span, for chaining
+     */
+    Span log(Map<String, Object> fields);
```
adriancole: realizing that this is supposed to end up the only way to log spans, you should probably accept `Map<String, ?>`, since there's no expectation of writing to the input map.

opentracing-api/src/main/java/io/opentracing/Span.java
```diff
@@ -111,4 +122,15 @@
      * @return this Span instance, for chaining
      */
     Span setOperationName(String operationName);
+
+    /**
+     * @deprecated use {@link #log(Map)} like this
+     * {@code span.log(Map.of("event", "timeout", "millis", 1500))}
```
adriancole: this is close, but not the best example, as the doc is on a single-event method. I'd leave out the "millis"

bensigelman:
@adriancole great suggestions, PTAL

opentracing-api/src/main/java/io/opentracing/Span.java
```diff
@@ -111,4 +122,19 @@
      * @return this Span instance, for chaining
      */
     Span setOperationName(String operationName);
+
+    /**
+     * @deprecated use {@link #log(Map)} like this
+     * {@code span.log(Map.of("event", "timeout"))}
```
adriancole: is there a constant file somewhere where we can define "event"? seems equal or more important than some of the trace keys already defined, as this tells the user how to just log a simple string.

bensigelman: We would like to define something akin to https://github.com/opentracing/opentracing-java/blob/master/opentracing-api/src/main/java/io/opentracing/tag/Tags.java for logging. I had hoped to get the basic KV mechanism in place first, then gather input about logging keys that ought to be standardized and promoted in this way.

adriancole: basically, we have a responsibility to think about general implementors. currently our option is to assume a single-entry map should be rendered as a plain string. The problem with axing the plain api is that now we have to make guesses for the most form of trace annotations: a simple string.

If we make a constant, we don't pollute existing systems with {"event":"foo"}, or require users of the generic api to always read the notes of each tracer impl when trying to figure out how to log a simple string

adriancole:
still very much dislike this as the only api. made a suggestion to make it less bad. It would be nice to have more java people on this thread.

bensigelman:
@adriancole 

> still very much dislike this as the only api.

Just to make sure I understand, would your concern be addressed by a `span.log(io.opentracing.Logging.eventOf("event name"))` API, or are you suggesting that something like `Span.logEvent(String)` sticks around in non-deprecated form?

adriancole:
knowing that we want to keep trunk shippable, punting on this runs the risk
of releasing without defining a predictable way of doing what was done
before.

I understand that there's a bike shed to be had about which key is best,
but I'd argue this.

don't deprecate the old method or even suggest a replacement until there's
a valid predicatable replacement. This puts the burden on those championing
the Map api with responsibility of making it usable.

bensigelman:
@adriancole your concerns are totally reasonable re standardizing on a key before deprecating the old way of doing things. Though it would be true to my heritage at google... what was the quote? "There's the way that's deprecated and the way that's not ready yet." Something like that.

Still curious what in particular you had in mind (per https://github.com/opentracing/opentracing-java/pull/47#issuecomment-248808946)... I would prefer a `Logging` namespace/class that would eventually house a bunch of different static methods that make it easier to work with the key-value `Map` underneath.

adriancole:
What I was thinking is that is should be possible to make a wrapper that
keeps the span simple and not pollute systems like zipkin with unsearchable
double encoded json or having to present awkward nested things in the UI.
Lacking a way to make OT simple, I frankly wouldn't suggest people use it.

In order to make it simple (or possible to be simple), we'd need to define
some way of saying this is a normal annotation (event is a narrowing of
that). Basically a string that is the only component to put on a timeline..
regardless of whether it is an "event" or not. In logging terms, the
message parameter.

How that happens I don't care much I just don't want to be in a position
where recommending OT is the same as recommending a 3 headed hammer. We
need to be portable with dapper annotations.

adriancole:
@bensigelman not sure why I can't reply to your question about preference. If `Span.logEvent(String)` stuck around, I wouldn't be tempted to make a wrapper to create that api.

bensigelman:
@adriancole the new github review thing has some weird rules about *where* you're allowed to reply to things. /shrug.

I'll propose something later today. Maybe we'll leave `logEvent(String)`, though I def want to `@deprecate` `logEvent(String, Object)`.

bensigelman:
@adriancole PTAL after the latest commit. TL;DR, the developer can now write

```
span.log(io.opentracing.Logs.event("soft error"))
```

opentracing-api/src/main/java/io/opentracing/Logs.java
```diff
@@ -0,0 +1,56 @@
...
+     * MESSAGE_KEY identifies a String log value that's human-readable and descriptive. Messages need not be stable.
+     */
+    public static final String MESSAGE_KEY = "message";
+
+    /**
+     * @param eventName the value for the "event" key in the log map
+     * @return a Map suitable for use with @link{io.opentracing.Span#log(Map)}
+     * @see Logs#EVENT_KEY
+     */
+    public static Map<String, ?> event(String eventName) {
```
bensigelman: note that these helpers do *not* act like builders... they only cover the case of a Map with a single entry, which is the common case @adriancole is concerned about. I was imagining that multi-entry maps can make use of the `EVENT_KEY` constants... a full-on Builder is also an option but is overkill right now IMO.

adriancole: FYI singleton maps are in all recent versions of java `Collections.singletonMap()`

so if we have a stable event or message key defined, then you could just as well suggest
`Collections.singletonMap(LogKeys.EVENT, "mary had a little lamb")`.

That is unless the goal is to eventually extend map with log view types, ex. 
```
class Event extends MapAdapter<String, String> {
  ...
}
```
In that case, you might want to direct them to a file like this.

I did this in Denominator, mainly to allow one to use rare or undefined DNS rdata types. It works, mostly, but has some limitations. https://github.com/Netflix/denominator/blob/master/model/src/main/java/denominator/model/rdata/SRVData.java#L23
https://github.com/Netflix/denominator/blob/master/model/src/main/java/denominator/model/ResourceRecordSets.java#L559

adriancole: if interested in the map-view thing in documentation form https://github.com/Netflix/denominator/wiki/Model#record-data-rdata

adriancole: to be clear I wouldn't recommend doing this.. one of the reasons I am not for map-based things as unlike DNS, there isn't a registry of known types. So, the caller must need to know what the sender can accept (coupling caller code to the tracing system). That's why I prefer a "straight-pipe" for normal dapper-style annotations, as it is easier to map a string or string with timestamp to a map, vs encouraging a map which is more complex and not supported on many tracing systems.

adriancole:
Regardless, here's the impact of map-only expressed in finagle form..

Existing related code in finagle (noted for context.. particularly none of the annotations have anything to do with Map):
```scala
sealed trait Annotation
object Annotation {
  case object WireSend                             extends Annotation
--snip--

  def record(record: Record): Unit = {
    record.annotation match {
      case tracing.Annotation.WireSend =>
        annotate(record, thrift.Constants.WIRE_SEND)
--snip--
```

How you would make opentracing work in finagle today.
```scala
  // existing signature
  protected def annotate(record: Record, value: String): Unit = {
    spanMap.update(record.traceId) { span =>
      // change the internals to have a map of opentracing spans, and use the log method and of course pass null for payload
      span.log(record.timestamp.inMicroseconds, value, null))
    }
  }
```

How you would make opentracing work in finagle when changed to a map.
```scala
  protected def annotate(record: Record, value: String): Unit = {
    spanMap.update(record.traceId) { span =>
      // change the internals to have a map of opentracing spans, and use the log method
      span.logKV(record.timestamp.inMicroseconds, Map("annotation or event" -> value), null))
    }
  }
  // change the recorder to unwrap that map to extract the "annotation or event" key
```

bensigelman:
@adriancole and I were chatting about something else (off github) and ended up discussing this PR for a bit.

To try to summarize, the `Map` approach as the *only* way to log makes the one existing/common case (i.e., logging an event) unwieldy. We revisited the idea of using a `Field`-centric API, we thought about just doing `log([ts,] key, value)` for various value types and telling the user to chain calls with the same timestamp to emulate Map behavior, and also considered alternative ways to handle the common case of "just logging a timestamped annotation/event".

Anyway, I've just pushed something that removes the `Logs.event` map-building helper from the previous commit, and adds `Span.log([long timestampMicros,] String eventName)` methods to handle the common case of logging a single event string. This is a concession to the numerous pieces of JVM/Zipkin instrumentation that expect something like this already.

I've also put some "caution tape" around the Map-based API since I think the thrashing on this PR suggests that the precise details of the interface may change with more experience.

Anyway, PTAL, anyone who cares. I'd still like to merge this soon.

opentracing-api/src/main/java/io/opentracing/Span.java
```diff
@@ -65,25 +67,70 @@
     Span setTag(String key, Number value);
 
     /**
-     * Add a new log event to the Span, accepting an event name string and an optional structured payload argument.
+     * Log key:value pairs to the Span with the current walltime timestamp.
+     *
+     * CAUTIONARY NOTE: not all Tracer implementations support key:value log fields end-to-end. Caveat emptor.
```
adriancole: `<p><strong>CAUTIONARY NOTE:</strong>`

opentracing-api/src/main/java/io/opentracing/Span.java
```diff
@@ -65,25 +67,70 @@
...
+     * <pre>{@code
+     span.log(timestampMicroseconds, Collections.singletonMap("event", event));
+     }</pre>
+     *
+     * @param timestampMicroseconds The explicit timestamp for the log record. Must be greater than or equal to the
+     *                              Span's start timestamp.
+     * @param event the event value; often a stable identifier for a moment in the Span lifecycle
+     * @return the Span, for chaining
+     */
+    Span log(long timestampMicroseconds, String event);
```
adriancole: 👍

adriancole:
:shipit: happy with this as it leaves the simple event in-tact, while letting folks experiment with more complex ones. Will be interesting to see how this is used

opentracing-impl-java8/src/main/java/io/opentracing/AbstractSpan.java
```diff
@@ -127,18 +123,46 @@ public String getBaggageItem(String key) {
     }
 
     @Override
-    public final Span log(String message, /* @Nullable */ Object payload) {
+    public final Span log(String event) {
+        return log(System.nanoTime() / 1000, event);
```
yurishkuro: this should be `System.currentTimeMillis() * 1000` instead. 

`nanoTime()`:

> This method can only be used to measure elapsed time and is not related to any other notion of system or wall-clock time. The value returned represents nanoseconds since some fixed but arbitrary origin time (perhaps in the future, so values may be negative). The same origin is used by all invocations of this method in an instance of a Java virtual machine; other virtual machine instances are likely to use a different origin.

bensigelman: yeah, I noticed this the other day, too.

With OT-Java impls I've done (which need to be Java 7 compatible) we've done nanoTime() deltas for duration and millis*1000 for the start timestamp.

bensigelman: @yurishkuro I replaced `nano/1000` with `millis*1000`... not really relevant to this PR, but such a small change I figured I'd include it.

(In AbstractSpan I was able to use `Instant` which obviates the problem... though this points to additional problems with the whole `AbstractSpan` concept. I would really like to get that whole artifact out of this particular github repo, but that's its own discussion.)

bensigelman:
@adriancole assuming this LGTY, please update the approval bit in github since the new review process has a big red "changes requested" and I would feel dirty ignoring that.

opentracing-impl-java8/src/main/java/io/opentracing/AbstractSpan.java
yurishkuro: Thanks - it just felt wrong leaving that in the code that could be considered "an example"

bensigelman:
Thanks for the discussion, all... merging.

michaelsembwever:
@bensigelman @adriancole can we please squash commits before merging PRs. there's no need for the flurry of checkpoint commits in the mainline.

bensigelman:
@michaelsembwever I'll enable that option for the repo. I would like to continue to support the option of merging individual commits since sometimes that's what makes the most sense.

bensigelman:
Oh -- looks like those options are already enabled.

image

Meh. @michaelsembwever if you feel strongly I can rewrite history and squash these commits down after the fact.

michaelsembwever:
Yes, cleaning up the commits as the final steps in a PR is fantastic.
Commits should be atomic, typically one per PR but not necessarily.

cwe1ss:
I updated the C# PR and included a few .NET specific changes. I'd appreciate your feedback!

see [this comment](https://github.com/opentracing/opentracing-csharp/pull/13#issuecomment-253982390) for the new syntax.

PR #50, opened 2016-09-07
Add Span#setOperationName

### LINKED ISSUE #48: Add possibility to set the operation name after the span has been created
I'm currently evaluating to use opentracing/jaeger as the foundation of stagemonitor and this is one of the things I've stumbled upon:

As of now, the operation name has to be provided in order to create a span:

```java
Span http = tracer.buildSpan("HandleHTTPRequest")
```

The problem is that you often don't know the operation name when a HTTP request has just arrived. Setting the operation name to the URL may be sufficient in some cases but as soon as you have url schemes like `/item/{itemNumber}` you probably want to group them together.

Stagemonitor instruments Spring MVC and JAX-RS to derive the operation name from the controller method. For example 

```java
@RequestMapping("/item/{itemNumber})
public void showItemDetail(@PathVaiable String itemNumber) {...}
```

Results in an operation name like `Show Item Details` or `ItemControler#showItemDetails`. Users also may choose to set the operation name at any time in stagemonitor. Note that the Span does not start at the controller method but as soon as the HTTP request hits the topmost servlet `Filter`.

So I think it would be a useful addition to be able to set the operation name after the start of the span.

felixbarny:
(closes #48)

### DIFF
--- a/opentracing-api/src/main/java/io/opentracing/Span.java
+++ b/opentracing-api/src/main/java/io/opentracing/Span.java
@@ -104,4 +104,11 @@ public interface Span extends AutoCloseable {
      * @return the value of the baggage item identified by the given key, or null if no such item could be found
      */
     String getBaggageItem(String key);
+
+    /**
+     * Sets the string name for the logical operation this span represents.
+     *
+     * @return this Span instance, for chaining
+     */
+    Span setOperationName(String operationName);
 }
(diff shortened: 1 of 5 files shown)


yurishkuro:
LGTM

michaelsembwever:
merged. thanks @felixbarny

Issue #45, opened 2016-08-19
Move Noop tracer to its own module opentracing-noop


yurishkuro:
Noop tracer is currently in the `-impl` module, which has the following issues:
- `-impl` module was meant to contain an abstract skeleton for tracer implementations
- Noop tracer is never going to use that skeleton, so bundling it in the same jar increases the end user's dependencies

@bensigelman I know you prefer for it to be in `-api`, but (a) it's a harder sell, (b) taking a more pedantic view, in the world where everything is DI-ed and "sensible defaults" are considered an anti-pattern (e.g. an echo http server on netty requires 100+ lines of code), an instrumentation of a framework technically does not need the Noop tracer, only the API. So I think it's cleaner to have is as a separate artifact.

michaelsembwever:
Can we come back to this at a later point? Seems like pre-mature optimisation, or just jumping the gun.

There no dependency tree introduced by anything here. 
In fact `-impl` has no dependencies what so ever already.
I can't see any immediate problem or debt.

bhs:
@yurishkuro this is done, right? Closing...

PR #56, opened 2016-09-28
A child of a NoopSpan is a NoopSpan.

michaelsembwever:
If a NoopSpan or Noop SpanContext is declared as a parent to the span during the SpanBuilder usage, then return the NoopSpanBuilder.

### DIFF
--- a/opentracing-api/src/test/java/io/opentracing/AssertionTest.java
+++ b/opentracing-api/src/test/java/io/opentracing/AssertionTest.java
@@ -1,0 +1,33 @@
+/**
+ * Copyright 2016 The OpenTracing Authors
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License"); you may not use this file except
+ * in compliance with the License. You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software distributed under the License
+ * is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express
+ * or implied. See the License for the specific language governing permissions and limitations under
+ * the License.
+ */
+
+package io.opentracing;
+
+import org.junit.Test;
+
+public final class AssertionTest {
+
+    @Test
+    public void test_assertions_enabled() {
+        boolean asserted = false;
+        try {
+            assert false;
+        } catch (AssertionError error) {
+            asserted = true;
+        }
+        if (!asserted) {
+            throw new AssertionError("assertions are not enabled");
+        }
+    }
+}
--- a/opentracing-impl-java8/src/main/java/io/opentracing/AbstractSpanBuilder.java
+++ b/opentracing-impl-java8/src/main/java/io/opentracing/AbstractSpanBuilder.java
@@ -13,6 +13,7 @@
  */
 package io.opentracing;
 
+import io.opentracing.Tracer.SpanBuilder;
 import java.time.Instant;
 import java.util.ArrayList;
 import java.util.HashMap;
@@ -51,12 +52,18 @@ abstract class AbstractSpanBuilder implements Tracer.SpanBuilder {
     }
 
     @Override
-    public final AbstractSpanBuilder asChildOf(SpanContext parent) {
+    public final SpanBuilder asChildOf(SpanContext parent) {
+        if (NoopSpanBuilder.INSTANCE == parent || NoopSpan.INSTANCE.context() == parent) {
+            return NoopSpanBuilder.INSTANCE;
+        }
         return this.addReference(References.CHILD_OF, parent);
     }
 
     @Override
-    public final AbstractSpanBuilder asChildOf(Span parent) {
+    public final SpanBuilder asChildOf(Span parent) {
+        if (NoopSpan.INSTANCE == parent) {
+            return NoopSpanBuilder.INSTANCE;
+        }
         return this.addReference(References.CHILD_OF, parent.context());
     }
 
--- a/opentracing-impl-java8/src/test/java/io/opentracing/AbstractTracerTest.java
+++ b/opentracing-impl-java8/src/test/java/io/opentracing/AbstractTracerTest.java
@@ -89,6 +89,23 @@ public final class AbstractTracerTest {
         assertEquals("Should find the marker", "whatever", ((TestSpanBuilder)result).operationName);
     }
 
+    @Test
+    public void testExtractAsParent() throws Exception {
+        Map<String,String> map = Collections.singletonMap("test-marker", "whatever");
+        TextMapExtractAdapter adapter = new TextMapExtractAdapter(map);
+        AbstractTracer tracer = new TestTracerImpl();
+        SpanContext parent = tracer.extract(Format.Builtin.TEXT_MAP, adapter);
+        assert NoopSpan.INSTANCE != tracer.buildSpan("child").asChildOf(parent).start();
+    }
+
+    @Test
+    public void testExtractOfNoParent() throws Exception {
+        AbstractTracer tracer = new TestTracerImpl();
+        assert NoopSpan.INSTANCE == tracer.buildSpan("child").asChildOf(NoopSpan.INSTANCE).start();
+        assert NoopSpan.INSTANCE == tracer.buildSpan("child").asChildOf(NoopSpan.CONTEXT).start();
+        assert NoopSpan.INSTANCE == tracer.buildSpan("child").asChildOf(NoopSpanBuilder.INSTANCE).start();
+    }
+
     final class TestTracerImpl extends AbstractTracer {
 
         static final String OPERATION_NAME = "operation-name";


yurishkuro:
-1

There has never been a requirement (across all languages) that a tracing implementation is required to deal with receiving Noop or any other incompatible span/context types. Pretty much everywhere the assumption has been a tracer only sees its own types and seeing anything else is a configuration error. This change sets an opposite precedent.

michaelsembwever:
This is a bug fix, not an improvement.

It's not needed as a requirement, it is needed to make the fluent API actually work.

The argument that implementations are only required to deal with their own types does not hold.
If an implementation chooses to build on top of *this java abstract implementation* then deal with Noop it must. The noop component and the abstract component were never meant to be separated as such, the separation was only between Java7 and Java8. This is evident in the naming of the components, as well as the dependency of one upon the other: https://github.com/opentracing/opentracing-java/blob/master/opentracing-impl-java8/pom.xml#L39

So to make clear: nothing here pollutes how the top level API works. 
The compatibility break here is  `AbstractSpanBuilder.childOf(..)` now returns `SpanContext` instead of `AbstractSpanContext`.

An illustration of how this is required is better shown in https://github.com/openzipkin/brave-opentracing/pull/5 specifically this test case: https://github.com/openzipkin/brave-opentracing/pull/5/files#diff-6c41a867ac46d2b874b3e7ee714271a3R340

The point of the fluent API is that library and framework code is expected to simply instrument with opentracing code, unconcerned as to whether the particular request is being traced or not.
For example

    public Response serverProcessingRequest(Requestr request) {
        TextMapExtractAdapter adapter = new TextMapExtractAdapter(request.getHeaders());
        SpanContext parent = tracer.extract(Format.Builtin.TEXT_MAP, adapter);
        try (Span span = tracer.buildSpan("some-name").asChildOf(parent).start()) {
            // server processes request
            return response;
        }
    }

The abstract implementation needs to return a noop builder (in some shape or form) when the incoming trace has nothing to extract. This has little to do with precedence within implementation details in other languages, but about fixing a bug in this implementation.

I intentionally haven't touched on the forcing tracing where extract returns a noop. That is functionality to add, and given it has not yet been dealt with it's outside of the scope of this PR.

Neither I have dealt with the issue of dealing with multiple parents. But again I would rather fix what's there first, and deal with improvements afterwards.

TL;DR the argument that the abstract-java7 and abstract-java8 are separate helper implementations does not hold, they form the one abstract helping implementation.

yurishkuro:
> The abstract implementation needs to return a noop builder (in some shape or form) when the incoming trace has nothing to extract. 

If we're talking about Extract, it needs to return something that can be passed to childOf() reference. It can be null, or an empty SpanContext (!= NoopSpanContext), but it never needs to be a builder. A builder is only returned from startSpan() method, and a real tracer never needs to return noop builder from startSpan().

For example, our tracer implementation (https://github.com/uber/jaeger-client-java) is fully functional without any dependency on `opentracing.NoopXXX`, so it **can** be done.

michaelsembwever:
> For example, our tracer implementation (https://github.com/uber/jaeger-client-java) is fully functional without any dependency on opentracing.NoopXXX, so it can be done.

No. You would have had to explicitly exclude the noop transitive dependency.
Like i said, Noop was never something separate. It was always intended as an integral part of the abstract implementation. In hindsight separate java7 and java8 was a mistake as it lead to this misunderstanding.

> If we're talking about Extract, it needs to return something that can be passed to childOf() reference. It can be null, or an empty SpanContext (!= NoopSpanContext), but it never needs to be a builder. A builder is only returned from startSpan() method, and a real tracer never needs to return noop builder from startSpan().

A NoopSpanContext can be passed to childOf(..).
A NoopSpanContext is considered the same as an empty SpanContext. That is, no parent or context exists. What other interpretation does a NoopSpanContext have?
None of the childOf(..) methods explicitly accept a builder.

yurishkuro:
> No. You would have had to explicitly exclude the noop transitive dependency.

The jar dependency we have is for NoopTracer, it's returned from configuration if the user asks to disable Jaeger Tracer. But our code has no dependency on NoopSpan/Context: https://github.com/uber/jaeger-client-java/search?utf8=%E2%9C%93&q=NoopSpan

> Like i said, Noop was never something separate. It was always intended as an integral part of the abstract implementation.

This is where we disagree. I even have an issue #45 opened for this. Noop implementation is the default that instrumentations/frameworks can fall back to if the end user does not configure them with a real tracer. OT APIs in other languages all provide such Noop tracer, in most cases the noop tracer is actually a part of the "-api" module.

bensigelman:
I'm not sure if it's confusion around terminology (or what), but further to what @yurishkuro is saying: the Noop impls (of Tracer, Span, SpanContext) are not intended to mix-and-match with impls from "real" (non-Noop) OpenTracing implementations. This is how it works – and works well – in other languages.

@michaelsembwever I think the fundamental thing here is that you still don't want to pass `null` around to represent a missing SpanContext... I can see that point of view, but if we want to address that issue, it should be done with something other than a NoopSpanContext.

(Tangential to this, I regret making the SpanBuilder inherit from SpanContext... it doesn't really cause problems with implementations, but I think it's misleading as far as the SpanContext concept is concerned... I don't mean to fork the thread, but I want to preemptively argue against anything that would extend that SpanBulider-is-a-SpanContext concept.)

michaelsembwever:
Ok, lets make this very clear to all. In this abstract implementation there was every intention to mix-and-match the noop and impls. To tell me, the author, otherwise i take offence to.

That this was not communicated clearly, or was misunderstood, or that implementations in other languages have taken other approaches is beside the point to the previous sentence. Where we stand now is that this agreement or stance did not previously exist in this code. 

To run through it, this intent is made obvious in a number of ways
 - the project structure (and discussions around it) are about java7 vs java8. Not about noop vs impls.
 - the discussion about moving the project structure to noop vs impls was raised and has not been approved, (which suggests original intent must have been something else).
 - the java8 subcomponent clearly is built (depends upon) the java7 subcomponent. This was the original discussion and intent of the project structure and afterwards it seems to have been hijacked by others without it being obvious to the author.
 - an api design that avoids passing around nulls, as is the java standard, exists.

We have to keep this realistic to what it is, not to what you hope it to be. (And i'm not against moving things one step at a time into what you hope it to be, but that's right now out of context.)

Moving forward. 
 - I don't really see the value of moving the Noop classes up to `-api`. Noop isn't a public concern, it's just adds something useful to implementing a fluent api. But that is a concern outside of the scope of this bug fix.
 - I can see the value of not permitting mix-and-match as it reduces the complexity in implementations. I don't get through how noop adds value beyond only the one example use case. As is shown here there a use case where an empty extracted context benefits (i'm not sure how to solve that without the mix-match though). But that is a concern outside of the scope of this bug fix.
 - I can see the value of merging together `-impl` and `-impl-java8`. That removes the confusion and allows us to move forward.

For now what i propose is to do the latter, merge together `-impl` and `-impl-java8`.
This leaves the code in the state as it was intended, and this patch will appear clearly as the bug fix it is. I'm happy to rename them from Noop to something else if we feel that "noop" is a taken terminology with a limited usage, as you write @bensigelman "something other than a Noop…". I'm also happy to make Noop subclass the abstract classes to further demonstrate their intended place in the `-impl` subcomponent (the problem with this would be that you can't then pull the noop classes up to the `-api` level).

Then from there we can evaluate whether Noop classes, for non-mix-and-match use, are brought, copied up to, or something clean implemented in, the `-api` level.

For now I just want the code to not throw a rather stupid exception.

michaelsembwever:
i'm rewriting the previous comment. keep me a second.

michaelsembwever:
Re-implemented in https://github.com/opentracing/opentracing-java/pull/58
 addressing the concerns around mix-and-matching.

yurishkuro:
> In this abstract implementation there was every intention to mix-and-match the noop and impls. To tell me, the author, otherwise i take offence to.

@michaelsembwever no offence intended, but you have repeatedly dismissed the comments about "other languages", which is quite upsetting and leads to this exact misunderstanding of the role of Noop classes. One of the primary goals of OpenTracing initiative is to provide a way of doing things _in the same conceptual way across all languages_. What we have today is "all these languages do it the same, and Java does it differently, because (period)." In all other languages the Noop tracer is a part of the official API. If I am a developer of some framework/library/application that I am instrumenting with opentracing, I have three options when it comes to configuration
  1. use OT API assuming it's always initialized by the end user and SEGFAULT/NPE if the tracer is null
  1. write my code with `if (tracer == null)` checks all other the place to tolerate uninitialized tracer
  1. default the tracer to Noop and let the end users override/configure my lib with a real impl if they want, so that they can use my framework with or without tracing.

I think the easiest way out of this is to resolve #45 and move the current noop classes into their own `opentracing-noop` module. Then, at least from my point of view, you can do whatever with the -impl module, since it becomes just yet another implementation and does not affect how people use the API.

michaelsembwever:
> but you have repeatedly dismissed the comments about "other languages", 

any dismissal was only meant in a temporary context.

I hear what you say and completely agree with moving the code towards the general patterns within opentracing. Frustration existed because this was being used to block an immediate bug fix.

> I think the easiest way out of this is to resolve #45 and move the current noop classes into their own opentracing-noop module.

This has been done in https://github.com/opentracing/opentracing-java/pull/58

PR #58, opened 2016-10-01
Fixes class cast exception , and Solves #45

michaelsembwever:
A child of a NoopSpan is a NoopSpan (without explicit forced sampling).
If a NoopSpan or Noop SpanContext is declared as a parent to the span during the SpanBuilder usage, then return the NoopSpanBuilder. This has been re-implemented so that a) Noop objects are separate entities in the opentracing landscape which can not be mix-and-matched, and b) can re-use them by concept so to enable fluent APIs (ie avoiding nulls being passed around).

**Actions**
Merge impl and impl-java8 to avoid confusion regarding their roles.
Move Noop classes to their own component "opentracing-noop" where they can't be confused to be part of, or mix-and-matched with, the impl classes.
Move the Noop classes into interfaces, while creating default implementations of this. This (along with using the create() factory method name) indicates that this is a marker interface and the instance only should not be compared against.

Opentracing-impl subclasses (extends) the Noop interfaces to provide an empty (or None) instance to the abstract classes provided. This prevents mix-and-match. Such an approach can be repeated by specific implementations too if they like.

**References**
Addressing the ClassCastException that is currently thrown when extract(..) is used.
Also solves https://github.com/opentracing/opentracing-java/issues/45

Supersedes https://github.com/opentracing/opentracing-java/pull/56

### DIFF
--- a/opentracing-noop/pom.xml
+++ b/opentracing-noop/pom.xml
@@ -20,23 +20,21 @@
     <parent>
         <groupId>io.opentracing</groupId>
         <artifactId>parent</artifactId>
-        <version>0.16.1-SNAPSHOT</version>
+        <version>0.17-SNAPSHOT</version>
     </parent>
 
-    <artifactId>opentracing-impl-java8</artifactId>
-    <name>OpenTracing-impl-java8</name>
-    <description>OpenTracing Java8 implementation</description>
+    <artifactId>opentracing-noop</artifactId>
+    <name>OpenTracing-noop</name>
+    <description>OpenTracing NoOp</description>
 
     <properties>
         <main.basedir>${project.basedir}/..</main.basedir>
-        <main.java.version>1.8</main.java.version>
-        <main.signature.artifact>java18</main.signature.artifact>
     </properties>
 
     <dependencies>
         <dependency>
             <groupId>${project.groupId}</groupId>
-            <artifactId>opentracing-impl</artifactId>
+            <artifactId>opentracing-api</artifactId>
         </dependency>
     </dependencies>
 </project>
--- a/opentracing-noop/src/main/java/io/opentracing/NoopSpan.java
+++ b/opentracing-noop/src/main/java/io/opentracing/NoopSpan.java
@@ -13,24 +13,17 @@
  */
 package io.opentracing;
 
-import java.util.Collections;
 import java.util.Map;
 
-final class NoopSpan implements Span {
+public interface NoopSpan extends Span {
+    static final NoopSpan INSTANCE = new NoopSpanImpl();
+}
 
-    static final NoopSpan INSTANCE = new NoopSpan();
+final class NoopSpanImpl implements NoopSpan {
 
-    static final SpanContext CONTEXT = new SpanContext() {
-        @Override
-        public Iterable<Map.Entry<String, String>> baggageItems() {
-            return Collections.EMPTY_MAP.entrySet();
-        }
-    };
-
-    private NoopSpan() {}
 
     @Override
-    public SpanContext context() { return CONTEXT; }
+    public SpanContext context() { return NoopSpanContextImpl.INSTANCE; }
 
     @Override
     public void finish() {}
@@ -78,3 +71,4 @@ final class NoopSpan implements Span {
     public Span setOperationName(String operationName) { return this; }
 
 }
+
--- a/opentracing-noop/src/main/java/io/opentracing/NoopTracer.java
+++ b/opentracing-noop/src/main/java/io/opentracing/NoopTracer.java
@@ -13,12 +13,22 @@
  */
 package io.opentracing;
 
-import io.opentracing.propagation.Injector;
-import io.opentracing.propagation.TextMap;
+import io.opentracing.propagation.Format;
 
-final class TestTextMapInjectorImpl implements Injector<TextMap> {
-    @Override
-    public void inject(SpanContext spanContext, TextMap carrier) {
-        carrier.put("test-marker", ((AbstractSpan)spanContext).getOperationName());
-    }
+public interface NoopTracer extends Tracer {
 }
+
+final class NoopTracerImpl implements NoopTracer {
+    final static NoopTracer INSTANCE = new NoopTracerImpl();
+
+    @Override
+    public SpanBuilder buildSpan(String operationName) { return NoopSpanBuilderImpl.INSTANCE; }
+
+    @Override
+    public <C> void inject(SpanContext spanContext, Format<C> format, C carrier) {}
+
+    @Override
+    public <C> SpanContext extract(Format<C> format, C carrier) { return NoopSpanBuilderImpl.INSTANCE; }
+
+}
+
--- a/opentracing-api/pom.xml
+++ b/opentracing-api/pom.xml
@@ -20,7 +20,7 @@
     <parent>
         <groupId>io.opentracing</groupId>
         <artifactId>parent</artifactId>
-        <version>0.16.1-SNAPSHOT</version>
+        <version>0.17-SNAPSHOT</version>
     </parent>
 
     <artifactId>opentracing-api</artifactId>
--- a/opentracing-api/src/test/java/io/opentracing/AssertionTest.java
+++ b/opentracing-api/src/test/java/io/opentracing/AssertionTest.java
@@ -1,0 +1,33 @@
+/**
+ * Copyright 2016 The OpenTracing Authors
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License"); you may not use this file except
+ * in compliance with the License. You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software distributed under the License
+ * is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express
+ * or implied. See the License for the specific language governing permissions and limitations under
+ * the License.
+ */
+
+package io.opentracing;
+
+import org.junit.Test;
+
+public final class AssertionTest {
+
+    @Test
+    public void test_assertions_enabled() {
+        boolean asserted = false;
+        try {
+            assert false;
+        } catch (AssertionError error) {
+            asserted = true;
+        }
+        if (!asserted) {
+            throw new AssertionError("assertions are not enabled");
+        }
+    }
+}
--- a/opentracing-impl/pom.xml
+++ b/opentracing-impl/pom.xml
@@ -20,7 +20,7 @@
     <parent>
         <groupId>io.opentracing</groupId>
         <artifactId>parent</artifactId>
-        <version>0.16.1-SNAPSHOT</version>
+        <version>0.17-SNAPSHOT</version>
     </parent>
 
     <artifactId>opentracing-impl</artifactId>
@@ -29,12 +29,18 @@
 
     <properties>
         <main.basedir>${project.basedir}/..</main.basedir>
+        <main.java.version>1.8</main.java.version>
+        <main.signature.artifact>java18</main.signature.artifact>
     </properties>
 
     <dependencies>
         <dependency>
             <groupId>${project.groupId}</groupId>
             <artifactId>opentracing-api</artifactId>
+        </dependency>
+        <dependency>
+            <groupId>${project.groupId}</groupId>
+            <artifactId>opentracing-noop</artifactId>
         </dependency>
     </dependencies>
 
--- a/opentracing-impl/src/main/java/io/opentracing/impl/AbstractSpan.java
+++ b/opentracing-impl/src/main/java/io/opentracing/impl/AbstractSpan.java
@@ -11,11 +11,17 @@
  * or implied. See the License for the specific language governing permissions and limitations under
  * the License.
  */
-package io.opentracing;
+package io.opentracing.impl;
 
+import io.opentracing.Span;
+import io.opentracing.SpanContext;
 import java.time.Duration;
 import java.time.Instant;
-import java.util.*;
+import java.util.ArrayList;
+import java.util.Collections;
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
 import java.util.concurrent.TimeUnit;
 
 abstract class AbstractSpan implements Span, SpanContext {
--- a/opentracing-impl/src/main/java/io/opentracing/impl/AbstractSpanBuilder.java
+++ b/opentracing-impl/src/main/java/io/opentracing/impl/AbstractSpanBuilder.java
@@ -11,8 +11,12 @@
  * or implied. See the License for the specific language governing permissions and limitations under
  * the License.
  */
-package io.opentracing;
+package io.opentracing.impl;
 
+import io.opentracing.References;
+import io.opentracing.Span;
+import io.opentracing.SpanContext;
+import io.opentracing.Tracer;
 import java.time.Instant;
 import java.util.ArrayList;
 import java.util.HashMap;
@@ -52,12 +56,16 @@ abstract class AbstractSpanBuilder implements Tracer.SpanBuilder {
 
     @Override
     public final AbstractSpanBuilder asChildOf(SpanContext parent) {
-        return this.addReference(References.CHILD_OF, parent);
+        return io.opentracing.NoopSpanContext.class.isAssignableFrom(parent.getClass())
+                ? NoopSpanBuilder.INSTANCE
+                : this.addReference(References.CHILD_OF, parent);
     }
 
     @Override
     public final AbstractSpanBuilder asChildOf(Span parent) {
-        return this.addReference(References.CHILD_OF, parent.context());
+        return io.opentracing.NoopSpan.class.isAssignableFrom(parent.getClass())
+                ? NoopSpanBuilder.INSTANCE
+                : this.addReference(References.CHILD_OF, parent.context());
     }
 
     @Override
--- a/opentracing-impl/src/main/java/io/opentracing/impl/AbstractTracer.java
+++ b/opentracing-impl/src/main/java/io/opentracing/impl/AbstractTracer.java
@@ -11,8 +11,10 @@
  * or implied. See the License for the specific language governing permissions and limitations under
  * the License.
  */
-package io.opentracing;
+package io.opentracing.impl;
 
+import io.opentracing.SpanContext;
+import io.opentracing.Tracer;
 import io.opentracing.propagation.Extractor;
 import io.opentracing.propagation.Format;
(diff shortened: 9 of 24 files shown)


yurishkuro:
Overall I like this change, but the use of static methods in Noop interfaces makes it incompatible with Java 1.7. Can we downgrade it to 1.7? For example, by adding something like NoopTracerFactory.create(), or whichever is the most idiomatic way in Java?

opentracing-noop/src/main/java/io/opentracing/NoopTracer.java
```diff
@@ -15,19 +15,25 @@
 
 import io.opentracing.propagation.Format;
 
-public final class NoopTracer implements Tracer {
+public interface NoopTracer extends Tracer {
+    static NoopTracer create() {
+        return NoopTracerImpl.INSTANCE;
+    }
```
yurishkuro: to make this Java 1.7 compatible, we could just move the constant from L25 here.

opentracing-impl/src/main/java/io/opentracing/AbstractSpanBuilder.java
```diff
@@ -52,11 +52,17 @@ public final AbstractSpanBuilder addReference(String referenceType, SpanContext
 
     @Override
     public final AbstractSpanBuilder asChildOf(SpanContext parent) {
+        if (parent instanceof NoopSpanContext) {
+            return AbstractNoopSpanBuilder.INSTANCE;
```
michaelsembwever: Implementations might want to return their own version of this.
I'm not sure the best way of achieving this.
Maybe the simplest is to remove the `final`  modifier, that way implementations can do it their way.

Keep in mind this is only a problem if a NoopSpanContext is actually passed in, and the implementation itself largely dictates whether that is possible.

michaelsembwever:
>  Can we downgrade it to 1.7? 

Just checking… is this really that important? (This is a brand new component with a fresh audience, and not required when using the `-api`).

opentracing-noop/src/main/java/io/opentracing/NoopTracer.java
michaelsembwever: The intention to the `create()` method was to make clear that the instance cannot be compared against. 

To avoid the mix-and-match problem you need code to check `span instanceof NoopSpan` and not `NoopSpan.INSTANCE == span`. The create method hopefully pushes the user into accepting this.

yurishkuro:
Unfortunately, yeah, it's still important. There are plenty of systems and companies today still using 1.7. There were even requests for 1.6 recently on Zipkin channels.

I wouldn't call this a new component since the classes are simply moving from -impl, which was 1.7 compatible. For example, our tracer has a dependency on noop being 1.7:

https://github.com/uber/jaeger-client-java/blob/master/jaeger-dropwizard/src/main/java/com/uber/jaeger/dropwizard/Configuration.java#L50

michaelsembwever:
> For example, our tracer has a dependency on noop being 1.7:
>
> https://github.com/uber/jaeger-client-java/blob/master/jaeger-dropwizard/src/main/java/com/uber/jaeger/dropwizard/Configuration.java#L50

ok.

michaelsembwever:
>> For example, our tracer has a dependency on noop being 1.7:
>>
>>    https://github.com/uber/jaeger-client-java/blob/master/jaeger-dropwizard/src/main/java/com/uber/jaeger/dropwizard/Configuration.java#L50
>
> ok.

done.

opentracing-impl/src/test/java/io/opentracing/AbstractTracerTest.java
```diff
@@ -89,6 +89,23 @@ public void testNonEmptyExtract() {
...
+        SpanContext parent = tracer.extract(Format.Builtin.TEXT_MAP, adapter);
+        assert AbstractNoopSpan.INSTANCE != tracer.buildSpan("child").asChildOf(parent).start();
+    }
+
+    @Test
+    public void testExtractOfNoParent() throws Exception {
+        AbstractTracer tracer = new TestTracerImpl();
+        assert AbstractNoopSpan.INSTANCE == tracer.buildSpan("child").asChildOf((Span)AbstractNoopSpan.INSTANCE).start();
+        assert AbstractNoopSpan.INSTANCE == tracer.buildSpan("child").asChildOf((SpanContext)AbstractNoopSpan.INSTANCE).start();
+        assert AbstractNoopSpan.INSTANCE == tracer.buildSpan("child").asChildOf(AbstractNoopSpanBuilder.INSTANCE).start();
```
yurishkuro: I am not sure I follow this. The way TestTextMapExtractorImpl is written it returns a NOOP span context if the carrier does not contain a "trace context" (conceptually). Then here we validate that by using the noop span context as a parent we get back a noop span (it's also the title of the PR). But if we ignore the noop part of it, what it boils down is if my server receives a request without tracing info, it does not start a new trace. I think the most common behavior of the tracers is actually the opposite: continue the trace if there is one, start a new one otherwise.

yurishkuro: I am not sure I follow this. The way TestTextMapExtractorImpl is written it returns a NOOP span context if the carrier does not contain a "trace context" (conceptually). Then here we validate that by using the noop span context as a parent we get back a noop span (it's also the title of the PR). But if we ignore the noop part of it, what it boils down is if my server receives a request without tracing info, it does not start a new trace. I think the most common behavior of the tracers is actually the opposite: continue the trace if there is one, start a new one otherwise.

opentracing-noop/pom.xml
```diff
@@ -0,0 +1,40 @@
...
+        <version>0.16-SNAPSHOT</version>
+    </parent>
+
+    <artifactId>opentracing-noop</artifactId>
+    <name>OpenTracing-noop</name>
+    <description>OpenTracing NoOp</description>
+
+    <properties>
+        <main.basedir>${project.basedir}/..</main.basedir>
+    </properties>
```
yurishkuro: can we add the version here?
```xml
        <main.java.version>1.7</main.java.version>
        <main.signature.artifact>java17</main.signature.artifact>
```

opentracing-noop/src/main/java/io/opentracing/NoopSpan.java
```diff
@@ -27,8 +30,6 @@
         }
     };
 
-    private NoopSpan() {}
-
     @Override
     public SpanContext context() { return CONTEXT; }
```
yurishkuro: why not return NoopSpanContextImpl.INSTANCE ?

opentracing-noop/src/main/java/io/opentracing/NoopTracer.java
```diff
@@ -15,19 +15,25 @@
...
-        return NoopSpanBuilder.INSTANCE;
+        return NoopSpanBuilderImpl.INSTANCE;
     }
 
     @Override
     public <C> void inject(SpanContext spanContext, Format<C> format, C carrier) {}
 
     @Override
-    public <C> SpanBuilder extract(Format<C> format, C carrier) { return NoopSpanBuilder.INSTANCE; }
+    public <C> SpanBuilder extract(Format<C> format, C carrier) { return NoopSpanBuilderImpl.INSTANCE; }
```
yurishkuro: The signature in the Tracer API is to return span context
```
```
What is the use case when `extract()` needs to return a builder?

The 3 most common uses cases for extract that I know of are
  1. extracting from a request that has no tracing context => return some form of "nothing", which can still be used in `childof()`, but will be ignored and a new trace is started
  1. extracting from a request that does have tracing context => return real span context that is used as a parent for the new span
  1. in data pipelines, extracting tracing context from some storage and using it as some Reference type when starting a new span

Neither of these cases requires the return value to be a builder.

Note that there is a special case for Zipkin-like tracers for use case 2 above, where the child of an RPC span is actually a new span with the same SpanID. However, that special case does not require a change in the API, it can be handled internally, [as done in Jaeger tracer](https://github.com/uber/jaeger-client-java/blob/master/jaeger-core/src/main/java/com/uber/jaeger/Tracer.java#L229).

opentracing-noop/pom.xml
michaelsembwever: This is already defined and inherited from the top level pom.xml

screenshot from 2016-10-03 15-46-25

opentracing-noop/src/main/java/io/opentracing/NoopSpan.java
michaelsembwever: good catch. thanks.

opentracing-noop/src/main/java/io/opentracing/NoopTracer.java
michaelsembwever: The NoopTracer has no need to return a SpanBuilder instead of a SpanContext.
Have fixed this. Thanks for pointing it out.

> 1. extracting from a request that has no tracing context => return some form of "nothing", which can still be used in childof(), but will be ignored and a new trace is started

I'm not sure I agree with this logic. Calling extract() doesn't result in a request that isn't being sampled becoming sampled. There would be a huge amplification of tracing if that was the case.

opentracing-impl/src/test/java/io/opentracing/AbstractTracerTest.java
michaelsembwever: > I think the most common behavior of the tracers is actually the opposite: continue the trace if there is one, start a new one otherwise.

This is a decision to be made (but I would like this decision not to interfere with this PR).

Currently there is no mechanism to indicate one way of the other.
In lei of that it makes most sense, and is the safest approach, not to intentionally initialise and start new traces. Otherwise this would cause an massive amplification of trace data.

This goes back to the original code snippet I posted that you would expect to write in library or framework code.

    public Response serverProcessingRequest(Requestr request) {
        TextMapExtractAdapter adapter = new TextMapExtractAdapter(request.getHeaders());
        SpanContext parent = tracer.extract(Format.Builtin.TEXT_MAP, adapter);
        try (Span span = tracer.buildSpan("some-name").asChildOf(parent).start()) {
            // server processes request
            return response;
        }
    }

> I intentionally haven't touched on the forcing tracing where extract returns a noop. That is functionality to add: and given it has not yet been dealt with; it's outside of the scope of this PR.

The instrumentation of libraries and frameworks **must not** dictate (or force) the sampling of traces. This is something that must be inherited either through the trace context (see Zipkin's `SAMPLED` header) or via a sampling mechanism used. In lei of that mechanism the default needs to be safe.

yurishkuro: I agree about sampling, but I wasn't talking about sampling. At some point some node in the architecture is going to receive a request that has no trace ID from upstream. What I'm saying is this node should always start a trace in this case. If it doesn't do that, then a trace may never be started in the whole architecture. The trace may be sampled or not, that's up to the implementation of the tracer, but a new trace ID should be created at that point and propagated downstream. This is even more important from the point of view of distributed context propagation (baggage) than the tracing data itself.

The instrumentation does not decide on sampling, but it does decide on starting the trace by calling `tracer.startTrace()` - I don't see why this call should ever produce a noop span unless the app is configured with noop tracer.

opentracing-noop/src/main/java/io/opentracing/NoopTracer.java
michaelsembwever: i'd forgotten to push :-(
done now.

opentracing-impl/src/test/java/io/opentracing/AbstractTracerTest.java
michaelsembwever: >  What I'm saying is this node should always start a trace in this case.

This is what i've referred to as "forced sampling". Where you look for an existing trace ID, and continue sampling if a trace id is found, or force sampling when one doesn't. The exact terminology used here isn't that important.

The default should be not to force sampling. The reasoning about what provides the minimal, most elegant, code.

A) Taking this default the situation of only tracing where a trace exists looks like 

    public Response serverProcessingRequest(Request request) {
        TextMapExtractAdapter adapter = new TextMapExtractAdapter(request.getHeaders());
        SpanContext parent = tracer.extract(Format.Builtin.TEXT_MAP, adapter);
        try (Span span = tracer.buildSpan("some-name").asChildOf(parent).start()) {
            // server processes request
            return response;
        }
    }

B) The tracing from the top level application, where a existing trace if won't ever exist, looks like

    public Response serverProcessingRequest(Request request) {
        try (Span span = tracer.buildSpan("some-name").asChildOf(parent).start()) {
            // server processes request
            return response;
        }
    }

C) And tracing from a service that must trace regardless if a trace exists or not looks like

    public Response serverProcessingRequest(Requestr request) {
        TextMapExtractAdapter adapter = new TextMapExtractAdapter(request.getHeaders());
        SpanContext parent = tracer.extract(Format.Builtin.TEXT_MAP, adapter);
        SpanBuilder builder = tracer.buildSpan("some-name");
        if (!parent instanceof NoopSpanContext) {
            builder = builder.asChildOf(parent);
        }
        try (Span span = builder.start()) {
            // server processes request
            return response;
        }
    }

To begin with (C) is clumsy. I expect a better approach, and I've referred to this as the forced sampling mechanism before. With the idea of tackling that after this PR let's accept it's clumsiness.

The reason this works can be shown when you look at a large enterprise platform. For example if we take 400 services in an enterprise microservices platform. One of those services is a web application: we presume no external traces exist and (B) is applied. Another of those services is a public API that may be used by the web application might also by other external agents: external traces may exist, but tracing is always initialised, so (C) is applied. The remaining 398 services only trace when a trace already exist, so (A) is applied. On top of this there are 50 libraries re-used in various degrees in the services: they have (A) applied as well.

This is just an example but here you see that even if we can reduce the "forced sampling" mechanism down to two lines of code we have still saved ~1K lines of code.

Another way to look at this is that we, doing it this way, make the code we put into libraries and frameworks (that know nothing about the tracing context and implementation, we have absolutely no relationship with where they are deployed) as non-intrusive as possible.

opentracing-impl/src/main/java/io/opentracing/AbstractNoopSpan.java
```diff
@@ -0,0 +1,42 @@
...
+ * Unless required by applicable law or agreed to in writing, software distributed under the License
+ * is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express
+ * or implied. See the License for the specific language governing permissions and limitations under
+ * the License.
+ */
+package io.opentracing;
+
+final class AbstractNoopSpan extends AbstractSpan implements NoopSpan, NoopSpanContext {
+
+    static final AbstractNoopSpan INSTANCE = new AbstractNoopSpan("noop");
```
yurishkuro: perhaps this class should have a different name, like "generic noop span", 'cause it's not really abstract.

Strictly speaking, I think it's not good that these implementation classes are in the core `io.opentracing` package. If they were in a different package, this could've been called simple NoopSpan.

michaelsembwever: I agree.

Moving them to a different package means moving everything in `-impl` to a new package.

Renaming them, I couldn't come up with anything other then either `AbstractNoopSpan` or `NoopAbstractSpan`.

opentracing-impl/src/test/java/io/opentracing/AbstractTracerTest.java
yurishkuro: I think you can achieve the same result by always writing instrumentation in form (A), where a NoopSpanContext is always passed to childOf(), which shields applications/instrumentation code from the need to even know about the existence of such thing as noopX (unless they really want to know, then they can use model C). Then the question of whether to start or not to start a new trace as a child of Noop context can be a configuration parameter of the tracer, addressed by service deployment, not changes in instrumentation (thus no code changes).

FWIW, at Uber all instrumentation is written to always start a new trace if none present in the inbound request. The reasoning here is that as a service owner I make a decision that I want tracing enabled in my service, regardless of who happened to call me, with or without upstream trace. Doing so allows various analytics of the traffic patterns because traces in a given service are predictable, not at the whim of the caller.

Anyway, agreed that this is not a blocker for this PR.

opentracing-noop/pom.xml
yurishkuro: Cool.

opentracing-noop/src/main/java/io/opentracing/NoopTracer.java
```diff
@@ -15,19 +15,23 @@
 
 import io.opentracing.propagation.Format;
 
-public final class NoopTracer implements Tracer {
+public interface NoopTracer extends Tracer {
+//    static NoopTracer create() {
+//        return NoopTracerImpl.INSTANCE;
+//    }
```
yurishkuro: nit: rm

bensigelman:
@michaelsembwever thank you for doing this

@yurishkuro thank you for reviewing this :)

Ought we mention something about `NoopSpan` and `NoopSpanContext` in the doc strings for `-api` classes that take either of the above? Seems like that's a must if we plan on replacing `null` with `NoopSpanContext` for the various reference-oriented methods...

opentracing-noop/src/main/java/io/opentracing/NoopTracer.java
michaelsembwever: done.

opentracing-impl/src/main/java/io/opentracing/AbstractNoopSpan.java
michaelsembwever: Have renamed package to `no.opentracing.impl`
This allows the noop classes to keep simpler names.

This will cause compatibility headaches, but we're pre-1.0 version.

michaelsembwever:
@bensigelman 
> Ought we mention something about NoopSpan and NoopSpanContext in the doc strings for -api classes that take either of the above?

Yes. I think that's outside of the scope of this PR.

Today the apidocs do not make any mentione of the `-api` classes returning `null`.
Therefore, by java standards, this means that null is never returned under any condition.

Given nothing regarding certain conditions has been documented there is an obvious documentation gap.

I'm a little reluctant though to prematurely tie it to the noop package with this PR. I say wait a little bit, until different java impls have tested this code.

michaelsembwever:
Updated and pushed.
Waiting 24-48 hrs before merging, given no further comments.

bensigelman:
(Sorry for the delay, I have been buried here)

@michaelsembwever, you wrote:
> Today the apidocs do not make any mentione of the -api classes returning null.
>
> Therefore, by java standards, this means that null is never returned under any condition.

In that case I'd argue that we update the API docs to specify that `extract()` may return `null` if no `SpanContext` could be found. This issue need not block this PR, though (which LGTM)... just wanted to be clear about my take on this.

michaelsembwever:
rebased off master.

PR #60, opened 2016-11-08
support opentracing-api module in jdk 1.6

wu-sheng:
The source code of opentracing-api is just interface or abstract class for OpenTracing specification. Is it necessary use jdk 1.7 to build? 
I also check the release version on `The Central Repository`, class file is also build by JDK 1.7

Many projects use opentracing.io specification in order to build trace system, which is using javaagent tech. Such as [sky-walking](https://github.com/wu-sheng/sky-walking)  (I am working on it). It's jdk level must match to target app. Also  we will support jdk 1.6, 1.7 and 1.8 .

Official opentracing-api should release on jdk 1.6. Is this look good to you?

### DIFF
--- a/opentracing-api/pom.xml
+++ b/opentracing-api/pom.xml
@@ -29,6 +29,8 @@
 
     <properties>
         <main.basedir>${project.basedir}/..</main.basedir>
+        <main.java.version>1.6</main.java.version>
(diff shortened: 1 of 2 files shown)


yurishkuro:
I'm ok with 1.6 for the API.

@michaelsembwever - any objections?

PR #61, opened 2016-11-09
Propagate baggage to child spans

mabn:
The specification says that baggage is ["propagated in-band to all child
Spans and their SpanContexts"](http://opentracing.io/documentation/pages/spec.html#baggage).

This PR makes AbstractSpanBuilder copy baggage from a parent Span or SpanContext when creating a child Span.


michaelsembwever:
thanks for catching this @mabn

Issue #55, opened 2016-09-12
Why does Tracer.SpanBuilder extend SpanContext?


yurishkuro:
@michaelsembwever it was added in otherwise non-API change #25, but no documentation why this is necessary.

mabn:
I'm implementing a tracer based on opentracing-impl and this is one of the confusing parts.

Because of this the signature of Extractor is:
```
public interface Extractor<C> {
    Tracer.SpanBuilder extract(C carrier);
}
```
which does not make sense until you remember that SpanBuilder implements SpanContext.

Additionally the intention is to be able to *extract* SpanContext and use it as a parent for new span:
```
SpanContext parent = extractor.extract(carrier);
spanBuilder.childOf(parent)
```
which in effect loses most of the state of the "parent" SpanBuilder implicitly.

There's also a thing with context identifiers (traceId, parentSpanId, spanId) - obviously SpanContext needs to hold identifiers, but SpanBuilder does not. There's `AbstractTracer.createSpanBuilder(String operationName)` method which does not allow to pass any identifiers, which is good, but for example *Extractor* implementations must supply identifiers.

IMO it should be a separate type - AbstractSpanContext and implementations should provide their specific subtype with the identifiers they need. It would make things more clear.

yurishkuro:
@mabn I agree with you. Do you want to try this change in a pull request? It requires some refactoring of the abstract impl.

I am not sure if the AbstractSpanContext is actually needed or not, I guess the only functionality it can provide is the baggage storage, but without a concrete constructor it's difficult to implement something like `SpanContext withBaggage(k, v String)` method, that was used in implementations in other languages in order to keep the context immutable.

bensigelman:
@mabn +1 that this is odd... kind of slipped past me in a PR but it seems odd to me conceptually and I don't understand why it was deemed necessary as part of #25.

michaelsembwever:
> @mabn I agree with you. Do you want to try this change in a pull request? It requires some refactoring of the abstract impl.

If you could at the same time provide the PR against https://github.com/openzipkin/brave-opentracing that would be appreciated.

I think https://github.com/openzipkin/brave-opentracing/pull/5 needs to be re-addressed here as well.

mabn:
I'll give it a shot

michaelsembwever:
> I think openzipkin/brave-opentracing#5 needs to be re-addressed here as well.

[...]

PR #68, opened 2016-11-15
WIP: Separate SpanContext from SpanBuilder

mabn:
This is an attempt to remove SpanContext interface from SpanBuilder. See https://github.com/opentracing/opentracing-java/issues/55

Apparently it has various repercussions...

There are two places where SpanContext should be created:
1. when creating new span - based on parent span context + some span id generation strategy. Good place for that is AbstractSpanBuilder, createSpan() in particular
2. when extracting SpanContext from a carrier. It’s based on baggage and trace state keys.

One option would be to create AbstractSpanContext and then AbstractSpanContextBuilder which would have methods like withTraceState or withBaggageItem, but to keep it simple I think it’s ok to reuse just stick to **immutable AbstractSpanContext** (where mutations create new instances). 
Still, there is a need for some way to instantiate concrete subclass - that's why **I've added createSpanContext(traceState) to AbstractTracer**.

I did few other changes:
 - moved AbstractSpanBuilder.isTraceState to AbstractTracer as it was no longer needed there and there was already isTraceState in the tracer
- removed SpanBuilder.withStateItem - no longer needed

But I've run into problems with Noop implementation.
Right now there are two - one in -impl, the other in -noop. AbstractSpanBuilder asChildOf(Span parent) - requires NoopSpanBuilder to extends AbstractSpanBuilder but on the other hand AbstractSpanBuilder asChildOf tests against NoopSpanContext from opentracing-noop
which cannot depend on AbstractSpanContext (because circular dependency).

I think that it needs to be cleaned up first, especially:
- not returning Noop implementations from AbstractTracer because this forces:
  - either Noop implementations to extend abstract implementation
  - or the abstract implementation to declare interfaces as return types and then downcast
- get rid of one of noop implementations :)

(Note - the this is **not ready to merge**)

### DIFF
--- a/opentracing-impl/src/main/java/io/opentracing/impl/AbstractSpanContext.java
+++ b/opentracing-impl/src/main/java/io/opentracing/impl/AbstractSpanContext.java
@@ -21,7 +21,7 @@ import java.util.Map;
 import static java.util.Collections.emptyMap;
 import static java.util.Collections.unmodifiableMap;
 
-public class AbstractSpanContext implements SpanContext {
+abstract class AbstractSpanContext implements SpanContext {
     private final AbstractTracer tracer;
     protected final Map<String, String> baggage;
     protected final Map<String, Object> traceState;
@@ -41,14 +41,10 @@ public class AbstractSpanContext implements SpanContext {
         return baggage.entrySet();
     }
 
-    SpanContext withBaggage(Map<String, String> baggage) {
-        return new AbstractSpanContext(traceState, baggage, tracer);
-    }
-
     public AbstractSpanContext setBaggageItem(String key, String value) {
         Map<String, String> newBaggage = new HashMap<>(baggage);
         newBaggage.put(key, value);
-        return new AbstractSpanContext(traceState, newBaggage, tracer);
+        return (AbstractSpanContext) tracer.createSpanContext(traceState, baggage);
     }
 
     public String getBaggageItem(String key) {
--- a/opentracing-impl/src/main/java/io/opentracing/impl/AbstractSpanContext.java
+++ b/opentracing-impl/src/main/java/io/opentracing/impl/AbstractSpanContext.java
@@ -44,4 +44,14 @@ public class AbstractSpanContext implements SpanContext {
     SpanContext withBaggage(Map<String, String> baggage) {
         return new AbstractSpanContext(traceState, baggage, tracer);
     }
+
+    public AbstractSpanContext setBaggageItem(String key, String value) {
+        Map<String, String> newBaggage = new HashMap<>(baggage);
+        newBaggage.put(key, value);
+        return new AbstractSpanContext(traceState, newBaggage, tracer);
+    }
+
+    public String getBaggageItem(String key) {
+        return baggage.get(key);
+    }
 }
--- a/opentracing-impl/src/main/java/io/opentracing/impl/AbstractSpanContext.java
+++ b/opentracing-impl/src/main/java/io/opentracing/impl/AbstractSpanContext.java
@@ -1,0 +1,47 @@
+/**
+ * Copyright 2016 The OpenTracing Authors
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License"); you may not use this file except
+ * in compliance with the License. You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software distributed under the License
+ * is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express
+ * or implied. See the License for the specific language governing permissions and limitations under
+ * the License.
+ */
+package io.opentracing.impl;
+
+import io.opentracing.SpanContext;
+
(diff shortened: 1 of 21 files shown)


yurishkuro:
> (Note - the this is not ready to merge)

Tip: prepend "WIP - " to the title, and remove when ready for review/merge

opentracing-impl/src/main/java/io/opentracing/impl/AbstractSpanContext.java
```diff
@@ -0,0 +1,57 @@
...
+    @Override
+    public final Iterable<Map.Entry<String, String>> baggageItems() {
+        return baggage.entrySet();
+    }
+
+    SpanContext withBaggage(Map<String, String> baggage) {
+        return new AbstractSpanContext(traceState, baggage, tracer);
+    }
+
+    public AbstractSpanContext setBaggageItem(String key, String value) {
```
yurishkuro: Just an FYI - in `basictracer` implementations in other languages there is only `withBaggage(key, value)` method, which creates a new immutable ctx.

opentracing-impl/src/main/java/io/opentracing/impl/AbstractTracer.java
```diff
@@ -35,6 +34,8 @@ protected AbstractTracer() {
     }
 
     abstract AbstractSpanBuilder createSpanBuilder(String operationName);
+    
+    abstract AbstractSpanContext createSpanContext(Map<String, Object> traceState);
```
yurishkuro: If I were designing this abstract implementation, I would've done it as:
```java
abstract class AbstractTracer<SPAN extends AbstractSpan, SCTX extends AbstractSpanContext> ... {
    abstract SPAN createSpanContext(Map<String, Object> traceState);
}

The implementations then would deal with their own concrete types instead of abstract superclass, and the abstract tracer would take care of the necessary downcasting (as long as it gets a concrete Class in the constructor)
```

michaelsembwever:
> I did few other changes:
> -   moved AbstractSpanBuilder.isTraceState to AbstractTracer as it was no longer needed there and there was already isTraceState in the tracer
> -    removed SpanBuilder.withStateItem - no longer needed


These are used, ref https://github.com/openzipkin/brave-opentracing/blob/master/src/main/java/io/opentracing/impl/BraveSpanBuilder.java#L102

I think you'll answer a lot of these changes if you can work on the brave-opentracing PR in parallel.

mabn:
I created a PR (https://github.com/openzipkin/brave-opentracing/pull/9) in brave-opentracing, but I think that the direction it's going does't look good enough.
I had to pretty much change everything and still run into issues with Noop implementation (there's like 3 NoopSpanContext classes - one interface and two implementations).

I'm starting to think that the abstract implementation does too much and it would be better if it was limited to some basic methods like the whole family of `SpanBuilder.withTag` or `Span.log` - without methods related to actual tracing (e.g. `AbstractSpanBuilder.createSpan`).

sjoerdtalsma:
>But I've run into problems with Noop implementation.
Right now there are two - one in -impl, the other in -noop. AbstractSpanBuilder asChildOf(Span parent) - requires NoopSpanBuilder to extends AbstractSpanBuilder but on the other hand AbstractSpanBuilder asChildOf tests against NoopSpanContext from opentracing-noop
which cannot depend on AbstractSpanContext (because circular dependency).

Seems solvable to me, e.g.:
- `asChildOf(Span)` for the abstract base-implementation shouldn't assume any typing (i.m.o.) and could just forward to `asChildOf(span.context())`?
- Make the built-in `NoopSpanContext` not rely on the base-implementation (e.g. it is trivial and doesn't need common functionality). Then put the 'smartness' into the abstract base-implementation by inspecting `instanceof NoopSpanContext` if you must.

To me that feels more pure. I'm curious why this would not work?

mabn:
It wouldn't work because:
1. `AbstractSpanBuilder.asChildOf()` returns `AbstractSpanBuilder`.
2. when called with `NoopSpanContext` argument it has to return NoopSpanBuilder
3. which means that `NoopSpanBuilder` has to extend `AbstractSpanBuilder`

2 is based on what @michaelsembwever wrote earlier, but I don't really like this approach - it's a very surprising behaviour. It's one thing to have a noop tracer which does nothing, but in this case receiving request without trace context extracts a `NoopSpanContext` which in turns causes all its children to be `NoopSpans`. Jaeger client for example does not handle it this way.

All in all I wouldn't use the abstract implementation anymore, it introduces more problems than it solves. Little changes in the abstract implementation led to lots of changes e.g. in the mentioned brave-opentracing. IMO it would be more useful if it provided implementation of all simple "non-factory" methods - `setTag`, `log`, etc., but not `createSpan` or `childOf`.

bhs:
@mabn 

> IMO it would be more useful if it provided implementation of all simple "non-factory" methods - setTag, log, etc., but not createSpan or childOf.

+1000 to this. I could not agree more. I also think it "would be nice" to move `Abstract*` out of this repo entirely, as it sends a weird signal that it's actually part of the API when it's just a convenience class for implementations.

bhs:
(I'm going to close this since #115 deals with the underlying problem and this has fallen out of date)

PR #72, opened 2016-12-13
MockSpan make finish operation idempotent

pavolloffay:
@bensigelman could you please review?

### DIFF
--- a/opentracing-mock/src/main/java/io/opentracing/mock/MockSpan.java
+++ b/opentracing-mock/src/main/java/io/opentracing/mock/MockSpan.java
@@ -33,6 +33,7 @@ public final class MockSpan implements Span {
     private MockContext context;
     private final long parentId; // 0 if there's no parent.
     private final long startMicros;
+    private boolean finished;
     private long finishMicros;
     private final Map<String, Object> tags;
     private final List<LogEntry> logEntries = new ArrayList<>();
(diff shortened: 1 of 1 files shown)


opentracing-mock/src/main/java/io/opentracing/mock/MockSpan.java
```diff
@@ -94,8 +95,11 @@ public void finish() {
 
     @Override
     public synchronized void finish(long finishMicros) {
-        this.finishMicros = finishMicros;
-        this.mockTracer.appendFinishedSpan(this);
+        if (!finished) {
```
bensigelman: since MockSpan is mainly used for testing, AND since it's actually undefined behavior to call `finish` multiple times on a single Span instance, I'd argue that this should at least log something in the `finished==true` case. thoughts?

yurishkuro: I agree. In fact, based on "mainly used for testing" argument, it's not unreasonable to even throw an exception (or at least make it a configuration option). It is usually a sign of poor instrumentation, not a problem that needs fixing in the tracer.

pavolloffay: I'm doing this change because span was added multiple times to `Tracer#finishedSpans`

> I'd argue that this should at least log something in the finished==true case.

Agree, this can help find issues in instrumentation. Log statement has been added. 

Actually why is not `Span.finish()` idempotent by definition? Is there any reason for it? If multiple calls on `finish()` are not defined then behavior from tracer to tracer might by different, then it could be more difficult to change a tracing system.

bensigelman: @pavolloffay "not defined" is just another way of saying "don't do it." To answer your question: in practice, high-performance Tracer impls can use free pools, and `finish()` is the cue to reuse the guts of the Span implementation.

pavolloffay: I agree with you. PR updated to throw exception on multiple `finish` calls.

pavolloffay: @bensigelman we probably forgot this one.

bensigelman:
Sorry to drop! Merging.

PR #73, opened 2016-12-13
Possible ClassCastException fix in AbstractSpanBuilder + API fixes.

sjoerdtalsma:
- AbstractSpanBuilder did unnecessary (and unsafe!) cast to AbstractSpan that was not necessary.
- Span implements AutoCloseable which is only available since Java7 while project delcares Java6.
   Changed to Closeable which also extends AutoCloseable and doesn't declare additional methods
   so is 100% compatible and Java6 compliant.
 - Fixed broken @see link in Reference class javadoc.

### DIFF
--- a/opentracing-api/src/main/java/io/opentracing/References.java
+++ b/opentracing-api/src/main/java/io/opentracing/References.java
@@ -18,7 +18,7 @@ package io.opentracing;
  *
  * References are used by Tracer.buildSpan() to describe the relationships between Spans.
  *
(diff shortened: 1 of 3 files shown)


sjoerdtalsma:
@michaelsembwever I think these are pretty straightforward fixes, could you please take a look?

adriancole:
2p I'd recommend adding animal sniffer to this so that these things break
the build.

michaelsembwever:
Thanks @sjoerdtalsma

Issue #74, opened 2016-12-14
MockTracer custom recorders


pavolloffay:
Hello,

I would like to contribute a custom recorder for `MockTracer`. In the most cases recording spans to the list is sufficient, however, it would be good to be able to send spans to an arbitrary recorder (e.g. console for demo purposes).

```
interface Recorder {
    void record(Span span);
}

....
public MockTracer() {
 this(Propagator.PRINTER, Recorder.LIST)
}

public MockTracer(Propagator propagator, Recorder recorder) {
}
```

@adriancole @bensigelman @yurishkuro what do you think?

bensigelman:
@pavolloffay I don't have any great objections... I can imagine creating something a little more powerful/flexible like https://github.com/opentracing/basictracer-go for Java and building off of that, too.

My only minor concern is that you're describing something other than a true "mock" tracer... so the name is drifting a bit. But I can live with that.

yurishkuro:
One concern I have is that currently finished spans are accessible via the MockTracer itself. If you introduce a Recorder concept, that becomes weird (leak in encapsulation).

Maybe a simpler solution is just to add `onSpanFinished(MockSpan span)` no-op method and override it where needed?

pavolloffay:
@yurishkuro `onSpanFinished(MockSpan span)` would work for me.

Thanks

pavolloffay:
Closing as https://github.com/opentracing/opentracing-java/issues/76 is merged.

PR #76, opened 2016-12-15
MockTracer onSpanFinished

pavolloffay:
@bensigelman @yurishkuro could you please review? 

https://github.com/opentracing/opentracing-java/issues/74

### DIFF
--- a/opentracing-mock/src/main/java/io/opentracing/mock/MockTracer.java
+++ b/opentracing-mock/src/main/java/io/opentracing/mock/MockTracer.java
@@ -65,6 +65,12 @@ public final class MockTracer implements Tracer {
     }
 
     /**
+     * Noop method called on {@link Span#finish()}.
+     */
+    protected void onSpanFinished(MockSpan mockSpan) {
+    }
+
+    /**
      * Propagator allows the developer to intercept and verify any calls to inject() and/or extract().
      *
      * By default, MockTracer uses Propagator.PRINTER which simply logs such calls to System.out.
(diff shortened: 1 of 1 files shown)


opentracing-mock/src/main/java/io/opentracing/mock/MockTracer.java
```diff
@@ -65,6 +65,12 @@ public synchronized void reset() {
     }
 
     /**
+     * Noop method called on {@link Span#finish()}.
+     */
+    public void onSpanFinished(MockSpan mockSpan) {
```
yurishkuro: I think you want `protected`

opentracing-mock/src/main/java/io/opentracing/mock/MockSpan.java
```diff
@@ -96,6 +96,7 @@ public void finish() {
     public synchronized void finish(long finishMicros) {
         this.finishMicros = finishMicros;
         this.mockTracer.appendFinishedSpan(this);
+        this.mockTracer.onSpanFinished(this);
```
yurishkuro: I think the tracer itself should call this from `appendFinishedSpan`

pavolloffay:
PR updated.

pavolloffay:
When do you plan to do next release? I will need it for `contrib/java-jaxrs`

bensigelman:
@pavolloffay I'll try(*) to cut a release right now.

(*): we are using this clever TravisCI release process... the last time we tried to use it "something broke", though, so let's cross our collective fingers.

pavolloffay:
@bensigelman @yurishkuro thanks!

pavolloffay:
@bensigelman We missed  a quite important thing https://github.com/opentracing/opentracing-java/blob/master/opentracing-mock/src/main/java/io/opentracing/mock/MockTracer.java#L32 

I will create a PR to fix it.

PR #80, opened 2017-01-03
MockSpan fix startMicros to implicit value

pavolloffay:
@adriancole @bensigelman could you please review?

### DIFF
--- a/opentracing-mock/src/main/java/io/opentracing/mock/MockSpan.java
+++ b/opentracing-mock/src/main/java/io/opentracing/mock/MockSpan.java
@@ -1,5 +1,5 @@
 /**
- * Copyright 2016 The OpenTracing Authors
+ * Copyright 2016-2017 The OpenTracing Authors
  *
  * Licensed under the Apache License, Version 2.0 (the "License"); you may not use this file except
  * in compliance with the License. You may obtain a copy of the License at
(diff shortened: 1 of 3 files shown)


pavolloffay:
@yurishkuro could you please review?

opentracing-mock/src/test/java/io/opentracing/mock/MockTracerTest.java
```diff
@@ -99,4 +101,19 @@ public void testChildSpan() {
...
         assertEquals(parent.context().traceId(), child.context().traceId());
     }
+
+    @Test
+    public void testStartTimestamp() {
+        MockTracer tracer = new MockTracer();
+        final long startMicros = System.currentTimeMillis() * 1000;
+        {
+            tracer.buildSpan("foo").start().finish();
+        }
```
yurishkuro: why extra scope?

pavolloffay: just minimizing sope. It's used in other tests in this file.

pavolloffay:
@yurishkuro PR updated. It assumes that builder should not be used reused. Start time is set only once, or if `withStartTimestamp(time)` is called.

opentracing-mock/src/main/java/io/opentracing/mock/MockSpan.java
```diff
@@ -1,5 +1,5 @@
 /**
- * Copyright 2016 The OpenTracing Authors
+ * Copyright 2016-2017 The OpenTracing Authors
```
pavolloffay: Making license plugin happy.

(PR editing `MockSpan` was created in 2016 and merged in 2017).

bensigelman:
Thanks and sorry again (again) for missing this the first time around.

PR #85, opened 2017-01-09
Noop implementation should be in java version 1.6

ivansenic:
As the API is explicitly defined as java 1.6, there is no reason why noop implementation should also not be build in v1.6, especially that there is nothing out there that would represent a problem for such thing. Thus I simply updated the ```pom.xml``` to overwrite the default setting from the main pom.


wu-sheng:
Thank you for bringing this up.

PR #86, opened 2017-01-09
Switches to release commit user instead of amplifying my commits

adriancole:
I get a large amount of people asking me about this repo. Because my user
is used in the release commits, it looks like I am doing releases, and stats
skew as if I'm doing a lot of the code here. This switches to anonymize
things like that, and in doing so stops the amplification. Eventually, things
will balance out.


adriancole:
cc @pavolloffay I think this approach is better for many reasons (including the same rationale author tags aren't awesome)

PR #87, opened 2017-01-11
Added Format.Builtin.toString() as they tend to show up in exception.

sjoerdtalsma:
For example, using Brave with Jersey, the following exception message showed up:
```
org.glassfish.jersey.server.ContainerException: java.lang.AssertionError:
no registered extractor for io.opentracing.propagation.Format$Builtin@427bb79a
```


opentracing-api/src/main/java/io/opentracing/propagation/Format.java
```diff
@@ -71,5 +71,16 @@
          * @see Format
          */
         public final static Format<ByteBuffer> BINARY = new Builtin<ByteBuffer>();
+
+        /**
+         * @return Short name for built-in formats as they tend to show up in exception messages.
+         */
+        @Override
+        public String toString() {
```
yurishkuro: I think it would be cleaner to define a private constructor that accepts a `String name` parameter and toString simply returns the `name`. It would also fix the leak in this class - its called "Buildin", but people can create more instances of it, which wasn't the intention.

sjoerdtalsma: Agree. I wasn't sure whether the public constructor was intentional.
In this case, I would propose changing Builtin to an enum, which automatically also fixes the toString() issue!

sjoerdtalsma: My bad, enum is not possible due to the Format generic.
Made the constructor private instead.

bensigelman:
LGTMToo. Thanks @sjoerdtalsma and @yurishkuro .

opentracing-api/src/main/java/io/opentracing/propagation/Format.java
```diff
@@ -70,6 +76,14 @@
...
          */
-        public final static Format<ByteBuffer> BINARY = new Builtin<ByteBuffer>();
+        public final static Format<ByteBuffer> BINARY = new Builtin<ByteBuffer>("BINARY");
+
+        /**
+         * @return Short name for built-in formats as they tend to show up in exception messages.
+         */
+        @Override
+        public String toString() {
+            return "Builtin." + name;
```
pavolloffay: btw. you can do:  `return Builtin.class.getSimpleName() + "." + name;`

bensigelman: @sjoerdtalsma do you want to make this change or should we merge this PR? thanks.

sjoerdtalsma: @bensigelman Thanks for the mention; I'll modify the PR; this change is trivial, looks like I missed @pavolloffay 's comment..

sjoerdtalsma:
Last activity on this PR is quite a while ago and there have been releases of `opentracing-java` since.
Any reason this didn't get merged?

wu-sheng:
@sjoerdtalsma merged, seem good enough. Thanks!!

PR #91, opened 2017-01-16
MockTracer TextMap propagator

pavolloffay:
It would be good to provide default `TextMap` propagator for `MockTracer`. I'm rewriting this code in all instrumentation tests.  

@yurishkuro @bensigelman could you please review?


bensigelman:
As with #92, this seems harmless and has already been reviewed/approved. Merging.

PR #92, opened 2017-01-16
Make MockSpanBuilder public so it can be used in different packages

pavolloffay:
`MockTracer.SpanBuilder` has package level visibility. It cannot be used in different packages (`io.opentracing.mock`) . It makes it hard to test instrumentation libraries which are usually in different packages.

Q:
`MockContext` should be probably also public. It is needed when implementing custom `MockTracer.Propagator` (which a paramter in `MockTracer` public constructor).

@yurishkuro @bensigelman could you please review?

### DIFF
--- a/opentracing-mock/src/main/java/io/opentracing/mock/MockTracer.java
+++ b/opentracing-mock/src/main/java/io/opentracing/mock/MockTracer.java
@@ -100,7 +100,7 @@ public class MockTracer implements Tracer {
     }
 
     @Override
-    public SpanBuilder buildSpan(String operationName) {
+    public Tracer.SpanBuilder buildSpan(String operationName) {
         return new SpanBuilder(operationName);
     }
 
--- a/opentracing-mock/src/main/java/io/opentracing/mock/MockSpan.java
+++ b/opentracing-mock/src/main/java/io/opentracing/mock/MockSpan.java
@@ -192,7 +192,7 @@ public final class MockSpan implements Span {
          *
          * @see MockContext#withBaggageItem(String, String)
          */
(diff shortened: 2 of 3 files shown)


yurishkuro:
What is the use case where you need to use the concrete type `MockTracer.SpanBuilder`?

I agree that MockContext should be public.

pavolloffay:
PR updated.

> What is the use case where you need to use the concrete type MockTracer.SpanBuilder?

I changed the code, `MockTracer#buildSpan()` should return `Tracer.SpanBuilder` and not concrete impl, which has package visibility.

> I agree that MockContext should be public.

I have changed `MockContext` constructor to be public. This enables implementing custom `MockTracer.Propagator` in different packages.

opentracing-mock/src/main/java/io/opentracing/mock/MockTracer.java
```diff
@@ -100,7 +100,7 @@ protected void onSpanFinished(MockSpan mockSpan) {
     }
 
     @Override
-    public SpanBuilder buildSpan(String operationName) {
+    public Tracer.SpanBuilder buildSpan(String operationName) {
```
yurishkuro: hm, strictly speaking this _reduces_ the visibility since you're returning an interface even when you invoke the method directly via mock tracer instance. Was that your intention?

pavolloffay: `mockTracer.buildSpan("parent").start()` fails in packages different what `io.opentracing.mock` because mock `SpanBuilder` has package visibility.

There are two options: 
* change visibility of mock `SpanBuilder` to public
* `MockTracer#buildSpan` return `Tracer.SpanBuilder`

PR #93, opened 2017-01-19
[impl] Improve the performance of the AbstractSpanBuilder and Abstrac…

manolama:
…tSpan

by lazily initializing maps and lists to avoid eating heap space when it
won't be needed.
Also add tons of unit tests for the AbstractSpanBuilder and AbstractSpan.
Some validation should occur here if these are meant to be overridden by
implementations.


wu-sheng:
Hi, I review your code, and see your point about `lazily initializing maps and lists`.

And here is a scenario, if a span object is hold by two thread, and `setTag` concurrently, your `lazily initializing mechanism` have a chance losing some `tag`s, because of map/list init-override. Although, a span object hold by two thread is not recommended, but not forbidden.

bhs:
This is going to be obviated by #115, btw. I'm going to close since this is a dormant PR. If that's a problem please let me know and we'll figure something out.

PR #94, opened 2017-02-09
trivial, comments: for travis encrypt to use a specific repository

pavolloffay:
The repository when issuing `travis encrypt` is automatically set to origin
from git settings, however it can cause problems if origin is set to a
fork..


pavolloffay:
@bensigelman could you please review? It's related to https://github.com/opentracing-contrib/meta/pull/7 This should prevent any potential problems with `travis encrypt`.

bensigelman:
LG, thanks!

PR #96, opened 2017-02-09
Allow consumers to define tags

pavolloffay:
Allow users to define framework/app... specific tags.

### DIFF
--- a/opentracing-api/src/main/java/io/opentracing/tag/Tags.java
+++ b/opentracing-api/src/main/java/io/opentracing/tag/Tags.java
@@ -1,5 +1,5 @@
 /**
- * Copyright 2016 The OpenTracing Authors
+ * Copyright 2016-2017 The OpenTracing Authors
  *
  * Licensed under the Apache License, Version 2.0 (the "License"); you may not use this file except
  * in compliance with the License. You may obtain a copy of the License at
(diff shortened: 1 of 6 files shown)


pavolloffay:
@bensigelman @yurishkuro @michaelsembwever could you please review?

opentracing-api/src/main/java/io/opentracing/tag/Tags.java
```diff
@@ -18,7 +18,8 @@
  * semantic information about the spans. Tracers may expose additional features based on these
  * standardized data points. Tag names follow a general structure of namespacing.
  *
- * @see http://opentracing.io/data-semantics/
+ * @see <a href="http://opentracing.io/documentation/pages/spec.html">http://opentracing.io/documentation/pages/spec
```
pavolloffay: This was wrong. I'm not sure which link should be used. This one or:

https://github.com/opentracing/specification/blob/master/data_conventions.yaml

opentracing-api/src/main/java/io/opentracing/tag/Tags.java
```diff
@@ -18,7 +18,7 @@
  * semantic information about the spans. Tracers may expose additional features based on these
  * standardized data points. Tag names follow a general structure of namespacing.
  *
- * @see http://opentracing.io/data-semantics/
+ * @see <a href="http://opentracing.io/documentation/pages/spec.html">http://opentracing.io/documentation/pages/spec.html</a>
```
pavolloffay: This was wrong. I'm not sure which link should be used. This one or:

https://github.com/opentracing/specification/blob/master/data_conventions.yaml

wu-sheng: I think `https://github.com/opentracing/specification/blob/master/data_conventions.yaml` is better.

bensigelman: I agree ^^^

PR #98, opened 2017-02-13
MockSpan throw an exception when setting data on finished span

pavolloffay:
call to any method after `finish` is undefined, therefore for testing purposes I find it useful to throw an exception if such situation happens.

### DIFF
--- a/opentracing-mock/src/main/java/io/opentracing/mock/MockSpan.java
+++ b/opentracing-mock/src/main/java/io/opentracing/mock/MockSpan.java
@@ -39,12 +39,15 @@ public final class MockSpan implements Span {
     private final List<LogEntry> logEntries = new ArrayList<>();
     private String operationName;
 
+    private final List<RuntimeException> errors = new ArrayList<>();
+
     public String operationName() {
         return this.operationName;
     }
 
     @Override
     public Span setOperationName(String operationName) {
+        finishedCheck("Setting operationName {%s} on already finished span", operationName);
         this.operationName = operationName;
         return this;
     }
@@ -83,6 +86,13 @@ public final class MockSpan implements Span {
         return new ArrayList<>(this.logEntries);
     }
 
+    /**
+     * @return a copy of exceptions thrown by this class (e.g. adding a tag after span is finished).
+     */
+    public List<RuntimeException> generatedErrors() {
+        return new ArrayList<>(errors);
+    }
+
     @Override
     public synchronized MockContext context() {
         return this.context;
@@ -95,13 +105,10 @@ public final class MockSpan implements Span {
 
     @Override
     public synchronized void finish(long finishMicros) {
-        if (!finished) {
-            this.finishMicros = finishMicros;
-            this.mockTracer.appendFinishedSpan(this);
-            this.finished = true;
-        } else {
-            throw new IllegalStateException("Span.finish() should be called only once!");
-        }
+        finishedCheck("Finishing already finished span");
+        this.finishMicros = finishMicros;
+        this.mockTracer.appendFinishedSpan(this);
+        this.finished = true;
     }
 
     @Override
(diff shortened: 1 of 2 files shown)


pavolloffay:
@bensigelman @yurishkuro @wu-sheng could you please review?

Once this is merged could you please do a micro release?

pavolloffay:
One thing: in some cases it would not fail the test. The exception would be just logged.. 

How to handle this and propagate exceptions/errors to tests? 

* remove span from `finishedSpans`? 
* throw this exception when accessing `finishedSpans`? 
* store an exception in span/tracer and make it available?
...

I would prefer if the exception/s would be available on a span object. 

e.g. `Assert.assertNull(mockSpan.error())` 
 
Potentially there could be more errors/exceptions (2 times added a tag after finish) per span, but I think keeping only one is sufficient.

pavolloffay:
PR updated. Span holds all encountered errors.

opentracing-mock/src/main/java/io/opentracing/mock/MockSpan.java
```diff
@@ -83,6 +86,13 @@ public long finishMicros() {
         return new ArrayList<>(this.logEntries);
     }
 
+    /**
+     * @return a copy of errors.
+     */
+    public List<RuntimeException> errors() {
```
bensigelman: how about `generatedErrors()`? I want to somehow clarify that these are not errors in the `setTag("error", true)` sense.

pavolloffay: Agree name changed to `generatedErrors` and javadoc updated  :octocat: 

I was also thinking about `runtimeExceptions()`.. (naming dilemma..)

bensigelman:
Well, this LGTM.

Let's give others 24h to take issue with this. If there are no concerns we can release a patch version after merging. Thanks again.

opentracing-mock/src/main/java/io/opentracing/mock/MockSpan.java
```diff
@@ -110,29 +117,35 @@ public void close() {
...
 
     @Override
     public final Span log(Map<String, ?> fields) {
         return log(nowMicros(), fields);
     }
+
     @Override
-    public final Span log(long timestampMicros, Map<String, ?> fields) {
+    public final synchronized Span log(long timestampMicros, Map<String, ?> fields) {
+        finishedCheck(String.format("Adding logs %s at %d to already finished span", fields.toString(),
```
yurishkuro: small nit: while the Mock span is not performance-critical, it would be more efficient (and more DRY) to just pass a variadic list of args to finishedCheck, and only invoke String.format() within that function if the condition is triggered.

opentracing-mock/src/main/java/io/opentracing/mock/MockSpan.java
```diff
@@ -262,4 +276,12 @@ static long nextId() {
     static long nowMicros() {
         return System.currentTimeMillis() * 1000;
     }
+
+    private synchronized void finishedCheck(String message) {
+        if (finished) {
+            RuntimeException ex = new IllegalStateException(message);
+            errors.add(ex);
```
yurishkuro: not sure I completely follow why it is necessary to record the exception if it's being thrown anyway - the caller already needs to deal with it, so it doesn't need a 2nd way to be told that exception happened.

but I don't object to merging it if it's useful

wu-sheng: First, I do not object to it, either. I am just curious about why need the `errors` list? It's like you want to check the list rather than count the `catch exception`?

pavolloffay: In some cases you cannot rely on the fact that a thrown exception by `MockSpan` would fail/error a test. E.g. in some cases an exception would be just logged by a container and everything in tests would look normal. Therefore I would prefer to have an explicit way to check if something non-compliant with spec happened.

pavolloffay:
PR updated, `String.format()` has been moved to the exception constructor.

opentracing-mock/src/main/java/io/opentracing/mock/MockSpan.java
wu-sheng: I see. Thx.

Issue #95, opened 2017-02-09
Add a new Binary Format to allow for variable SpanContext size


gkumar7:
A ```Binary``` format is defined in opentracing-api which uses a ```ByteBuffer``` as a carrier. ```ByteBuffer``` needs to be allocated with a predetermined size for the ```SpanContext``` to be injected. This in turn requires a size calculation of baggage. 

Possible approaches:
1. Place an upper limit on the size of the baggage.
2. Define a new carrier class (similar to ```TextMap```) and format (similar to ```BINARY```). Essentially, this would provide a wrapper for ```byte[]```.

```java
// Additional Format specification (Format.java)
public final static Format<SimpleBinary> SIMPLE_BINARY = new Builtin<SimpleBinary>();
```
```java
public class SimpleBinary {
    private byte[] payload;

    public byte[] getPayload() {
        return payload;
    }

    public void setPayload(byte[] payload) {
        this.payload = payload;
    }
}
```
BinaryInjectorImpl and BinaryExtractorImpl would be library implementation-specific (below is an example)
```java
final class BinaryInjectorImpl implements Injector<SimpleBinary> {

    private final AbstractTracer tracer;
    private boolean baggageEnabled = AbstractTracer.BAGGAGE_ENABLED;

    BinaryInjectorImpl(AbstractTracer tracer) {
        this.tracer = tracer;
    }

    @Override
    public void inject(SpanContext spanContext, SimpleBinary carrier) {
        byte[] contextBytes = convertSpanContextToBytes(spanContext);
        carrier.setPayload(contextBytes);
    }

    void setBaggageEnabled(boolean baggageEnabled) {
        this.baggageEnabled = baggageEnabled;
    }
}
```

This allows the library to not have to calculate the size of the ```SpanContext``` as this size will depend on how the context is converted to bytes in ```convertSpanContextToBytes``` (ie using avro, etc).

yurishkuro:
personally, my preference is for streaming-friendly carriers, such as the standard io.Writer we used in Go:
```go
type Writer interface {
	Write(p []byte) (n int, err error)
}
```

The key difference here is that the `Write` method can be called by the tracer many times. This allows the tracer to be implemented without having to pre-allocate `[]byte` for the whole payload.

gkumar7:
Yes, I also prefer streaming-friendly buffers. The way the Format is specified currently:
```java
public final static Format<ByteBuffer> BINARY = new Builtin<ByteBuffer>();
```
[...]

PR #99, opened 2017-02-16
Variable size ByteBuffer implementation

gkumar7:
This patch includes an implementation of a variable-size ```BinaryHolder``` to abstract away size calculation for a ```ByteBuffer```. The ```BinaryHolder``` is flexible enough to be used as a replacement for the ```Binary``` format. For more information, see  #95.

@bensigelman, @yurishkuro, @wu-sheng, @mabn, @michaelsembwever  could you please review?

### DIFF
--- a/opentracing-api/src/main/java/io/opentracing/propagation/BinaryHolder.java
+++ b/opentracing-api/src/main/java/io/opentracing/propagation/BinaryHolder.java
@@ -25,7 +25,7 @@ import java.nio.ByteBuffer;
  * @see io.opentracing.Tracer#extract(Format, Object)
  */
 public final class BinaryHolder {
-    private ByteBuffer carrier;
+    private ByteBuffer payload;
 
     /**
      * Creates a BinaryHolder with no initial size allocation.
@@ -39,25 +39,25 @@ public final class BinaryHolder {
      * Creates a BinaryHolder with provided ByteBuffer.
      * Useful for ensuring wrapped ByteBuffer has particular properties prior to Tracer.inject().
      *
-     * @param carrier The ByteBuffer to wrap within the BinaryHolder.
+     * @param payload The ByteBuffer to wrap within the BinaryHolder.
      *
      * @see Format.Builtin#BINARY_HOLDER
      */
-    public BinaryHolder(ByteBuffer carrier) {
-        this.carrier = carrier;
+    public BinaryHolder(ByteBuffer payload) {
+        this.payload = payload;
     }
 
     /**
      * If no initial ByteBuffer was allocated for the BinaryHolder, the payload becomes the wrapped ByteBuffer.
-     * Otherwise, the payload is placed in existent ByteBuffer.
+     * Otherwise, the payload is placed in existing ByteBuffer.
      *
      * @param payload a ByteBuffer.
      */
     public void addPayload(ByteBuffer payload) {
-        if (carrier == null) {
-            carrier = payload;
+        if (this.payload == null) {
+            this.payload = payload;
         } else {
-            carrier.put(payload);
+            this.payload.put(payload);
         }
     }
 
@@ -71,7 +71,7 @@ public final class BinaryHolder {
      * @see ByteBuffer#limit()
      * @see ByteBuffer#capacity()
      */
-    public ByteBuffer getCarrier() {
-        return carrier;
+    public ByteBuffer getPayload() {
+        return payload;
     }
 }
--- a/opentracing-api/src/main/java/io/opentracing/propagation/BinaryHolder.java
+++ b/opentracing-api/src/main/java/io/opentracing/propagation/BinaryHolder.java
@@ -1,0 +1,77 @@
+/**
+ * Copyright 2016-2017 The OpenTracing Authors
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License"); you may not use this file except
+ * in compliance with the License. You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software distributed under the License
+ * is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express
+ * or implied. See the License for the specific language governing permissions and limitations under
+ * the License.
+ */
+package io.opentracing.propagation;
+
+import io.opentracing.SpanContext;
+
+import java.nio.ByteBuffer;
+
+/**
+ * BinaryHolder is a built-in carrier for Tracer.inject() and Tracer.extract(). BinaryHolder allows inject to convert
+ * the SpanContext to a byte encoding without a need to know the ByteBuffer size to allocate beforehand.
+ *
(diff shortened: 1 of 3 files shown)


bhs:
@gkumar7 sorry for the insane delay, I am looking at pending issues/PRs briefly. This branch has a merge conflict right now which is problematic... can you address?

gkumar7:
@bhs Thanks for the response! Seems like there were a few conflicting changes while this review was out, fixed now.

opentracing-api/src/main/java/io/opentracing/propagation/Format.java
```diff
@@ -79,6 +79,16 @@ private Builtin(String name) {
         public final static Format<ByteBuffer> BINARY = new Builtin<ByteBuffer>("BINARY");
 
         /**
+         * The BINARY_HOLDER format allows for variable binary encoding of SpanContext state for Tracer.inject and
```
bhs: nit: "... variable-length binary ..."

opentracing-api/src/main/java/io/opentracing/propagation/BinaryHolder.java
```diff
@@ -0,0 +1,77 @@
...
+     *
+     * @see Format.Builtin#BINARY_HOLDER
+     */
+    public BinaryHolder(ByteBuffer carrier) {
+        this.carrier = carrier;
+    }
+
+    /**
+     * If no initial ByteBuffer was allocated for the BinaryHolder, the payload becomes the wrapped ByteBuffer.
+     * Otherwise, the payload is placed in existent ByteBuffer.
```
bhs: nit: "existing"

opentracing-api/src/main/java/io/opentracing/propagation/BinaryHolder.java
```diff
@@ -0,0 +1,77 @@
...
+     *
+     * @param payload a ByteBuffer.
+     */
+    public void addPayload(ByteBuffer payload) {
```
bhs: why `addPayload` and `getCarrier`? shouldn't the names be symmetric?

yurishkuro:
@bhs I don't have strong feeling about it. I never particularly liked the ByteBuffer-based API, I would rather use java.io streams.

opentracing-api/src/main/java/io/opentracing/propagation/BinaryHolder.java
gkumar7: Good point, fixed.

gkumar7:
java.io.streams would be an interesting approach, although a bit tricky as there are both input and output streams. I do not think one Format specification will be able to handle both cases

yurishkuro:
> bit tricky as there are both input and output streams. I do not think one Format specification will be able to handle both cases

@gkumar7 that's why my original proposal for Format was not `Format<C>` but `Format<R, W>`, but it was shot down as "confusing" ¯\_(ツ)_/¯

gkumar7:
Ah, that makes sense. @bhs could we merge these changes for now so people have a workaround to the ```ByteBuffer``` format and then we can discuss more about the Format specification in a separate issue?

yurishkuro:
> could we merge these changes for now 

@gkumar7 I'd want to be careful, because merging this has implications, it becomes a requirement for all existing tracers to support this new carrier.

gkumar7:
Yes, very true, but I would argue that tracers should support a ```BinaryHolder``` to provide users of the library an alternative to the fundamental limitation of a ```ByteBuffer```. 

The ```ByteBuffer``` format is there, but many java tracing libraries do not formally provide an example usage. Anyone using the ```ByteBuffer``` will hit a roadblock when having to specify the size of the ```ByteBuffer``` before serialization occurs.

bhs:
@gkumar7 FYI, I have been thinking about this some more, apologies for the insanely high latency.

I would like to merge this change, but not until we're close to cutting a new major release for OT-Java... maybe we can get rid of the other binary format altogether at that point (since we're introducing breaking changes anyway).

gkumar7:
@bhs Sounds good, I will keep this pr open until then. Thanks!

tedsuo:
@gkumar7 @yurishkuro I propose this change should get rolled into `0.31`. We'll need to update this PR to point at the `opentracing:v0.31.0` branch.

I'm wondering if anyone supports the current `BINARY` carrier type in java, and whether we can change that type rather than add `BINARY_HOLDER`.

gkumar7:
+1 to update ```BINARY``` type, our team has not been using it.

malafeev:
+1 
I need it to implement cassandra server side tracing. Payload is sent using ByteBuffer there.

opentracing-impl/src/test/java/io/opentracing/impl/TestBinaryHolder.java
```diff
@@ -0,0 +1,71 @@
...
+
+        for (int i = 0; i < 2; i++) {
+            binaryHolder.addPayload(randomBytes51);
+            randomBytes51.rewind();
+        }
+    }
+
+    @Test
+    public void testBinaryHolderWithSufficientCapacity() {
+        System.out.println("binary holder with sufficient capacity");
```
carlosalberto: When/if we get this PR merged, we should get rid of these `println` lines ;)

carlosalberto:
So my impression, after going through the comments and #95, is that people need some kind of binary format *now*. So what about:

* Removing/deprecating the current BINARY format.
* Adding this `BinaryHolder` format?
* Later, as/when needed, add a `BINARY_STREAM` format?

That way, we could have people use their code using the required binary format, and later on offer to a stream-based binary format.

Opinions?

yurishkuro:
@carlosalberto I'm not seeing the benefits of such approach, it just creates an API churn. The holder class is already new/breaking API, why not  just define an equivalent of Go's io.Reader/io.Writer now and make it the only BINARY format? Those are very small interfaces. The holder class can be an adapter for them.

tedsuo:
I agree, let's go with a streaming BINARY format. Since this will be a breaking change, let's get it in on v31.

carlosalberto:
@gkumar7 @malafeev Hey, would you mind jumping into the discussion/review of #223 ? You may be interested as that's an effort to support BINARY through a stream-friendly interface (the PR itself inspired in this one ;) )

malafeev:
@carlosalberto I added comment to #223 .
Without seeing how it would work in MockTracer inject/extract it's hard to say.

gkumar7:
@carlosalberto, thanks for creating a new pull request (although it would have been nice to make changes to the existing one). Regardless, I would agree with @malafeev that some example usage would be helpful.

opentracing-impl/src/test/java/io/opentracing/impl/TestBinaryHolder.java
gkumar7: As you might have noticed this pull request was hanging around for about 9 months. The ```opentracing-java``` code has changed quite a bit since then. Initially, the structure of the test code in ```opentracing-impl``` (this module has been removed) had test classes with a ```println``` at the beginning of each method. Even I was not sure of the the reason for this, but included similar lines in this pull request to remain consistent with the rest of the codebase.

gkumar7:
Closing as #223 is merged.

PR #101, opened 2017-02-17
MockSpanBuilder return its type not Tracer.SpanBuilder

pavolloffay:
Currently in tests when one is building a span he has to cast it be able to access `MockSpan` attributes.

```
MockSpan parent = (MockSpan) mockTracer.buildSpan("foo")
            .start();
```

a generic builder would be better but probably you have already discussed it.

@bensigelman @yurishkuro @michaelsembwever  could you please review?


yurishkuro:
lgtm

wu-sheng:
Reasonable  changes.

bhs:
(going through pending PRs)

Issue #82, opened 2017-01-06
MockSpan add toString method


pavolloffay:
Hello,

I'm using MockSpan for demo/debug purposes. It would simplify my life if `MockSpan` could be easily converted to string (with all tags, logs..), therefore I would like to add `toString` method to `MockSpan` class.

What do you think? @bensigelman

Issue #104, opened 2017-02-23
Move abstract tracer to contrib


yurishkuro:
This comes over and over again (most recently [here](https://github.com/opentracing/opentracing-java/pull/68#issuecomment-281818073)), I think we should address this. The reasons for moving abstract tracer out of `opentracing-java` are:
  1. it is not used by many (most?) existing production tracers
  1. it does not necessarily reflect how the tracers should be implemented
  1. it has some unexpected, but hard to remove design choices  (#55)
  1. it often confuses newcomers being in the official OT Java repo
  1. it's not clear if there are active maintainers of `opentracing-java` who also want to maintain the abstract tracer (and they can still do it in contrib)
  1. abstract tracer is not needed for development of OT API itself, because implementation questions can be tried out in Noop and Mock tracers, both of which do not extend the abstract tracer.

bhs:
I will cc @michaelsembwever since he is most likely to come to its defense. :)  That said, I haven't heard from mck in many moons so I'm not confident that he has cycles to weigh in.

I am wildly in favor of this move and merely lack the time to actually do the mechanical work involved. If we're unable to find the time to move it, at least we could change the SpanContext/SpanBuilder relationship which is surprising to all comers.

wu-sheng:
@yurishkuro +1, I have explained so many times to my co-workers... Many of them thought they are part of Spec or Spec-java, and must be followed when implement OT-SPEC-JAVA.

I am glad about this.

sjoerdtalsma:
@yurishkuro 👍 
My biggest gripe with this is it forces subclasses to become residents of the `io.opentracing.impl` package which could become real crowded real soon if this continues to be encouraged.

bhs:
Closed via #121.

PR #108, opened 2017-03-16
MockSpan toString method

pavolloffay:
`MockSpan#toString()` for debug purposes. It also simplifies implementation of "`LoggingTracer`"

```java
class LoggingTracer extends MockTracer {
        protected void onSpanFinished(MockSpan mockSpan) {
            System.out.println(mockSpan);
        }
    }
```

It closes https://github.com/opentracing/opentracing-java/issues/82

### DIFF
--- a/opentracing-mock/src/main/java/io/opentracing/mock/MockSpan.java
+++ b/opentracing-mock/src/main/java/io/opentracing/mock/MockSpan.java
@@ -283,4 +283,13 @@ public final class MockSpan implements Span {
             throw ex;
         }
     }
+
+    @Override
+    public String toString() {
+        return "{" +
+                "traceId:" + context.traceId() +
+                ", spanId:" + context.spanId() +
+                ", parentId:" + parentId +
+                ", operationName:\"" + operationName + "\"}";
+    }
 }


pavolloffay:
@bhs @yurishkuro could you please review?

yurishkuro:
In Jaeger clients we tend to only include span context + span name in the toString output. Are you sure you want all of the fields? If so, perhaps a JSON representation would make more sense.

Aside from that I'm ok with this change.

pavolloffay:
I was thinking about it too, however I could not find any contra argument against all fields. Actually, it can help with debugging and It can be also used for simple demo as mentioned above (when https://github.com/opentracing/opentracing-java/issues/104 is moved to contrib there could be some logging tracer for demo purposes). 

+1 on JSON.

pavolloffay:
Actually is JSON really necessary here? Implement it properly for all logs/tags is not very clean in java 1.6 without introducing any 3rd party lib.

Better would be to stick with `{spanId: 1, traceId: 1, parentId:0, operationName: "name"}` or do not use JSON and just call `#toString()` on all span's properties.

yurishkuro:
> `{spanId: 1, traceId: 1, parentId:0, operationName: "name"}`

+1 (only better order them traceId, spanId, parentId)

bhs:
(LGTM)

bhs:
I imagine this is uncontroversial, so let's merge. thanks again.

PR #118, opened 2017-04-03
Add toString() to no-op implementations.

sjoerdtalsma:
The [GlobalTracer PR](https://github.com/opentracing/opentracing-java/pull/109/commits/dc3550b7027627e295776be27387c6b1d1ee5a44) added its delegate tracer to the toString representation which is initially the NoopTracer.
 The standard Java toString is the fully qualified classname (i.e. `io.opentracing.NoopSpanImpl`) followed by its hashcode (which is irrelevant for a singleton).
 Therefore I suggest using `NoopSpan`, `NoopSpanBuilder`, `NoopSpanContext` and `NoopTracer` as their toString() representations.


yurishkuro:
Looks non-controversial, merging.

PR #119, opened 2017-04-07
Define tag SAMPLING_PRIORITY as an integer

pavolloffay:
It is inconvenient to use `short` for this. It's a breaking change but I think it's cleaner/easier to use `integer`.

Spec defines it as an `integer` https://github.com/opentracing/specification/blob/master/semantic_conventions.yaml#L26

### DIFF
--- a/opentracing-api/src/main/java/io/opentracing/tag/ShortTag.java
+++ b/opentracing-api/src/main/java/io/opentracing/tag/ShortTag.java
@@ -13,6 +13,7 @@
  */
 package io.opentracing.tag;
 
+@Deprecated
 public class ShortTag extends AbstractTag<Short> {
     public ShortTag(String key) {
         super(key);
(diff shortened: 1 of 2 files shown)


pavolloffay:
see how it is used in jaeger-java https://github.com/uber/jaeger-client-java/search?utf8=%E2%9C%93&q=Tags.SAMPLING_PRIORITY&type=

wu-sheng:
LGTM.  In JAVA, short is inconvenient to use.

opentracing-api/src/main/java/io/opentracing/tag/ShortTag.java
```diff
@@ -13,6 +13,7 @@
  */
 package io.opentracing.tag;
 
+@Deprecated
```
pavolloffay: There won't be any usage of this, therefore deprecating.

wu-sheng: Your means, we do NOT encourage using `ShortTag` in the future. If so, we should add a comment to explain this.

bhs:
(also LGTM)

pavolloffay:
Added javadoc that `ShortTag` is deprecated and use `IntTag`.

Let's wait 48h or so and merge.

wu-sheng:
Merged, since this waits more than 6 days....

PR #120, opened 2017-04-15
Make javadoc work and update license headers

bhs:
Carving this out of #115 to reduce noise in the diff per @yurishkuro's request.


bhs:
I will just merge this since I can't imagine the slightest controversy about it.

PR #121, opened 2017-04-15
Address two main sources of cruft in opentracing-java

### LINKED ISSUE #44: Reorganize modules
I propose the following layout
- `opentracing-api` module remains as is
- `opentracing-mock` module remains as is
- `opentracing-impl` is renamed to `opentracing-noop`, and `propagation` package removed from it, or moved to `-impl-java8` as that's the only thing that uses it
- `-impl-java8` is removed from this repo completely and migrated to `basictracer-java` repo, where it can give rise to the actual reference implementation, not just an SPI

### LINKED ISSUE #46: Rename `-impl` and `-impl-java8` modules to better reflect what they are
These modules are meant to be abstract skeletons for tracer implementations. Calling them "impl" is misleading to the people visiting the repo.

### LINKED ISSUE #55: Why does Tracer.SpanBuilder extend SpanContext?

bhs:
This PR has two goals:
- Remove the inheritance relationship between `Tracer.SpanBuilder` and `SpanContext` (which has baffled many people and frankly never made sense to me). Closes #55.
- Remove the `opentracing-impl` module entirely. The core `opentracing-java` github repo should be as lean as possible and exists to define the OpenTracing API contract for callers; it is not intended to be a source of helper code for `Tracer` implementations: such a thing might be valuable, but would be better place into a repo in the `opentracing-contrib` organization. Also, no known production `Tracer` implementations actually use `opentracing-impl`, nor does `opentracing-mock`. Closes #46. Closes #44 (... or at least makes it obsolete).

cc @yurishkuro @pavolloffay @objectiser @wu-sheng @sjoerdtalsma who have expressed enthusiasm about various pieces of this in the past. I was thinking it would be nice to merge this before #115, as that PR would be simpler without the -impl stuff getting in the way.

Also cc @michaelsembwever who has been MIA for a while but probably won't like this. :-/  I would like to give you an opportunity to argue against it.


bhs:
I def won't merge this for at least a couple of days in case this angers anybody :)

bhs:
Given the apparent unanimity here, I am just going to merge this so that the #115 review will be easier to read.

PR #122, opened 2017-04-26
Added SpanUtils: utility methods intended to help instrumenters follow the semantic conventions

jakerobb:
I've just begun instrumenting a set of Java services that interact using HTTP/REST and Kafka. We're using OT-api with a Zipkin backend. I'm new to every aspect of this (this is even my first PR); I'm sure there are ways this could be better, but please be gentle. :)

In short order, I found myself writing code that seemed broadly useful. I asked on Gitter and was told that this was a good idea and I should submit a PR. And so:

I've created SpanUtils, a first stab at a set (currently of size 1) of utility methods intended to help instrumenters follow the semantic conventions.

### DIFF
--- a/opentracing-util/src/main/java/io/opentracing/util/SpanUtils.java
+++ b/opentracing-util/src/main/java/io/opentracing/util/SpanUtils.java
@@ -1,0 +1,57 @@
+/**
+ * Copyright 2016-2017 The OpenTracing Authors
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License"); you may not use this file except
+ * in compliance with the License. You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software distributed under the License
+ * is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express
+ * or implied. See the License for the specific language governing permissions and limitations under
+ * the License.
+ */
+package io.opentracing.util;
+
+import io.opentracing.Span;
+import io.opentracing.tag.Tags;
+
+import java.io.PrintWriter;
+import java.io.StringWriter;
+import java.util.HashMap;
+import java.util.Map;
+
+/**
+ * Provides convenience methods for logging and tagging spans.
+ *
+ * Follows conventions defined in https://github.com/opentracing/specification/blob/master/semantic_conventions.md
+ */
+public class SpanUtils {
+    private SpanUtils() {
+    }
+
+    /**
+     * Tags the span as error=true and logs the occurrence of any Throwable (including all appropriate fields).
+     *
+     * @param span the span on which the error should be tagged/logged
+     * @param t throwable which will be included on the log entry including stacktrace
+     * @param message message for the log entry.
+     */
+    public static void logException(Span span, Throwable t, String message) {
+        Map<String, Object> fields = new HashMap<String, Object>();
+        fields.put("error.kind", "Exception");
+        fields.put("error.object", t);
+        fields.put("event", "error");
+        fields.put("message", message);
+        fields.put("stack", getStacktraceAsString(t));
+
+        span.setTag(Tags.ERROR.getKey(), true);
+        span.log(fields);
+    }
+
+    private static String getStacktraceAsString(Throwable t) {
+        StringWriter stackWriter = new StringWriter();
+        t.printStackTrace(new PrintWriter(stackWriter));
+        return stackWriter.toString();
+    }
+}
--- a/opentracing-util/src/main/java/io/opentracing/util/SpanUtils.java
+++ b/opentracing-util/src/main/java/io/opentracing/util/SpanUtils.java
@@ -16,8 +16,6 @@ package io.opentracing.util;
 import io.opentracing.Span;
 import io.opentracing.tag.Tags;
 
-import java.io.PrintWriter;
-import java.io.StringWriter;
 import java.util.HashMap;
 import java.util.Map;
 
@@ -33,25 +31,46 @@ public class SpanUtils {
     /**
      * Tags the span as error=true and logs the occurrence of any Throwable (including all appropriate fields).
      *
-     * @param span the span on which the error should be tagged/logged
-     * @param t throwable which will be included on the log entry including stacktrace
-     * @param message message for the log entry.
+     * This is a shortcut for {@link SpanUtils#logException(Span, Throwable, String) logException(Span, Throwable, null)}.
+     * If there is additional context you can provide for the error, consider calling that method and including that context
+     * in the message argument.
+     *
+     * @param span the span to which the tag/log should be applied.
+     * @param t throwable which will be included on the log entry
+     */
+    public static void logException(Span span, Throwable t) {
+        logException(span, t, null);
+    }
+
+    /**
+     * Tags the span as error=true and logs the occurrence of any Throwable (including all appropriate fields).
+     *
+     * Note: the semantic conventions also specify a field called "stack". This will not be populated here (due
+     * to performance concerns), but the Tracer can choose to do so based on the value we store in "error.object".
+     *
+     * @param span the span to which the tag/log should be applied.
+     * @param t throwable which will be included on the log entry
+     * @param message message for the log entry. This will be concatenated with (and therefore should not be the
+     *                same as) the exception message. Null- and empty-tolerant.
      */
     public static void logException(Span span, Throwable t, String message) {
+        if (span == null) {
+            throw new IllegalArgumentException("Span must not be null.");
+        }
+        if (t == null) {
+            throw new IllegalArgumentException("Throwable must not be null.");
+        }
         Map<String, Object> fields = new HashMap<String, Object>();
-        fields.put("error.kind", "Exception");
-        fields.put("error.object", t);
         fields.put("event", "error");
-        fields.put("message", message);
-        fields.put("stack", getStacktraceAsString(t));
+        if (message == null || message.trim().isEmpty()) {
+            fields.put("message", t.getMessage());
+        } else {
+            fields.put("message", String.format("%s | %s", message, t.getMessage()));
+        }
+        fields.put("error.kind", t.getClass().getName());
+        fields.put("error.object", t);
 
         span.setTag(Tags.ERROR.getKey(), true);
         span.log(fields);
-    }
-
-    private static String getStacktraceAsString(Throwable t) {
-        StringWriter stackWriter = new StringWriter();
-        t.printStackTrace(new PrintWriter(stackWriter));
-        return stackWriter.toString();
     }
 }
--- a/opentracing-api/src/main/java/io/opentracing/Field.java
+++ b/opentracing-api/src/main/java/io/opentracing/Field.java
@@ -1,0 +1,60 @@
+/**
+ * Copyright 2016-2017 The OpenTracing Authors
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License"); you may not use this file except
+ * in compliance with the License. You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software distributed under the License
+ * is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express
+ * or implied. See the License for the specific language governing permissions and limitations under
+ * the License.
+ */
+package io.opentracing;
+
+/**
+ * The following log fields are recommended for instrumentors who are trying to capture more
+ * information about a logged event. Tracers may expose additional features based on these
+ * standardized data points.
+ *
+ * @see <a href="https://github.com/opentracing/specification/blob/master/semantic_conventions.md">https://github.com/opentracing/specification/blob/master/semantic_conventions.md</a>
+ */
+public enum Field {
+    /**
+     * A stable identifier for some notable moment in the lifetime of a Span. For instance, a mutex
+     * lock acquisition or release or the sorts of lifetime events in a browser page load described
+     * in the Performance.timing specification. E.g., from Zipkin, "cs", "sr", "ss", or "cr". Or,
+     * more generally, "initialized" or "timed out". For errors, "error"
+     */
+    EVENT("event"),
+
+    /**
+     * A concise, human-readable, one-line message explaining the event. E.g., "Could not connect
+     * to backend", "Cache invalidation succeeded"
+     */
+    MESSAGE("message"),
+
+    /**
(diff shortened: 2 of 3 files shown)


opentracing-util/src/main/java/io/opentracing/util/SpanUtils.java
```diff
@@ -0,0 +1,57 @@
...
+    /**
+     * Tags the span as error=true and logs the occurrence of any Throwable (including all appropriate fields).
+     *
+     * @param span the span on which the error should be tagged/logged
+     * @param t throwable which will be included on the log entry including stacktrace
+     * @param message message for the log entry.
+     */
+    public static void logException(Span span, Throwable t, String message) {
+        Map<String, Object> fields = new HashMap<String, Object>();
+        fields.put("error.kind", "Exception");
```
devinsba: I'm not sure, but maybe this should be the specific exception type: `t.getClass()`

opentracing-util/src/main/java/io/opentracing/util/SpanUtils.java
```diff
@@ -0,0 +1,57 @@
...
+        fields.put("error.object", t);
+        fields.put("event", "error");
+        fields.put("message", message);
```
devinsba: It might make more sense to take the message from the exception here `t.getMessage()`

opentracing-util/src/main/java/io/opentracing/util/SpanUtils.java
```diff
@@ -0,0 +1,57 @@
...
+import java.io.StringWriter;
+import java.util.HashMap;
+import java.util.Map;
+
+/**
+ * Provides convenience methods for logging and tagging spans.
+ *
+ * Follows conventions defined in https://github.com/opentracing/specification/blob/master/semantic_conventions.md
+ */
+public class SpanUtils {
```
devinsba: I wonder if this should be specific to exceptions instead of Spans. Not sure if we have to worry about SpanUtils getting too large

opentracing-util/src/main/java/io/opentracing/util/SpanUtils.java
yurishkuro: It is a good practice to give context to the exceptions in the code that catches them, so the `message` argument to the function is a good thing and generally should be different from `t.getMessage()`. But when logging the message it makes sense to concat the two messages.

opentracing-util/src/main/java/io/opentracing/util/SpanUtils.java
```diff
@@ -0,0 +1,57 @@
...
+        fields.put("error.object", t);
+        fields.put("event", "error");
+        fields.put("message", message);
+        fields.put("stack", getStacktraceAsString(t));
+
+        span.setTag(Tags.ERROR.getKey(), true);
+        span.log(fields);
+    }
+
+    private static String getStacktraceAsString(Throwable t) {
```
yurishkuro: I think this is best left to the tracer, because it may choose not to do that when the trace is not sampled, to avoid wasting perf and heap.

opentracing-util/src/main/java/io/opentracing/util/SpanUtils.java
yurishkuro: actually, if `t` is already logged as `error.object`, I think specifying the kind is redundant, the tracer can do that internally if needed.

What's the reason for using string constants for tag/field names?  Don't we have constants defined for all tags, per https://github.com/opentracing/specification/blob/master/semantic_conventions.md ?

opentracing-util/src/main/java/io/opentracing/util/SpanUtils.java
```diff
@@ -0,0 +1,57 @@
...
+     * @param t throwable which will be included on the log entry including stacktrace
+     * @param message message for the log entry.
+     */
+    public static void logException(Span span, Throwable t, String message) {
+        Map<String, Object> fields = new HashMap<String, Object>();
+        fields.put("error.kind", "Exception");
+        fields.put("error.object", t);
+        fields.put("event", "error");
+        fields.put("message", message);
+        fields.put("stack", getStacktraceAsString(t));
```
yurishkuro: redundant, as it's already available to the tracer via `t`

opentracing-util/src/main/java/io/opentracing/util/SpanUtils.java
jakerobb: FWIW, I considered this. 

https://github.com/opentracing/specification/blob/master/semantic_conventions.md explicitly gives "Exception" as an example, but was unclear about the intent of the field, so I wasn't sure.

The exception type will be in the stack. 

I'm happy either way.

opentracing-util/src/main/java/io/opentracing/util/SpanUtils.java
jakerobb: That will be in the stack. My intent here was to allow the caller to specify a message potentially containing some additional context.

opentracing-util/src/main/java/io/opentracing/util/SpanUtils.java
jakerobb: Fair point. The "stack" field is specified in the conventions, though. You think it's better to omit?

opentracing-util/src/main/java/io/opentracing/util/SpanUtils.java
jakerobb: I thought about that. The obvious name, "ExceptionUtils," is already in use by commons-lang among others and I think it'd be preferable to choose something else. Suggestions?

jakerobb: Food for thought: the semantic conventions document only covers four cases: RPCs, messages, database calls, and errors. Seems like there'd be at most 2 public methods for each.

My intent was to limit the scope of these methods to what's suggested in the conventions. Perhaps that could drive suggestions for a better name? SemanticConventionUtils feels a little contrived, but maybe you guys disagree?

opentracing-util/src/main/java/io/opentracing/util/SpanUtils.java
yurishkuro: The convention is written to accommodate languages with and without exceptions. I think if you can set an `error.object`, you don't need anything else (except `message with a different context than the exception message).

opentracing-util/src/main/java/io/opentracing/util/SpanUtils.java
jakerobb: Using my local snapshot of this lib, I'm thinking it would be useful to allow the caller to (optionally) specify a Map<String, Object> of additional fields to be included on the log event. This method would merge the two maps and log the event, and there'd be a convenience method for when the caller has no additional fields.

jakerobb:
I got an email notification with a comment asking why I didn't use constants from io.opentracing.tag.Tags, but that comment doesn't show up here -- I'm assuming it was edited. But I think that's a valid question that deserves an answer: there are not constants that correspond to these strings, and technically, these are fields on a log event, not tags on a span. Should these have constants in Tags? Should there be a separate set of constants for log event fields?

jakerobb:
Also, what's the going-forward practice here? Should I close this PR and submit a new one with changes? Or should I update this one?

sjoerdtalsma:
@jakerobb I normally just commit to the branch from which the PR originated and the commits will automatically show up here.

jakerobb:
I've accounted for all feedback, and matured things a bit based on my own experience using it. I feel like the name of the class is still an open point of discussion, so I left that alone.

opentracing-util/src/main/java/io/opentracing/util/SpanUtils.java
```diff
@@ -0,0 +1,76 @@
...
+    public static void logException(Span span, Throwable t, String message) {
+        if (span == null) {
+            throw new IllegalArgumentException("Span must not be null.");
+        }
+        if (t == null) {
+            throw new IllegalArgumentException("Throwable must not be null.");
+        }
+        Map<String, Object> fields = new HashMap<String, Object>();
+        fields.put("event", "error");
+        if (message == null || message.trim().isEmpty()) {
```
jakerobb: Calling trim() here might be trying to hard. Thoughts?

opentracing-util/src/main/java/io/opentracing/util/SpanUtils.java
```diff
@@ -0,0 +1,76 @@
...
+            fields.put("message", t.getMessage());
+        } else {
+            fields.put("message", String.format("%s | %s", message, t.getMessage()));
```
jakerobb: Thoughts on the delimiter?

yurishkuro:
> I got an email notification with a comment asking why I didn't use constants from io.opentracing.tag.Tags, but that comment doesn't show up here -- I'm assuming it was edited.

github collapses the comments when you push new commits that change the code in question. But you can still see them on the Conversation tab via "Show outdated"

opentracing-util/src/main/java/io/opentracing/util/SpanUtils.java
```diff
@@ -0,0 +1,76 @@
...
+        }
+        fields.put("error.kind", t.getClass().getName());
+        fields.put("error.object", t);
+
+        span.setTag(Tags.ERROR.getKey(), true);
```
yurishkuro: I don't think this is needed. A logged exception may not necessarily mean the span is in error.

opentracing-util/src/main/java/io/opentracing/util/SpanUtils.java
```diff
@@ -0,0 +1,76 @@
...
+            throw new IllegalArgumentException("Throwable must not be null.");
+        }
+        Map<String, Object> fields = new HashMap<String, Object>();
+        fields.put("event", "error");
+        if (message == null || message.trim().isEmpty()) {
+            fields.put("message", t.getMessage());
+        } else {
+            fields.put("message", String.format("%s | %s", message, t.getMessage()));
+        }
+        fields.put("error.kind", t.getClass().getName());
```
yurishkuro: I think this is redundant. We want the most performant implementation, and giving the exception object to the tracer is all that's needed

opentracing-util/src/main/java/io/opentracing/util/SpanUtils.java
yurishkuro: I don't think you need to concatenate - the goal should be to have as little code as necessary without loss of information, and by passing the Throwable to the tracer you already give it all the info it needs.

opentracing-util/src/main/java/io/opentracing/util/SpanUtils.java
```diff
@@ -0,0 +1,76 @@
...
+     * to performance concerns), but the Tracer can choose to do so based on the value we store in "error.object".
+     *
+     * @param span the span to which the tag/log should be applied.
+     * @param t throwable which will be included on the log entry
+     * @param message message for the log entry. This will be concatenated with (and therefore should not be the
+     *                same as) the exception message. Null- and empty-tolerant.
+     */
+    public static void logException(Span span, Throwable t, String message) {
+        if (span == null) {
+            throw new IllegalArgumentException("Span must not be null.");
```
yurishkuro: I would rather `return` than blow up

opentracing-util/src/main/java/io/opentracing/util/SpanUtils.java
```diff
@@ -0,0 +1,76 @@
...
+        }
+        if (t == null) {
+            throw new IllegalArgumentException("Throwable must not be null.");
```
yurishkuro: I think it could be a valid case to not have a Throwable, what if I just want to log `event:error message:xxx`?

opentracing-util/src/main/java/io/opentracing/util/SpanUtils.java
```diff
@@ -0,0 +1,76 @@
...
+        }
+        Map<String, Object> fields = new HashMap<String, Object>();
+        fields.put("event", "error");
```
yurishkuro: if we don't have constants for these standard log fields, we should define them, e.g. in `Fields` class (in a separate PR).

opentracing-util/src/main/java/io/opentracing/util/SpanUtils.java
jakerobb: Should the Tracer then be modified so that it knows to extract this information and create additional fields on the log event?

yurishkuro: The semantic specification is designed to provide more information to the tracing backend. In this case I'd leave the actual handling of the Throwable to the tracer implementation.

mabn:
What about the stacktrace?

See this implementation: https://github.com/opentracing-contrib/java-web-servlet-filter/blob/master/opentracing-web-servlet-filter/src/main/java/io/opentracing/contrib/web/servlet/filter/ServletFilterSpanDecorator.java#L115

opentracing-util/src/main/java/io/opentracing/util/SpanUtils.java
```diff
@@ -0,0 +1,76 @@
...
+     */
+    public static void logException(Span span, Throwable t) {
+        logException(span, t, null);
+    }
+
+    /**
+     * Tags the span as error=true and logs the occurrence of any Throwable (including all appropriate fields).
+     *
+     * Note: the semantic conventions also specify a field called "stack". This will not be populated here (due
+     * to performance concerns), but the Tracer can choose to do so based on the value we store in "error.object".
```
mabn: Is relying on tracer implementation for capturing stack traces the way to go?

opentracing-util/src/main/java/io/opentracing/util/SpanUtils.java
pavolloffay: Agree It should be first

opentracing-util/src/main/java/io/opentracing/util/SpanUtils.java
jakerobb: In my initial commit, I was capturing stacktraces manually, and I got feedback that it should be left up to the tracer implementation. I have no stake in either position; all I know is that for my own purposes, I will want the stack to be captured, and so it will need to happen somewhere.

opentracing-util/src/main/java/io/opentracing/util/SpanUtils.java
jakerobb: Agreed that it's a valid case, but I'd argue that's not what this method is for.

opentracing-util/src/main/java/io/opentracing/util/SpanUtils.java
yurishkuro: it is often useful to capture the reasoning (from earlier comments) in the function comments

jakerobb:
I've closed this for a few reasons:
1. I don't know what I'm doing with Git, and inadvertently added files to it that were intended for a separate PR.
2. The conversation died because I was unable to revisit feedback for two weeks
3. The feedback I was getting indicated that there was not sufficient agreement among maintainers on how this should work (e.g. whether to delegate capturing stacktraces to the tracer).

PR #127, opened 2017-04-29
Clean copy of #115

bhs:
REVIEWERS: *Feel free to look only at #115; I will keep these in lock-step until #115 merges.* ... but #115 is breaking Github, and this one hopefully won't :)

cc @cwe1ss who asked for this.

PS: I am planning to manually merge back interesting discussions into #115, btw. No need to keep track of this one unless you want to.


devinsba:
since the merge this probably should be closed up

PR #129, opened 2017-05-01
Mock baggage

bhanafee:
Implementation and tests to propagate baggage items in MockTracer.Propagator.TEXT_MAP

### DIFF
--- a/opentracing-mock/src/main/java/io/opentracing/mock/MockTracer.java
+++ b/opentracing-mock/src/main/java/io/opentracing/mock/MockTracer.java
@@ -139,11 +139,7 @@ public class MockTracer implements Tracer {
                 }
 
                 if (traceId != null && spanId != null) {
-                    if (baggage.isEmpty()) {
-                        return new MockSpan.MockContext(traceId, spanId, Collections.<String, String>emptyMap());
-                    } else {
-                        return new MockSpan.MockContext(traceId, spanId, baggage);
-                    }
+                    return new MockSpan.MockContext(traceId, spanId, baggage);
                 }
 
                 return null;
--- a/opentracing-mock/src/main/java/io/opentracing/mock/MockTracer.java
+++ b/opentracing-mock/src/main/java/io/opentracing/mock/MockTracer.java
@@ -13,7 +13,11 @@
  */
 package io.opentracing.mock;
 
-import java.util.*;
+import java.util.ArrayList;
+import java.util.Collections;
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
 
 import io.opentracing.References;
 import io.opentracing.Span;
--- a/opentracing-mock/src/main/java/io/opentracing/mock/MockTracer.java
+++ b/opentracing-mock/src/main/java/io/opentracing/mock/MockTracer.java
@@ -1,4 +1,4 @@
-/**
+/*
  * Copyright 2016-2017 The OpenTracing Authors
  *
  * Licensed under the Apache License, Version 2.0 (the "License"); you may not use this file except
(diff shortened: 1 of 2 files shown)


opentracing-mock/src/main/java/io/opentracing/mock/MockTracer.java
```diff
@@ -119,24 +121,29 @@ protected void onSpanFinished(MockSpan mockSpan) {
...
+                            baggage.put(entry.getKey(), entry.getValue());
                         }
                     }
                 } else {
                     throw new IllegalArgumentException("Unknown carrier");
                 }
 
                 if (traceId != null && spanId != null) {
-                    return new MockSpan.MockContext(traceId, spanId, Collections.<String, String>emptyMap());
+                    if (baggage.isEmpty()) {
```
yurishkuro: why do this check? Especially since you've already allocated the hashmap.

bhanafee: Purely to ensure that the behavior if there are no baggageItems exactly matches current behavior. I didn't want to affect anything that MockContext might do with the empty baggage. It's _probably_ safe to skip the check and just pass the hashmap all the time.

pavolloffay: I think this if-else is redundant.

opentracing-mock/src/test/java/io/opentracing/mock/MockTracerTest.java
```diff
@@ -156,16 +157,21 @@ public void testTextMapPropagatorTextMap() {
...
+                    .start();
+            childSpan.setBaggageItem("barbag", "baritem");
+            childSpan.finish();
         }
         List<MockSpan> finishedSpans = tracer.finishedSpans();
 
         Assert.assertEquals(2, finishedSpans.size());
         Assert.assertEquals(finishedSpans.get(0).context().traceId(), finishedSpans.get(1).context().traceId());
         Assert.assertEquals(finishedSpans.get(0).context().spanId(), finishedSpans.get(1).parentId());
+        Assert.assertEquals(finishedSpans.get(0).getBaggageItem("foobag"), "fooitem");
```
pavolloffay: Could you please change the order of these arguments to align with`Assert.assertEquals(expected, actual)`. (lines 173, 174 too)

bhanafee:
Requested changes complete and pushed

bhs:
This seems stalled but looks good to me and has no pending PR comments – are there any concerns from @yurishkuro or @pavolloffay?

pavolloffay:
I think a prefix is missing there. 

We should standardize it across all mock impls. E.g. Go and Java are using different identifiers.

pavolloffay:
https://github.com/opentracing/specification/issues/70

yurishkuro:
Agree with @pavolloffay - if this is meant to actually be used over the wire, it's better to have some prefix like `baggage-` or `ot-baggage-`, otherwise baggage can just clobber normal http headers.

bhanafee:
Seems like we're mixing concerns here. The very first comment on the issue says "The mock tracer is not supposed to be used in real scenarios." The only real requirement here should be that the baggage items that go in should come back out.

For anything that *does* go over the wire, it's a longer though perfectly legitimate conversation. From the perspective of the PR, I think the question is just whether it would make resolution of https://github.com/opentracing/specification/issues/70 any more difficult to implement.

yurishkuro:
@bhanafee fair enough, I'm ok to merge this as is.

bhs:
I agree with @bhanafee... MockTracer is designed for unittests and doesn't need to be multi-lingual-consistent. I have no problem with standardizing across languages where convenient/easy, of course, but that's a nice-to-have in my book. I think we ought to merge this in 24-48h sans objections.

pavolloffay:
Agree, with header normalization can be done in next iterations, however this impl can override what is sent over the wire as there is no prefix. 

Now we can go with anything `baggage-`, `ot-mock-baggage` and later change it as should be an implementation detail.

bhanafee:
@pavolloffay Are you asking me to add a prefix before this PR is processed? If so:

- Do you care what the prefix is?
- Do you want a prefix applied to SPAN_ID_KEY and TRACE_ID_KEY too, or leave them as-is?

pavolloffay:
Yes please, only for baggage. 

 `baggae-` would be enough or `ot-baggage`

pavolloffay:
@bhanafee thanks, it looks good I will merge it.

PR #131, opened 2017-05-03
Remove deprecated log methods and ShortTag

pavolloffay:
Remove deprecated `Span#log` methods and `ShortTag`.

cc @opentracing/otsc @opentracing/otiab


bhs:
@pavolloffay I'm ok killing ShortTag but feel a little more conflicted about the deprecated logging methods. I'm curious what others think...

basvanbeek:
TBH I'm against the deprecation of the logging methods. They make much more sense for tracing systems that do not consider structured logging as part of their ecosystem like Zipkin. 

The current log methods (with a null payload) as well as the opentracing-go version:
```go
LogEvent(event string)
```

map so much nicer with Zipkin's Annotations.

bhs:
@basvanbeek 

> TBH I'm against the deprecation of the logging methods.

I agree about the event-logging method... @adriancole convinced me a while back that there's value in special-casing that case. I am fine deprecating (or removing, since we may be bumping the major version soon) the versions that have a "payload" Object.

pavolloffay:
Rebased,

@bhs this PR removes only log methods with "Object payload". ` S log(String event);` is still there. 

@basvanbeek it seems that `Object payload` is ignored in brave-ot
https://github.com/openzipkin/brave-opentracing/blob/master/src/main/java/brave/opentracing/BraveSpan.java#L169
https://github.com/openzipkin/brave-opentracing/blob/master/src/main/java/brave/opentracing/BraveSpan.java#L176

bhs:
@pavolloffay thank you for the adjustments. I would be glad to see the `Object payload` variants go away.

I'd like to hear back from @basvanbeek for sure. We also need to figure out whether there are any simple/boring/non-breaking changes we definitely want to incorporate before we starting merging the breaking PRs (this one, the `noop` PR #150, etc).

basvanbeek:
I'm good with removing the `Object payload` variants. The payload data did not make a lot of sense in how Zipkin handles timestamped annotations.

pavolloffay:
I will also re cc @opentracing/otsc and @opentracing/otiab :)

opentracing-mock/src/test/java/io/opentracing/mock/MockTracerTest.java
```diff
@@ -64,21 +62,15 @@ public void testRootSpan() {
         assertEquals(7, tags.get("int"));
         assertEquals("foo", tags.get("string"));
         List<MockSpan.LogEntry> logs = finishedSpan.logEntries();
-        assertEquals(3, logs.size());
+        assertEquals(2, logs.size());
         {
             MockSpan.LogEntry log = logs.get(0);
-            assertEquals(1001, log.timestampMicros());
-            assertEquals("event name", log.fields().get("event"));
```
yurishkuro: unless there is another test elsewhere, I would keep this test but instead of log(event, payload) only test log(event)

tedsuo:
@pavolloffay is this ready to merge?

opentracing-mock/src/test/java/io/opentracing/mock/MockTracerTest.java
pavolloffay: Don't think it is necessary related to this PR, but the test for `log(str)` has been added.

pavolloffay:
Yes, it is ready to go in.

PR #136, opened 2017-05-11
Add RC releasing to travis publish script

devinsba:
This is the format @bhs tried to use. I've seen `0.0.0.RC1` from the spring projects so I'm not sure which would make more sense to people


pavolloffay:
Maybe it's better to do the release locally for these non-standard versions.

bhs:
I really like the Travis releases... so much less of a pain in the a**, and this way we can at least standardize on "spelling" for these things. Let's go with the spring format since it's a thing already.

bhs:
I pushed the change to use the Spring format... will merge shortly.

PR #142, opened 2017-05-15
Enable coveralls

pavolloffay:


.travis.yml
```diff
@@ -39,6 +39,9 @@ branches:
   except:
     - /^[0-9]/
 
+after_success:
+  - mvn clean test jacoco:report coveralls:report
```
yurishkuro: this causes the tests to be run twice, since `publish.sh` also runs them

pavolloffay: good point, removed

PR #143, opened 2017-05-15
Change copyright comment in java files from /** */ style to /* */.

sjoerdtalsma:
This avoids automatically adding `<P>` between the paragraphs every time an IDE (like IntelliJ) auto-formats the code.

### DIFF
--- a/opentracing-mock/src/main/java/io/opentracing/mock/MockSpan.java
+++ b/opentracing-mock/src/main/java/io/opentracing/mock/MockSpan.java
@@ -1,4 +1,4 @@
-/**
+/*
  * Copyright 2016-2017 The OpenTracing Authors
  *
  * Licensed under the Apache License, Version 2.0 (the "License"); you may not use this file except
@@ -16,11 +16,7 @@ package io.opentracing.mock;
 import io.opentracing.Span;
 import io.opentracing.SpanContext;
 
-import java.util.ArrayList;
-import java.util.Collections;
-import java.util.HashMap;
-import java.util.List;
-import java.util.Map;
+import java.util.*;
 import java.util.concurrent.atomic.AtomicLong;
 
 /**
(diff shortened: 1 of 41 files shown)


opentracing-mock/src/main/java/io/opentracing/mock/MockSpan.java
```diff
@@ -16,11 +16,7 @@
 import io.opentracing.Span;
 import io.opentracing.SpanContext;
 
-import java.util.ArrayList;
-import java.util.Collections;
-import java.util.HashMap;
-import java.util.List;
-import java.util.Map;
+import java.util.*;
```
yurishkuro: this is not recommended by many style checks

sjoerdtalsma: Good catch, must've been a side-effect from my IntelliJ settings, I'll correct this!

opentracing-util/src/main/java/io/opentracing/util/ThreadLocalActiveSpan.java
```diff
@@ -22,13 +22,13 @@
...
 import java.util.concurrent.atomic.AtomicInteger;
 
- /**
-  * {@link ThreadLocalActiveSpan} is a simple {@link ActiveSpan} implementation that relies on Java's
-  * thread-local storage primitive.
-  *
-  * @see ActiveSpanSource
-  * @see Tracer#activeSpan()
-  */
+/**
```
pavolloffay: Did you change this manually?

sjoerdtalsma: Nope, must have been an IntelliJ autoformat.
By the way, I agree with you on the license-only commit; that was my intention, the formatting 'came free' by the tooling :wink:

sjoerdtalsma:
I'll change the code to "license-only" and try to revert all formatting changes, please don't merge yet.

sjoerdtalsma:
Looks like the accidental auto-import changes were finally reverted.
Sorry guys: turns out I had "Optimize imports" flag on at "before commit" and since all Java files were touched by the license plugin this problem got re-introduced several times even after I explicitly reverted the import changes :man_facepalming:

PR #144, opened 2017-05-15
Remove unnecessary imports from Span.java.

sjoerdtalsma:
The imports for `Closeable` and `Map` were no longer used.


yurishkuro:
Thanks!

PR #148, opened 2017-05-28
Qualify use of GlobalTracer

### LINKED ISSUE #140: OpenTracing encourages classloader-scoped configuration
I find it very strange that in 2017, OpenTracing are encouraging and proliferating classloader-scoped configuration as status quo. Considering all the work in the last decade in JVM configuration, it is surprising that new instrumentation are being made that rely *first* on the expanding new set of apis like `DefaultSpanManager.getInstance()`. There are numerous better ways to accomplish the task. For example, why would you configure your service with CDI, Guice, Spring, OSGi, Dagger, yet have it poke around for classloader statics?

The only reasons I can think of for this optimization are:
* people who don't want to be bothered with integrating with various configuration methodologies, and find it more convenient to route users through static config instead.
* people here are trying to emulate existing logging frameworks, except loggers are different as they often live far below request (ex at bootstrap time), and have to include a system like static initialization.

Rather than proliferate dozens of projects that pin people to statics, I would suggest putting better advice here. There's a lot of instrumentation manufacturing going on which don't seem to gel with best practices. I suspect if people knew better, they would make static lookup as a last resort or opt-in vs *the way* to configure tracing.

bhs:
Inasmuch as a singleton mechanism is needed in some places, it ought to be part of OT-Java's core API. That said, it's not a recommended best practice when more modern options are available.

Closes #140.

### DIFF
--- a/README.md
+++ b/README.md
@@ -22,7 +22,13 @@ Packages are deployed to Maven Central under the `io.opentracing` group.
 
 ### Initialization
 
-Initialization is OpenTracing-implementation-specific. Generally speaking, the pattern is to initialize a `Tracer` once for the entire process and to use that `Tracer` for the remainder of the process lifetime. The [GlobalTracer](https://github.com/opentracing/opentracing-java/blob/master/opentracing-util/src/main/java/io/opentracing/util/GlobalTracer.java) provides a helper for singleton access to the `Tracer`.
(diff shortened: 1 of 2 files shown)


bhs:
(cc @adriancole)

bhs:
I'll merge this since it's an improvement and only affects documentation – if there are concerns large or small, happy to address in follow-ons.

PR #150, opened 2017-05-30
Fixes #145 - Rename package in opentracing-noop to io.opentracing.noop

### LINKED ISSUE #145: Identical packages in opentracing-api and opentracing-noop cause build failure with JDK 9 + jigsaw
Maven dependency on opentracing-util required to use GlobalTracer.get() introduced transitive dependency on opentracing-noop and opentracing-api.
Unfortunately both opentracing-noop and opentracing-api contains the same package "io.opentracing" and this causes error in JDK 9 + jigsaw projects containing module-info.java.

[ERROR] COMPILATION ERROR : 
[INFO] -------------------------------------------------------------
[ERROR] module <module_name> reads package io.opentracing from both opentracing.api and opentracing.noop
[ERROR] module  reads package io.opentracing from both opentracing.noop and opentracing.api
[ERROR] module opentracing.api reads package io.opentracing from both opentracing.noop and opentracing.api
[ERROR] module opentracing.util reads package io.opentracing from both opentracing.noop and opentracing.api
[ERROR] module opentracing.noop reads package io.opentracing from both opentracing.noop and opentracing.api

It would be good to rename "io.opentracing" package in opentracing-noop to something like "io.opentracing.noop"

kratz74:
Signed-off-by: Tomas Kraus <tomas.kraus@oracle.com>

### DIFF
--- a/opentracing-mock/src/main/java/io/opentracing/mock/MockTracer.java
+++ b/opentracing-mock/src/main/java/io/opentracing/mock/MockTracer.java
@@ -16,7 +16,7 @@ package io.opentracing.mock;
 import io.opentracing.ActiveSpan;
 import io.opentracing.ActiveSpanSource;
 import io.opentracing.BaseSpan;
-import io.opentracing.NoopActiveSpanSource;
+import io.opentracing.noop.NoopActiveSpanSource;
 import io.opentracing.References;
 import io.opentracing.Span;
 import io.opentracing.SpanContext;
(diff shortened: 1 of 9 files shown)


kratz74:
@bhs Just sending requested notification.

bhs:
@kratz74 thank you.

The good news is that this is an improvement. The bad news is that it's a breaking change. Since @pavolloffay offered to think of alternative solutions on #145 I would like to wait for him to offer an opinion before merging. Since we're about to do a 1.0 for OT-Java, my sense is that we should probably merge this since it'll just be more painful to do later.

pavolloffay:
Based on the discussion with @judovana, It seems that this is the best solution to this.

bhs:
@pavolloffay thanks.

I guess we ought to merge, then? Not sure what reasonable alternative we have.

bhs:
(I'll give it 24h)

bhs:
⏰ went off... merged.

Thanks @kratz74!

PR #152, opened 2017-05-31
toString for thread local span

pavolloffay:


tedsuo:
@pavolloffay this is approved, can we go ahead and merge?

PR #153, opened 2017-06-05
Add constants for log fields

### LINKED ISSUE #146: Define constants for standard logging fields
Something similar to now-closed #122

My preference is to follow the `tag` package and have `log` with `log.Fields` class listing the constants. For now they can be simple string constants.

yurishkuro:
Fixes #146


pavolloffay:
wow this happened super fast. Could we wait some time before merging? Quite some people complained that changes here happen very quickly.

yurishkuro:
My bad

Issue #157, opened 2017-06-13
GlobalTracer::tracer should be volatile


raphw:
As the [GlobalTracer::tracer](https://github.com/opentracing/opentracing-java/blob/master/opentracing-util/src/main/java/io/opentracing/util/GlobalTracer.java#L68) is only set in a synchronized block but might be dereferenced outside of it, the field must be volatile.

wu-sheng:
Although these is little chance about initializing `GlobalTracer` in multi-thread, IMO it should initialize in starting stage, still this is a risk. I have send #158 for this.

wu-sheng:
Has already added. Thanks

PR #158, opened 2017-06-13
Add volatile keyword.

wu-sheng:
Reply for #157


bhs:
Sorry for the delay. This LGTM... any code that calls on the GlobalTracer frequently enough to care about a read from main memory is "doing it wrong".

If there are no objections I think we ought to merge this in 48h. cc @yurishkuro @pavolloffay

yurishkuro:
+1

bhs:
Thanks @wu-sheng

PR #161, opened 2017-06-21
Increase test coverage

pavolloffay:


opentracing-api/src/test/java/io/opentracing/propagation/TextMapExtractAdapterTest.java
```diff
@@ -0,0 +1,49 @@
+/*
+ * Copyright 2016-2017 The OpenTracing Authors
```
yurishkuro: just 2017, I think, since it's a new file

pavolloffay: tried, it failed the build.

Issue #162, opened 2017-06-23
Tags.PEER_HOST_IPV4 is IntTag but spec says it should be String


szetheli:
Current version of spec calls for peer.ipv4 to be type string, but the java impl Tags has it as IntTag.  This leads to confusing output for the IP address in the impls.

peer.ipv4=2130706433 (as integer)

peer.ipv4=127.0.0.1 (as spec calls for)

Should the Tags class change to match the spec or the spec change and call out that implementations should special case this if they want readability?

wu-sheng:
@opentracing/otsc I think we have discussed similar scenario in `peer.port`, the result is we change the PEER_PORT from **ShortTag** to **IntTag**. IMO, we should change **ipv4** to **StringTag** too.

yurishkuro:
it's a breaking change. Not worth it, imo. Maybe for 1.0

wu-sheng:
Sorry for missing this info. Intend to change for 1.0

yurishkuro:
@szetheli you can use `peer.address` tag, it accepts strings.

wu-sheng:
@szetheli Now ipv4 can accept integer and string both.

PR #163, opened 2017-06-24
Change Tags.PEER_HOST_IPV4 to a StringTag

wu-sheng:
Related #162


yurishkuro:
breaking change with little benefit.

wu-sheng:
Set milestone to 1.0

tylerbenson:
With all the other breaking changes going into this next release, I think we should get this in sooner than later.

wu-sheng:
+1 on merging this in RC. @opentracing/otsc

yurishkuro:
We recently made a similar change but in a backwards-compatible way. Why can't we keep the tag as _both_ int and string, and mark the `int` function deprecated?

wu-sheng:
>  Why can't we keep the tag as both int and string, and mark the int function deprecated?

Definitely we can. I think we have broken somethings from 0.30.0 to RC, so maybe we break things together. :P  And provide a more perfect(Remove `@deprecated`s in 0.x) version in 1.0.  Thoughts? @yurishkuro

yurishkuro:
Having broken something with 0.30 is not the best reason to continue breaking people, especially with such a minor improvement that can be easily made backwards compatible. We can leave for the future the decision when `@deprecated` methods can be removed (I'd say in the next major, 2.x)

wu-sheng:
@yurishkuro Sure, I can do some adjustments. But do you have any suggestions about naming? since we will have two ipv4 tag variables.

yurishkuro:
I am suggesting a single constant, but with two methods, `set(BaseSpan<?> span, String tagValue)` and `set(BaseSpan<?> span, int tagValue)`. E.g. you can subclass IntTag <- IntOrStringTag

wu-sheng:
@yurishkuro Sure, I will do this at night.

wu-sheng:
@yurishkuro @tylerbenson I opened a new pull request based on the current master. Closing this. Please check  #184 for more discussion.

PR #164, opened 2017-06-29
Fixes release pattern

adriancole:
A recent change to the ./mvnw script changed output slightly and broke
the logic which detected versions. This fixes it.

This should be applied to anything that uses this script

### DIFF
--- a/travis/publish.sh
+++ b/travis/publish.sh
@@ -72,7 +72,7 @@ check_release_tag() {
 }
 
 is_release_commit() {
-  project_version=$(./mvnw help:evaluate -N -Dexpression=project.version|grep -v '\[')
+  project_version=$(./mvnw help:evaluate -N -Dexpression=project.version|sed -n '/^[0-9]/p')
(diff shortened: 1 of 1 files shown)


travis/publish.sh
```diff
@@ -72,7 +72,7 @@ check_release_tag() {
 }
 
 is_release_commit() {
-  project_version=$(./mvnw help:evaluate -N -Dexpression=project.version|grep -v '\[')
+  project_version=$(./mvnw help:evaluate -N -Dexpression=project.version|sed -n '/^[0-9]/p')
```
adriancole: this makes sure no extra output like download or PWD end up in the version. Notably, this could cause a release to not sync to maven central automatically

bhs: thanks!

bhs: Oh, one thing, though... this script allows `RC` releases...

```
"You must specify a tag of the format 'release-0.0.0' or 'release-0.0.0.RC0' to release this project."
```
https://github.com/opentracing/opentracing-java/pull/164/files#diff-f515214ad913b3a43604369f3a277d96R68

adriancole: thx for the heads up.. this pattern only looks at the start of the line, so shouldn't hurt that.

adriancole:
build fail is unrelated

Issue #173, opened 2017-09-03
MockTracer's activeSpan is not finished in try-resource block.


skrityak:
The MockTracer example in https://github.com/opentracing/opentracing-java/tree/master/opentracing-mock does not work. 

The activeSpan created from MockTracer is NoopActiveSpan so calling deactivate()/close() by try-resource block does not finish the MockSpan .

malafeev:
```
MockTracer tracer = new MockTracer();
``` 
should be replaced with 
```
MockTracer tracer = new MockTracer(new ThreadLocalActiveSpanSource())
```

mabn:
Could it become the default? I was also surprised that `MockTracer` by default "does not work".

pavolloffay:
+1 I have also seen people confused that MockTracer _does not work_ by default.

malafeev:
fixed in #176
can be closed now.

PR #175, opened 2017-09-07
add missing ThreadLocalActiveSpanSource #173

malafeev:


pavolloffay:
LGTM, I think we should make it default (it only confuses people why it does not work out of the box)

malafeev:
`ThreadLocalActiveSpanSource` is located in `opentracing-util`.
But `opentracing-util` depends on `opentracing-mock`.
Therefore adding dependency on `opentracing-util` introduces  cyclic reference.

pavolloffay:
I know, I am wondering if we can solve it, e.g. write tests in util without mock tracer

malafeev:
got it!

malafeev:
https://github.com/opentracing/opentracing-java/pull/176

malafeev:
because #176 is merged, this pull request is redundant

PR #176, opened 2017-09-07
set ThreadLocalActiveSpanSource default in MockTracer #173

malafeev:


malafeev:
@pavolloffay could review it?

pavolloffay:
LGTM, for 1.0 I would also change the default propagator to text map.

PR #179, opened 2017-09-11
MockTracer use text map propag as default constructor

### LINKED ISSUE #177: MockTracer use TextMap propagator by default
`MockTracer`, currently uses no/"printer" propagator by default. A better default is to use text map propagator.

pavolloffay:
Resolves #177

### DIFF
--- a/opentracing-mock/src/main/java/io/opentracing/mock/MockTracer.java
+++ b/opentracing-mock/src/main/java/io/opentracing/mock/MockTracer.java
@@ -13,22 +13,22 @@
  */
 package io.opentracing.mock;
 
-import io.opentracing.ActiveSpan;
-import io.opentracing.ActiveSpanSource;
-import io.opentracing.BaseSpan;
-import io.opentracing.noop.NoopActiveSpanSource;
-import io.opentracing.References;
-import io.opentracing.Span;
-import io.opentracing.SpanContext;
-import io.opentracing.Tracer;
-import io.opentracing.propagation.Format;
-import io.opentracing.propagation.TextMap;
-import io.opentracing.util.ThreadLocalActiveSpanSource;
-
 import java.util.ArrayList;
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
+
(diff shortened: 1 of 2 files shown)


opentracing-mock/src/main/java/io/opentracing/mock/MockTracer.java
```diff
@@ -39,16 +39,12 @@
...
+    private final List<MockSpan> finishedSpans = new ArrayList<>();
     private final Propagator propagator;
-    private ActiveSpanSource spanSource;
+    private final ActiveSpanSource spanSource;
 
     public MockTracer() {
-        this(new ThreadLocalActiveSpanSource(), Propagator.PRINTER);
-    }
-
-    public MockTracer(ActiveSpanSource spanSource) {
```
pavolloffay: removing this to keep the API clean.

pavolloffay:
If there are no issues I will merge in 48h.

cc @opentracing/otsc @opentracing/otiab

pavolloffay:
anybody from @opentracing/otsc to approve the PR?

carlosalberto:
@pavolloffay Was thinking if we should re-target this to 0.31 directly (as that's the new defacto 'upstream' branch). Let me know, as I could re-target it myself (so we can enjoy it for RC 2 ;) )

carlosalberto:
As a reference, I re-targeted this patch for v0.31.0 - and the result is here: https://github.com/pavolloffay/opentracing-java/commits/master. @pavolloffay hope this helps to get it integrated soon ;)

pavolloffay:
> Was thinking if we should re-target this to 0.31 directly (as that's the new defacto 'upstream' branch). Let me know, as I could re-target it myself (so we can enjoy it for RC 2 ;) )

@carlosalberto what about rebasing `v0.31` against the master before publishing RC2?

carlosalberto:
@pavolloffay That's actually a good idea. So, lets merge this change and to the rebase later today.

pavolloffay:
+1 IIR there are more commits in master than in `v0.31.`. I will merge

PR #184, opened 2017-09-17
Provide a new Tag for PEER_HOST_IPV4

wu-sheng:
I am creating a new pull request for #163. It is for #162 . And taken @yurishkuro 's suggestion.

### DIFF
--- a/opentracing-api/src/main/java/io/opentracing/tag/Tags.java
+++ b/opentracing-api/src/main/java/io/opentracing/tag/Tags.java
@@ -63,7 +63,7 @@ public final class Tags {
     /**
      * PEER_HOST_IPV4 records IPv4 host address of the peer.
      */
-    public static final IntTag PEER_HOST_IPV4 = new IntTag("peer.ipv4");
+    public static final IntOrStringTag PEER_HOST_IPV4 = new IntOrStringTag("peer.ipv4");
 
     /**
      * PEER_HOST_IPV6 records the IPv6 host address of the peer.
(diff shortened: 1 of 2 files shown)


opentracing-api/src/main/java/io/opentracing/tag/Tags.java
```diff
@@ -63,7 +63,7 @@ private Tags() {
     /**
      * PEER_HOST_IPV4 records IPv4 host address of the peer.
      */
-    public static final IntTag PEER_HOST_IPV4 = new IntTag("peer.ipv4");
+    public static final IntOrStringTag PEER_HOST_IPV4 = new IntOrStringTag("peer.ipv4");
```
pavolloffay: By doing this there is not a way how to deprecate `int` part. Spec says that it's `string` https://github.com/opentracing/specification/blob/master/semantic_conventions.md.

pavolloffay:
PR open-review-merge in 4 hours (on Sunday) does not leave any time for other people to look/review/comment.

wu-sheng:
@pavolloffay Because this is a pr last several months... and @yurishkuro think this is a bad time to break the API, and I agree him. 

So, by @yurishkuro suggestion (good one), I did in this pr. Origin pr is this: https://github.com/opentracing/opentracing-java/pull/163. The discussion ends 3 days ago. I delayed my work... because of releasing skywalking 3.2

I known the spec said string, but we have released so many versions. We must consider that.

PR #190, opened 2017-09-25
Code cleanup

mcumings:
- Propagate @Deprecated annotation to implementations
- Remove redundant "static final" definition from interface
- Use Collections.emptyMap instead of Collections.EMPTY_MAP to preserve type


opentracing-mock/src/main/java/io/opentracing/mock/MockTracer.java
```diff
@@ -85,6 +85,7 @@ public synchronized void reset() {
     /**
      * Noop method called on {@link Span#finish()}.
      */
+    @SuppressWarnings({"WeakerAccess", "unused"}) // API method
```
pavolloffay: How this produces a warning message? E.g. here https://travis-ci.org/opentracing/opentracing-java/builds/276509101

pavolloffay: Could we rather move it to class level and keep the sources clean?

mcumings: The WeakerAccess on `Propagator TEXT_MAP` could be pushed up to the anonymous inner class level without additional exposure.

Moving the `@SuppressWarnings({"WeakerAccess", "unused"})` on `onSpanFinished` up to the class level would suppress those conditions from far more code than required and is generally looked down upon as a code smell.

With respect to your first question, it looks like the opentracing-java library is not cleaning its build dir: ```[INFO] --- maven-compiler-plugin:3.6.1:compile (default-compile) @ opentracing-mock ---
[INFO] Nothing to compile - all classes are up to date
```
IDEs run this sort of static analysis all the time though.  Fixing this stuff prevents having to wade through a forest of warnings.

Comment (mcumings) [reply opentracing-mock/src/main/java/io/opentracing/mock/MockTracer.java]: WRT the Propagator impl - I was originally interpreting the public static vars as API but looking more into it it looks like that concept wouldn't make as much sense.  Is there a reason they are not `private static final`?

File: opentracing-mock/src/main/java/io/opentracing/mock/MockTracer.java [hunk 2]
```diff
@@ -258,6 +262,7 @@ public Scope startActive(boolean finishOnClose) {
         }
 
         @Override
+        @Deprecated
```
pavolloffay: +1 we should also use `inheritDoc` in some places

opentracing-mock/src/test/java/io/opentracing/mock/MockSpanTest.java
```diff
@@ -26,55 +26,59 @@
...
         MockTracer tracer = new MockTracer();
-        Span span = tracer.buildSpan("foo").start();
+        Span span = tracer.buildSpan("foo").startManual();
         span.finish();
 
         try {
             span.setOperationName("bar");
             Assert.fail();
         } catch (RuntimeException ex) {
+            // Fall through
```
sjoerdtalsma: If we're going to comment this, please comment _why_.
I can see it's a fall through. In this case it's a unit test and the exception is the expected behaviour (given the `Assert.fail()` above it).
I'd rather rename `ex` to `expected` and add something like `assertThat("Expected exception", expected is(notNullValue()));` for documentation instead of stating the obvious in a comment.

opentracing-mock/src/test/java/io/opentracing/mock/MockSpanTest.java
```diff
@@ -26,55 +26,59 @@
...
         }
         Assert.assertEquals(1, tracer.finishedSpans().get(0).generatedErrors().size());
     }
 
     @Test
     public void testSetTagAfterFinish() {
         MockTracer tracer = new MockTracer();
-        Span span = tracer.buildSpan("foo").start();
+        Span span = tracer.buildSpan("foo").startManual();
```
sjoerdtalsma: On all the `start()` -> `startManual()` changes: Shouldn't we leave at least some regression tests for the deprecated behaviour in place? Possibly in a new test class.

opentracing-mock/src/test/java/io/opentracing/mock/MockSpanTest.java
mcumings: I'd rather not test standard JVM behavior (i.e., that exceptions caught are not null).  I can change the comment to something more descriptive though.  Perhaps "Comment to prevent empty block warning" or some such?

opentracing-mock/src/test/java/io/opentracing/mock/MockSpanTest.java
mcumings: Sure. :+1:

opentracing-mock/src/main/java/io/opentracing/mock/MockTracer.java
mcumings: Ah, it's not worth it to me.  I just restored the warnings.

opentracing-mock/src/test/java/io/opentracing/mock/MockSpanTest.java
sjoerdtalsma: What I meant is the `assertThat` could be _after_ the try-catch and the catch could be a simple assignment. Then you can lose the nondescript `fail()` and the nondescript comment.
Something like:
```java
RuntimeException exception = null;
try {
    span.setOperationName("bar");
} catch (RuntimeException rte) {
    exception = rte;
}
assertThat("Exception setting after finish", exception, notNullValue());
```

But in the end that's just a personal preference, so I'm okay either way.

mcumings:
@pavolloffay - Where do you stand with respect to this PR?  Merge worthy?

opentracing-mock/src/test/java/io/opentracing/mock/MockSpanDeprecationsTest.java
```diff
@@ -0,0 +1,39 @@
...
+import static org.junit.Assert.assertTrue;
+
+/**
+ * Tests against deprecated behavior and/or API.
+ */
+@SuppressWarnings("deprecation")
+public class MockSpanDeprecationsTest {
+
+    @Test
+    public void testOldStartMethodWtihFinishMicros() throws Exception {
```
pavolloffay: The similar test is in `MockTracerTest`.

opentracing-mock/src/test/java/io/opentracing/mock/MockSpanTest.java
```diff
@@ -26,55 +26,59 @@
...
         span.finish();
 
         try {
             span.setOperationName("bar");
             Assert.fail();
         } catch (RuntimeException ex) {
+            // Comment to prevent empty block warning
```
pavolloffay: Or you can assign to local var and test equals in the next assert statement.

opentracing-mock/src/test/java/io/opentracing/mock/MockTracerTest.java
```diff
@@ -128,7 +128,7 @@ public void testStartExplicitTimestamp() throws InterruptedException {
         {
             tracer.buildSpan("foo")
                     .withStartTimestamp(startMicros)
-                    .start()
+                    .startManual()
```
pavolloffay: btw. `startManual` will be probably removed in 1.0/0.31.0 I would just wait with this.

opentracing-mock/src/main/java/io/opentracing/mock/MockTracer.java
pavolloffay: It helps with _propagation_ tests where you want to be sure that span context is present in the carrier.

opentracing-mock/src/test/java/io/opentracing/mock/MockSpanTest.java
sjoerdtalsma: Kind of what I meant in [my comment](https://github.com/opentracing/opentracing-java/pull/190#discussion_r141122680)

mcumings:
Okay, I've reduced this PR to the small set of things that everyone immediately agrees with.

mcumings:
And it looks like travis is still not picking this up so I'll squash and push a new PR.

mcumings:
Closing in favor of a fresh PR with hopeful Travis run: https://github.com/opentracing/opentracing-java/pull/194

mcumings:
Retargeted to v0.31.0 branch: https://github.com/opentracing/opentracing-java/pull/196

PR #195, opened 2017-09-27
Update the Travis script to allow publishing from v.0.0.0 branches.

carlosalberto:
Thing is, we cannot publish from 0.0.0-style branches as they are
excluded, based on the current global Travis configuration, thus
we try to publish from branches following a v.0.0.0 style, if any.

### DIFF
--- a/travis/publish.sh
+++ b/travis/publish.sh
@@ -36,11 +36,11 @@ is_pull_request() {
 }
 
 is_travis_branch_master_or_release() {
-  if [[ "${TRAVIS_BRANCH}" == "master" || "${TRAVIS_BRANCH}" =~ ^[0-9]+\.[0-9]+\.[0-9]+$ ]]; then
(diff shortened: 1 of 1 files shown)


carlosalberto:
Hey Pavol,

Managed to get an actual patch after going through the slightly intricate workflow for publishing. Let me know if this patch makes sense to you.

carlosalberto:
@pavolloffay Going to merge then, and will issue a `release-` branch. Thanks!

pavolloffay:
do not forget to change the branch :P

adriancole:
I think there was no tag for 0.31.0-RC1, maybe it was for this commit? If so, can you please tag so that git diff between tags can be made?  http://search.maven.org/#search%7Cgav%7C1%7Cg%3A%22io.opentracing%22%20AND%20a%3A%22opentracing-api%22

carlosalberto:
Hey @adriancole thanks for pointing this out*. Would you be fine with a manually (annotated) tag that resembles the one automatically created by the deploy plugin? If yes, I can go ahead and do it myself. Let me know.

(Else we would need to do some juggling here and there ;) )

*I remember we had a slight problem releasing from a branch other than `master`, and thus we ended up without this tag.

PR #199, opened 2017-09-29
Code cleanup (for master)

mcumings:
- Propagate @deprecated annotation to implementations
- Remove redundant "static final" definition from interface
- Use Collections.emptyMap instead of Collections.EMPTY_MAP to preserve type


mcumings:
Okay thanks.  How do I get this merged now that it is approved?

pavolloffay:
I will merge right away, it's only consmetic change not changing API.

@mcumings thanks

pavolloffay:
@mcumings thanks

PR #200, opened 2017-10-02
Idiomatic Java timestamps (#180)

mcumings:
Adds the specification of TimeUnit to the API in places which
previously required supplying values in microseconds.

### DIFF
--- a/opentracing-mock/src/test/java/io/opentracing/mock/MockSpanTest.java
+++ b/opentracing-mock/src/test/java/io/opentracing/mock/MockSpanTest.java
@@ -13,15 +13,76 @@
  */
 package io.opentracing.mock;
 
+import io.opentracing.Span;
 import org.junit.Assert;
 import org.junit.Test;
 
-import io.opentracing.Span;
+import java.util.HashMap;
+import java.util.Map;
+import java.util.concurrent.TimeUnit;
 
 /**
  * @author Pavol Loffay
  */
 public class MockSpanTest {
+
+    @Test
+    @SuppressWarnings("deprecation")
+    public void testExplicitStartInMicroseconds() {
+        MockTracer tracer = new MockTracer();
+        long now = System.currentTimeMillis() * 1000;
+        Span span = tracer.buildSpan("foo")
+                .withStartTimestamp(now)
+                .startManual();
+        span.finish();
+
+        Assert.assertEquals(now, tracer.finishedSpans().get(0).startMicros());
+    }
+
+    @Test
+    public void testExplicitStart() {
+        MockTracer tracer = new MockTracer();
+        long now = System.currentTimeMillis();
+        Span span = tracer.buildSpan("foo")
+                .withStartTimestamp(now, TimeUnit.MILLISECONDS)
+                .startManual();
+        span.finish();
+
+        Assert.assertEquals(now, tracer.finishedSpans().get(0).startTimestamp(TimeUnit.MILLISECONDS));
+    }
+
+    @Test
+    @SuppressWarnings("deprecation")
+    public void testExplicitFinishInMicroseconds() {
+        MockTracer tracer = new MockTracer();
+        Span span = tracer.buildSpan("foo").startManual();
+        long finishTime = 133L;
+        span.finish(finishTime);
+
+        try {
+            span.setOperationName("bar");
+            Assert.fail();
+        } catch (RuntimeException ex) {
+            // Fall through
+        }
+        Assert.assertEquals(finishTime, tracer.finishedSpans().get(0).finishMicros());
+    }
+
+    @Test
+    public void testExplicitFinish() {
+        MockTracer tracer = new MockTracer();
+        Span span = tracer.buildSpan("foo").startManual();
+        long finishTime = System.currentTimeMillis();
+        span.finish(finishTime, TimeUnit.MILLISECONDS);
+
+        try {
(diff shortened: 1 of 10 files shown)


pavolloffay:
This PR has been reviewed here https://github.com/opentracing/opentracing-java/pull/181.

I think this should be opened against the master branch. It's better to keep `v0.31.0` dedicated only for span management changes.

opentracing-mock/src/test/java/io/opentracing/mock/MockSpanTest.java
```diff
@@ -13,17 +13,78 @@
  */
 package io.opentracing.mock;
 
+import io.opentracing.Span;
```
carlosalberto: Any reason to reorganize the `Span` imports?

carlosalberto:
These changes look very nice - I'm only wondering personally on the decision to make the current overloads not using `TimeUnit` as `Deprecated` - is it maybe of the desire of having a single, unified method, or because of the dislike for `micros` (opposed to `millis`)?

yurishkuro:
The micros methods become redundant, so we need to deprecated them to be able to remove them in the future.

mcumings:
@pavolloffay - I was asked to merge this to v0.31.0 instead of master, hence this PR.  I think the intent is to get the change in front of folks early via the RC builds.

carlosalberto:
Correct - *as far* as I had understood, we want all changes/improvements to come to this branch, and eventually merge it into `master` (when we release the stable 0.32.0 version). So *in theory* only bug fixes should go into master too.

tedsuo:
It makes sense that new API changes should go onto the 0.31.0, as that represents the unstable rather than master, to prevent tracer implementations from having to make a cascading series of changes. I'm mentioning this on Gitter; let's have the debate there where it's a little more visible.

pavolloffay:
-1 from my side to use `TimeUnit` as a unit for timestamps.

* `TimeUnit` is from `java.util.concurrent` and it's used to specify **timeouts**. The API allows something like this `TimeUnit.SECONDS.sleep(1);`
* I haven't seen using this API for unix timestamps any java code.
* `TimeUnit` contains, days, hours, minutes, seconds, millis, micro, nano. For timestamp it only makes sense to use millis and/or more precise units.
* there is no way how to combine timeunits e.g. if you choose days you cannot add hours minutes, so in the end you have to use the lowest possible unit. Original comment https://github.com/opentracing/opentracing-java/pull/181#issuecomment-334143918

For me what makes sense is to provide milliseconds variants. Milliseconds are the most common units for timestamps in Java.

mcumings:
Can we please keep the PR comments targeted to the proposed changes themselves and the higher level discussion in the associated issue?

adriancole:
hats off to @pavolloffay for thinking this through aloud! speaking up on pavol's behalf :)

This change is actually harder to convert to other java idioms such as instant (jre 8+ version of timestamp which is a pair epoch seconds, nanos offset). Other options include using a double as AWS use, where the whole number is seconds iirc. whatever change should be better than before and enumerate what's being solved as otherwise it is bothering people with busywork.

For example, these apis are usually used for recording after-the-fact work, such as when a tracer is being used as a span recorder into another system. In that system, they will know the precision they can accept, but that isn't necessarily a good reason to use TimeUnit to achieve that. IMHO, the apis with timestamps are rarely a good or common user api anyway (in hindsight better as a separate after-the-fact recording api, which needs no context management for example)

If the goal is to use nanos precision (should you have that available and it isn't configurable on the tracer itself), then that should be listed as a key driver, and yeah there are other choices to achieve that.

If only to stop doing multiplication or division, this change doesn't accomplish that.

mcumings:
So what I'm gathering is that there is a definite preference to JDK 8 semantics (to which I agree) that we can't directly use and that folks would want to limit any future work to a global search and replace in their codebase.

There is an underlying requirement that the precision should not be chosen by the API directly as doing this forces a reduction in precision via division or a false promotion of precision via multiplication.  Both of these choices have a runtime cost.  The exception being when the provision and consumption are done at the same level of precision in which case the cost is a no-op.  It is this exceptional case which *is* handled by the TimeUnit change, @adriancole.

The only way that I am aware of to solve the above is to create a dedicated clock API.  Doing so would likely be more effort than it is worth if there is any hope of moving to Java 8 language level anytime soon.

Unless anyone else has any other new ideas then I'm guessing we'll choose to be stuck with direct microseconds specification - arguably the worst possible choice for Java.

tylerbenson:
I agree that the current implementation is rather unfortunate, and would love to see something better.  That being said, I think it would be good for us to consider @adriancole's suggestion of removing these "after-the-fact" recording API's from the main API.  (Correcting decisions from the API seems to be the theme of the current RC.)

@mcumings do you think that would solve the problem well?  Maybe this PR could be updated to only deprecate these problematic methods for future removal and consider a replacement at a later date.

mcumings:
@adriancole / @tylerbenson - For me the proposal to eliminate the API altogether would be problematic.  Doing so assumes that there is no other actor in the system which is providing timestamped data of interest that would be attached to the tracing data.  i.e., it is the only performance tracking system in the effective ecosystem.  My app has a few, focused, such subsystems and I have to apply an adapter to pull the data into the OpenTracing ecosystem.

If it helps, the uses of the above are currently limited to `Span#log(...)` and do not extend to Span lifecycle.

PR #202, opened 2017-10-05
Add MDC support for basic ThreadLocalActiveSpan

### LINKED ISSUE #201: Include MDC in default ActiveSpanSource
Why not include SLF4J concept of MDC in the default implementation of ActiveSpanSource?

fzakaria:
The documentation heavily mentions MDC in the interfaces but fails to
provide an example in the basic implementations of
ThreadLocalActiveSpan.

This commit adds this functionality via SLF4J which is a very popular
facade to many common logging libraries. As a result, a new dependency
is introduced on slf4j-api however I believe this to be a useful
dependency so that logging can be applied in a runtime agnostic fashion
if additional example implementations would like to be provided.

Fixes #201

### DIFF
--- a/opentracing-util/pom.xml
+++ b/opentracing-util/pom.xml
@@ -34,6 +34,10 @@
 
     <dependencies>
         <dependency>
+            <groupId>org.slf4j</groupId>
+            <artifactId>slf4j-api</artifactId>
+        </dependency>
+        <dependency>
             <groupId>${project.groupId}</groupId>
             <artifactId>opentracing-api</artifactId>
         </dependency>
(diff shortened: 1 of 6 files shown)


yurishkuro:
The idea was that if someone wanted MDC propagation they would extend/reimplement the active span source and provide it to the tracer. I don't think building this functionality directly into the main API is the right approach.

tylerbenson:
I agree with @yurishkuro that this doesn't belong here.  That being said, it might be useful to transfer this idea/concept/work over to a new project that focuses on a simple tracer implementation that only maintains the MDC state.

adriancole:
whereever you go with this, you should add that the following properties can be visible to MDC, else you will end up changing later to do the same:

traceId, parentId, spanId, sampled

Issue #168, opened 2017-08-10
Shall we allow nullable return value when no SpanContext extracted?


tutufool:
https://github.com/opentracing/opentracing-java/blob/578d4e74d9b16c18268717e537000a88d7e6779c/opentracing-api/src/main/java/io/opentracing/Tracer.java#L95

You may check this issue for details:

https://github.com/openzipkin/brave-opentracing/issues/49

Thanks
Leon

yurishkuro:
Documenting `null` has my vote. I proposed it [a year ago](https://github.com/opentracing/opentracing-java/pull/43/files#diff-9f484babc6e66bd62198ce009d9247dcR79), which lead to a heated debate and didn't materialize, even though it is the de-facto behavior in a number of tracers today anyway.

cc @adriancole @bhs @pavolloffay @basvanbeek

pavolloffay:
+1

bhs:
Also +1... not something I feel adamant about, but `null` is a natural way to model "not present". The only counterargument that holds water for me would be protection against NPEs... which there's merit there, it would add marker types/instances and complexity to follow suit which is not worth it IMO.

mabn:
It would be good to annotate methods that can return null with `@Nullable` as it allows IDE to complain if returns type is not checked.

yurishkuro:
the issue we previously had with annotations is that they are not supported in JRE 1.6

mabn:
What do you mean? `@Nullable` is not present in any Java version, this needs to be taken from a dependency. E.g. guava depends on [com.google.code.findbugs:jsr305](https://github.com/google/guava/blob/6ded67ff124f1be4f318dbbeae136d0c995faf37/guava/pom.xml#L13-L26) but it's an optional dependency. Annotations are supported since Java 1.5.

yurishkuro:
what happens if someone tries to use an API that has annotations from a lib that is not on the classpath?

mabn:
Annotations have several retention policies - `@Nullable` has `RUNTIME` which means that it's visible at runtime.

It is references in the class file:
```
@Nullable
public String a() {
  return null;
}


public class Main
  minor version: 0
  major version: 52
  flags: ACC_PUBLIC, ACC_SUPER
Constant pool:
(...)
  #49 = Utf8               Ljavax/annotation/Nullable;
(...)
  public java.lang.String a();
    descriptor: ()Ljava/lang/String;
    flags: ACC_PUBLIC
(...)
[...]

PR #203, opened 2017-10-06
Clarify that Tracer.extract can return null

mabn:
According to https://github.com/opentracing/opentracing-java/issues/168 in practice the tracer implementations return null when `SpanContext` is not found in the carrier. This change documents this behaviour.

It also adds a dependency on `com.google.code.findbugs:jsr305` but it's optional, it is not listed in the pom and client applications do not need to depend on jsr305. 

Alt text

Compiling the code above with:
```
javac -Werror -Xlint:all -cp \
/Users/mabn/.m2/repository/io/opentracing/opentracing-api/0.30.1-SNAPSHOT/opentracing-api-0.30.1-SNAPSHOT.jar:\
/Users/mabn/.m2/repository/io/opentracing/opentracing-noop/0.30.1-SNAPSHOT/opentracing-noop-0.30.1-SNAPSHOT.jar:\
/Users/mabn/.m2/repository/io/opentracing/opentracing-util/0.30.1-SNAPSHOT/opentracing-util-0.30.1-SNAPSHOT.jar \
Main.java
```
procudes no warnings.

### DIFF
--- a/opentracing-api/pom.xml
+++ b/opentracing-api/pom.xml
@@ -31,6 +31,14 @@
         <main.basedir>${project.basedir}/..</main.basedir>
         <main.java.version>1.6</main.java.version>
         <main.signature.artifact>java16</main.signature.artifact>
+        <findbugs.jsr305.version>3.0.2</findbugs.jsr305.version>
     </properties>
 
+    <dependencies>
+        <dependency>
+            <groupId>com.google.code.findbugs</groupId>
+            <artifactId>jsr305</artifactId>
+            <scope>provided</scope>
+        </dependency>
+    </dependencies>
 </project>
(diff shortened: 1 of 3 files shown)


opentracing-api/pom.xml
```diff
@@ -31,6 +31,14 @@
         <main.basedir>${project.basedir}/..</main.basedir>
         <main.java.version>1.6</main.java.version>
         <main.signature.artifact>java16</main.signature.artifact>
+        <findbugs.jsr305.version>3.0.2</findbugs.jsr305.version>
     </properties>
 
+    <dependencies>
+        <dependency>
+            <groupId>com.google.code.findbugs</groupId>
```
pavolloffay: Can we do this in a separate PR? It might apply also at different places

mabn: Do you mean to split adding jsr305 and marking `Tracer.extract` into two separate PRs?

mabn:
I'll close this PR as I haven't looked into this for a while.

whiskeysierra:
I'd still like to see this happening, tbh.

PR #197, opened 2017-09-28
Add Examples for async use cases

tylerbenson:
See #183 for further details and discussion.

PR #205, opened 2017-10-17
Add a new example for Fire-and-Forget style messages.

carlosalberto:
New example based on Actor System style messages, where messages can be of fire-and-forget style.


tylerbenson:
This is already a case that is covered in #197.

carlosalberto:
It's pretty similar, yes. Maybe we should move it to an additional test case

carlosalberto:
Closing this PR since, as Tyler pointed out, this is a very similar case, and even though it spots some differences, it's probably not as important/needed.

PR #207, opened 2017-10-19
Debug ID in MDC

mabn:
An example of keeping Span's debug_id inside MDC - for version 0.31.0+

Puts active Spans's debug_id when scope is activated and cleans it
afterwards.

see also: https://github.com/opentracing/opentracing-java/issues/206

### DIFF
--- a/opentracing-examples/src/test/java/io/opentracing/examples/slf4j_mdc_debug_id/SimpleMdcScopeManager.java
+++ b/opentracing-examples/src/test/java/io/opentracing/examples/slf4j_mdc_debug_id/SimpleMdcScopeManager.java
@@ -33,9 +33,9 @@ public class SimpleMdcScopeManager implements ScopeManager {
     }
 
     /**
-     * @param mdcKey        MDC key which value will be set to debug_id
+     * @param mdcKey  MDC key which value will be set to debug_id
      * @param debugId debug_id provider - specific to a Tracer implementation
-     * @param wrapped       ScopeManager to wrap
+     * @param wrapped ScopeManager to wrap
      */
     public SimpleMdcScopeManager(String mdcKey, DebugIdProvider debugId, ScopeManager wrapped) {
         this.mdcKey = mdcKey;
--- a/opentracing-examples/src/test/java/io/opentracing/examples/slf4j_mdc_debug_id/SimpleMdcScopeManager.java
+++ b/opentracing-examples/src/test/java/io/opentracing/examples/slf4j_mdc_debug_id/SimpleMdcScopeManager.java
@@ -1,3 +1,16 @@
+/*
+ * Copyright 2016-2017 The OpenTracing Authors
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License"); you may not use this file except
+ * in compliance with the License. You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software distributed under the License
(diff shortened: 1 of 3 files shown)


opentracing-examples/src/test/java/io/opentracing/examples/slf4j_mdc_debug_id/SimpleMdcScopeManagerTest.java
```diff
@@ -0,0 +1,102 @@
...
+
+    private static String getDebugId(Span span) {
+        MockSpan.MockContext ctx = ((MockSpan) span).context();
+        return ctx.traceId() + ":" + ctx.spanId();
+    }
+
+    private static class MockSpanDebugId implements SimpleMdcScopeManager.DebugIdProvider {
+        @Override
+        public String get(Span span) {
+            MockSpan.MockContext ctx = ((MockSpan) span).context();
```
mabn: Real implementation will also require casting so other libs that wrap Span will make it fail :|

opentracing-examples/src/test/java/io/opentracing/examples/slf4j_mdc_debug_id/SimpleMdcScopeManager.java
```diff
@@ -0,0 +1,87 @@
...
+
+    @Override
+    public Scope activate(Span span) {
+        return activate(span, true);
+    }
+
+    @Override
+    public Scope activate(Span span, boolean finishSpanOnClose) {
+        Scope scope = wrapped.activate(span, finishSpanOnClose);
+        String debugId = MDC.get(mdcKey);
```
carlosalberto: Maybe call it something other than `debugId` so you don't have to use `this.debugId` for the instance field? (same name but different types, etc)

carlosalberto:
Out of curiosity: is this something you would foresee being implemented as a wrapper class? I ask because if the user does `scopeManager.active().close()`, the logic you used for `MDC` will not be called, i.e.

```
            } finally {
                 MDC.put(mdcKey, previousDebugId);
            }
```

mabn:
Oh, you're right. I should pass my `MdcScope` when calling `wrapped.activate(..)`

If this wouldn't be a wrapper then it wouldn't be possible to use it with another scope manager (e.g. `AutoFinishScopeManager`). Until there is an API to register observers/listeners/extensions then wrapping is the only way :|

sudr:
Wondering when this feature will become available.

whiskeysierra:
@mabn Isn't this what https://github.com/opentracing-contrib/java-api-extensions is for?

yurishkuro:
I haven't tried api-extensions, but subclassed the scope manager to add trace IDs to logs.

https://github.com/PacktPublishing/Mastering-Distributed-Tracing/blob/master/Chapter11/exercise1/lib/src/main/java/lib/MDCScopeManager.java

whiskeysierra:
As long as 0.32.0 is not released there is no way to get this done without being implementation-specific, right?

https://github.com/PacktPublishing/Mastering-Distributed-Tracing/blob/master/Chapter11/exercise1/lib/src/main/java/lib/MDCScopeManager.java#L29

yurishkuro:
@whiskeysierra yeah. Even with 0.32, the `traceSampled` field cannot be logged generically.

mabn:
@whiskeysierra yes, java-api-extensions would be suitable, but IIRC at the time they didn't have some of the callbacks I needed

I'm closing this PR as I haven't worked on this for quite a while.

Issue #170, opened 2017-08-27
GlobalTracer breaks tests because it can be initialized only once


mabn:
`GlobalTracer.register` throws an exception when called again with another tracer instance. It makes unit tests fail if they happen to initialize GlobalTracer.

Has anyone run into this problem? I could run tests in separate JVM instances, but I'd rather not. Maybe adding a "reset" method would make it easier to use for testing purposes?

malafeev:
I solve this problem only with `maven-surefire-plugin`:
```
 <plugin>
    <groupId>org.apache.maven.plugins</groupId>
    <artifactId>maven-surefire-plugin</artifactId>
    <version>2.20</version>
    <configuration>
       <reuseForks>false</reuseForks>
       <forkCount>1</forkCount
    </configuration>
</plugin>
```

tylerbenson:
I solve this problem by using reflection to set the tracer field in `GlobalTracer` back to a `NoopTracer`.  This would probably be considered a hack, but is more efficient than starting a new JVM for every test.

pavolloffay:
Reflection is also used in this repo 
https://github.com/opentracing/opentracing-java/blob/578d4e74d9b16c18268717e537000a88d7e6779c/opentracing-util/src/test/java/io/opentracing/util/GlobalTracerTest.java#L48_L57. I agree it's a little bit inconvenient.

mabn:
I also went with reflection as I'd prefer not to spawn new JVM for each test case - it is especially expensive when starting Spring context - it can be reused for multiple tests if there's one JVM, but forking prevents that.

sjoerdtalsma:
Maybe we could share the reflection utility code in a maven [`test-jar`](https://maven.apache.org/guides/mini/guide-attached-tests.html) for opentracing-util ?

That would save people from reinventing this hacky wheel, while not exposing unwanted behaviour in the main api.

pavolloffay:
I think we should rather provide `reset` method directly on the `GlobalTracer`. It will make things much simpler, consumers should understand implications when calling it.

sjoerdtalsma:
@pavolloffay Do you know of a use-case where this would be called in production code?

pavolloffay:
Currently, I cannot think of any, but it seems something wrong with API design if we have to deal with reflection to make things work.

sjoerdtalsma:
[...]

PR #209, opened 2017-10-26
Globaltracer testutil

sjoerdtalsma:
This may help people needing to reset the GlobalTracer (see #170).


sjoerdtalsma:
Are there any objections against merging this?

PR #213, opened 2017-11-05
Fixes #210 : add support for multi reference and expose references co…

### LINKED ISSUE #210: MockTracer support multiple parents and reference types
At the moment `MockTracer` supports only one parent (`childOf` or `followsFrom`) but it's not storing reference type.

hypnoce:
…ntext and type from MockSpan.

### DIFF
--- a/opentracing-mock/src/main/java/io/opentracing/mock/MockSpan.java
+++ b/opentracing-mock/src/main/java/io/opentracing/mock/MockSpan.java
@@ -13,11 +13,13 @@
  */
 package io.opentracing.mock;
 
+import io.opentracing.References;
 import java.util.ArrayList;
 import java.util.Collections;
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
+import java.util.Objects;
 import java.util.concurrent.atomic.AtomicLong;
 
 import io.opentracing.Span;
@@ -42,6 +44,7 @@ public final class MockSpan implements Span {
     private final Map<String, Object> tags;
     private final List<LogEntry> logEntries = new ArrayList<>();
     private String operationName;
+    private final List<Reference> references;
 
     private final List<RuntimeException> errors = new ArrayList<>();
 
@@ -57,11 +60,10 @@ public final class MockSpan implements Span {
     }
 
     /**
-     * TODO: Support multiple parents in this API.
-     *
-     * @return the spanId of the Span's parent context, or 0 if no such parent exists.
+     * @return the spanId of the Span's first {@value References#CHILD_OF} reference, or the first reference of any type, or 0 if no reference exists.
      *
      * @see MockContext#spanId()
+     * @see MockSpan#references()
      */
     public long parentId() {
         return parentId;
@@ -95,6 +97,10 @@ public final class MockSpan implements Span {
      */
     public List<RuntimeException> generatedErrors() {
         return new ArrayList<>(errors);
+    }
+
+    public List<Reference> references() {
+        return new ArrayList<>(references);
     }
 
     @Override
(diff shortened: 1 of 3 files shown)


opentracing-mock/src/main/java/io/opentracing/mock/MockSpan.java
```diff
@@ -13,11 +13,13 @@
  */
 package io.opentracing.mock;
 
+import io.opentracing.References;
 import java.util.ArrayList;
 import java.util.Collections;
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
+import java.util.Objects;
```
yurishkuro: isn't this only available from 1.7+?

hypnoce: Yes. I thought the compilation source and target were 1.7 since it is set in the pom file.

yurishkuro: ah, ok - the api module is 1.6, the rest are 1.7, so no new change is introduced here

opentracing-mock/src/main/java/io/opentracing/mock/MockSpan.java
```diff
@@ -241,17 +280,49 @@ public long timestampMicros() {
...
     }
 
+    private static MockContext findPreferredParentRef(List<Reference> references) {
+        if(references.isEmpty()) {
+            return null;
+        }
+        Reference preferredReference = references.get(0);
+        for (Reference reference : references) {
+            // childOf takes precedence as a preferred parent
+            if (References.CHILD_OF.equals(reference.getReferenceType())
```
yurishkuro: why not just return immediately in this case?

hypnoce: Indeed, I don't have a strong opinion on that. The control flow is the same. I find it more readable to have a single return at the end but I will change this.

yurishkuro: what I meant was this:
```
if list.isEmpty { return nil; }
for r in list {
  if r.isChildOf { return r.context() }
}
return list.get(0).context();
```
faster and simpler

yurishkuro:
lgtm

cc @pavolloffay

opentracing-mock/src/main/java/io/opentracing/mock/MockTracer.java
```diff
@@ -231,9 +231,8 @@ public SpanBuilder ignoreActiveSpan() {
 
         @Override
         public SpanBuilder addReference(String referenceType, SpanContext referencedContext) {
-            if (firstParent == null && (
-                    referenceType.equals(References.CHILD_OF) || referenceType.equals(References.FOLLOWS_FROM))) {
-                this.firstParent = (MockSpan.MockContext)referencedContext;
+            if (referenceType.equals(References.CHILD_OF) || referenceType.equals(References.FOLLOWS_FROM)) {
```
pavolloffay: Why this check? It can support arbitrary references.

pavolloffay:
LGTM, 

Java doc in `MockSpan.parentId()` has to be changed: remove todo, make a note that it returns the first `childOf` reference and add link to `MockSpan.references()`.

opentracing-mock/src/main/java/io/opentracing/mock/MockTracer.java
hypnoce: Indeed, I just kept the current behaviour. I will change this.

hypnoce:
@pavolloffay I made the required changes. Let me know if that looks good to you ! Thanks !

PR #215, opened 2017-11-10
Move our old ActiveSpan related APIs to sub v_030 packages.

carlosalberto:
In order to keep backwards compatibility with the 0.30 version, the ActiveSpan API is kept under a v_030 su package (for each one of our modules).

SpanContext, as well as propagation and tags packages are kept in place, as they did not change between 0.30 and 0.31.

For your consideration. Feedback is appreciated.

### DIFF
--- a/opentracing-api/src/main/java/io/opentracing/v_030/ActiveSpan.java
+++ b/opentracing-api/src/main/java/io/opentracing/v_030/ActiveSpan.java
@@ -1,0 +1,111 @@
+/*
+ * Copyright 2016-2017 The OpenTracing Authors
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License"); you may not use this file except
+ * in compliance with the License. You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software distributed under the License
+ * is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express
+ * or implied. See the License for the specific language governing permissions and limitations under
+ * the License.
+ */
+package io.opentracing.v_030;
+
+import io.opentracing.SpanContext;
+
+import java.io.Closeable;
+
+/**
+ * {@link ActiveSpan} inherits all of the OpenTracing functionality in {@link BaseSpan} and layers on in-process
+ * propagation capabilities.
+ *
+ * <p>
+ * In any given thread there is at most one {@link ActiveSpan "active" span} primarily responsible for the work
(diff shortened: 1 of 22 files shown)


tedsuo:
Thanks @carlosalberto!

- I believe this work is to support adding a shim which would allow a v0.31 Tracer to be wrapped in a v0.30 API, so users can consume updated tracers without tracer implementors needing to maintain to APIs. The shim is not included here, but when it is, can we have both GlobalTracer singletons point to the same instrance? That would allow users to gradually migrate their instrumentation, which would be helpful.
- In terms of users doing find/replace/refactor to change their import paths over to `v_030`, would it be easier for them if we had `io.opentracing.v_030` versions of SpanContext and everything else? So that find/replace does not need to be so selective.
- Pedantic: is there any reason to prefer `io.opentracing.mock.v_030` vs `io.opentracing.v_030.mock` as the pattern? This is almost the definition of a bikeshed, both are fine to me.

tylerbenson:
I'm not really sure how this helps us.  What will actually be using this code? Can a tracer still be compatible with both apis concurrently?

carlosalberto:
Hey @tedsuo, thanks for the feedback!

Indeed, the idea is to support a Shim (see #216 ) and let users use a 0.31 Tracer in libraries still using 0.30.

* GlobalTracer will be indeed updated (for now, in #216 is simply a duplicated, which stands as a proof-of-concept).
* About having their own versions of `SpanContext` and other classes, yes, I had considered it and I think we could make it possible (by simply extending the original class, instead of repeating such code). Agree that this could help users to keep things 'clear'.
* About the `v_030` position in the package naming: I had as well considered that one, but my personal preference simply was on the current approach, but definitely open to change it to the other way if needed (`io.opentracing.v_030.util` and similar, that is).

carlosalberto:
Hey @tylerbenson!

By using a Shim layer (like the one showcased as a proof of concept in #216 ), we could let Tracer authors provide backwards compatibility with libraries still using the 0.30 API (maybe even using the two of them at some point, as part of an incremental porting approach, by having the 'old' GlobalTracer point to the new one). And part of that would be, precisely, to keep this old API around.

yurishkuro:
I kinda question the value of the shim approach since it still requires changing the namespaces in all of the code that uses the API. Is it that different from changing the API calls themselves?

tedsuo:
It's a much more mechanical refactor to change the import path, rather that the instrumentation. Especially because some instrumentation may require rethinking.

If there is an even lower touch approach, we should explore that as well. The advantage with a separate namespace is that allows both API versions to interop cleanly, and we can provide a gradual migration strategy for users with large codebases.

carlosalberto:
Closing this one on behalf of #221

PR #216, opened 2017-11-13
0.30 shim (proof of concept)

carlosalberto:
This is a proof-of-concept idea for providing a shim layer for the new `Scope`s API in 0.31 to support the old `ActiveSpan` API from 0.30. This would let `Tracer` authors to migrate their implementations to use the new API, and still support instrumentation code that uses the old paradigm. Needs polishing and tests, but should give an idea overall.

The approach consists of:

* `io.opentracing.util.v_030.TracerShim` and related classes, to wrap 0.31 Tracer instances, as well as their related objects (no `ActiveSpan.capture()` support).
* `io.opentracing.util.v_030.AutoFinishTracerShim` to wrap 0.31 Tracer instances using the special `AutoFinishScopeManager`, which keeps around the Continuation concept to support ref-counting (which has been left out of the main API).
* `io.opentracing.util.AutoFinishScopeManager`: Scope manager supporting Continuation objects.

Things I still don't have clear yet:

1. Should we be fine by exposing a `AutoFinishScopeManager`? On one hand, it helps this layer to keep backwards compatibility, and may as well be used in Scala's asynchronous frameworks, but there's always the risk it may be overused.
2. Forcing the usage of `AutoFinishScopeManager` for a given `Tracer` to use the shim supporting `Continuation`s: feels like not friendly enough, but at the same time, such ref-count or capturing seems to be hard to have at the shim layer itself, and so much easier at the `ScopeManager` support instead.
3. Maybe support `Continuation`s all the way, and removing the shim object that doesn't support them at all?

Any feedback is welcome!

### DIFF
--- a/opentracing-api/src/main/java/io/opentracing/v_030/ActiveSpan.java
+++ b/opentracing-api/src/main/java/io/opentracing/v_030/ActiveSpan.java
@@ -1,0 +1,111 @@
+/*
+ * Copyright 2016-2017 The OpenTracing Authors
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License"); you may not use this file except
+ * in compliance with the License. You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software distributed under the License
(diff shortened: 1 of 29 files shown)


carlosalberto:
Closing this one on behalf of #221

PR #221, opened 2017-11-21
Add a 0.30 compatibility package/layer

carlosalberto:
This is an updated PR for the compatibility layer for the 0.30 API.

The changes are mostly:

* Put all the 0.30 packages under a new `opentracing-v030` package (so not mess with the existing code). For keeping things uniform, the code from 0.30 has been mostly copied over.

* Have a `io.opentracing.v_030.shim.TracerShim` class to wrap `io.opentracing.Tracer` instances and expose them under the 0.30 API. This required the addition of `AutoFinishScopeManager`, which is a `ScopeManager` supporting reference count. I decided to put it under `util` as it may be used in other components later on, and not only the Shim (like in Scala asynchronous frameworks).

Any feedback is appreciated ;) Observe the idea is to let instrumentation authors to incrementally support the new API while keeping backwards compatibility against the old one.

### DIFF
--- a/opentracing-v030/src/main/java/io/opentracing/v_030/shim/BaseTracerShim.java
+++ b/opentracing-v030/src/main/java/io/opentracing/v_030/shim/BaseTracerShim.java
@@ -1,0 +1,163 @@
+/*
+ * Copyright 2016-2017 The OpenTracing Authors
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License"); you may not use this file except
+ * in compliance with the License. You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software distributed under the License
+ * is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express
+ * or implied. See the License for the specific language governing permissions and limitations under
+ * the License.
+ */
+package io.opentracing.v_030.shim;
+
+import io.opentracing.ScopeManager;
+import io.opentracing.Scope;
+import io.opentracing.v_030.ActiveSpan;
+import io.opentracing.v_030.BaseSpan;
+import io.opentracing.v_030.Span;
+import io.opentracing.v_030.SpanContext;
+import io.opentracing.v_030.Tracer;
+import io.opentracing.v_030.propagation.Format;
+
+import java.util.Map;
+
+public class BaseTracerShim implements Tracer {
+    final io.opentracing.Tracer tracer;
+
+    protected BaseTracerShim(io.opentracing.Tracer tracer) {
+        checkArgumentNotNull(tracer, "tracer");
+
+        this.tracer = tracer;
+    }
+
+    void checkArgumentNotNull(Object value, String errorMessage) {
+        if (value == null)
+            throw new IllegalArgumentException(errorMessage);
+    }
+
+    protected ActiveSpanShim createActiveSpanShim(Scope scope) {
+        return new ActiveSpanShim(scope);
+    }
+
+    protected SpanShim createSpanShim(io.opentracing.Span span) {
+        return new SpanShim(span);
+    }
+
+    @Override
+    public ActiveSpan activeSpan() {
+        if (tracer.scopeManager().active() == null)
+            return null;
+
+        return createActiveSpanShim(tracer.scopeManager().active());
+    }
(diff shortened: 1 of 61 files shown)


rbtcollins:
Haven't reviewed in detail, but thank you!

malafeev:
I cannot set tag on ActiveSpan now:
```
import io.opentracing.v_030.ActiveSpan;
import io.opentracing.v_030.tag.Tags;

ActiveSpan activeSpan = ...;
Tags.COMPONENT.set(activeSpan, "component");
```

carlosalberto:
@malafeev Hey, thanks for the catch. Now it's fixed (the `tags` subpackage was referencing `Span` instead of `BaseSpan`).

malafeev:
thanks @carlosalberto , verified

pavolloffay:
> Have a io.opentracing.v_030.shim.TracerShim

does it mean that all imports (ActiveSpan, BaseSpan) have to be changed to this?

carlosalberto:
@pavolloffay Yes - for people using this Shim + the old API, that is (as `io.opentracing.` will contain the *new* API).

There were some ideas in how to help people with this. Maybe some Gradle/Maven plugin/script to refactor their 'old' code. Maybe this can be played with down the road?

As a note, the idea is to let people using the 0.31 API to incrementally update their instrumentation code using 0.30, so this would be a small price to pay.

malafeev:
I would be more strict: don't add compatibility package. 
Who wants to migrate - will do.
Using this compatibility also takes time. Better migrate to new api.

tylerbenson:
I agree with @malafeev.  I would rather not add all this additional complexity.  My main concern then is does this require trace implementers to then implement both apis?

rbtcollins:
Updating an import: 30 seconds. Changing programming model across a framework? Hours or days.

The refcount tracking code should be entirely generic and not need any implementation by backend providers AIUI.

malafeev:
If we have production issue then 30 seconds vs hours/days makes sense.
But in case of not stable yet api supporting previous api just to free couple of hours/(days) looks not right.
I prefer force people to upgrade until we have 1.0 and start thinking how to make breaking changes.

rbtcollins:
So you'll place users in a choice of having to do a more complex thing, or be at risk of old dependencies which may have security flaws in it. As soon e.g. spring upgrades their bindings to 0.31, everyone with custom code written in spring also has to upgrade, also in lockstep: this is a transitive problem.

Forcing a lock-step upgrade when none is strictly required brings no benefit to the project, but can make your users life more stressful. I really don't understand why you'd ever do that.

Given that code is already written, please don't force other people to coordinate changes to your schedule.

malafeev:
is there any example of using it?
I'm not sure that it will work as you described for spring app.
You cannot mix instrumented libs with v030 and v031, it will not work even with this package.

rbtcollins:
But you can update with super-little work with the shim, no?

malafeev:
ok, if it works for you. for me it's a headache to make this super-little work.

opentracing-v030/src/main/java/io/opentracing/v_030/shim/BaseTracerShim.java
```diff
@@ -0,0 +1,163 @@
...
+    private final class SpanBuilderShim implements Tracer.SpanBuilder {
+        io.opentracing.Tracer.SpanBuilder builder;
+
+        public SpanBuilderShim(io.opentracing.Tracer.SpanBuilder builder) {
+            this.builder = builder;
+        }
+
+        @Override
+        public SpanBuilderShim asChildOf(SpanContext parent) {
+            checkArgumentNotNull(parent, "parent");
```
pavolloffay: this should not throw exception

pavolloffay: the same for other child/references methods

pavolloffay:
I didn't find any tests for Sope-ActiveSpan move (and vice versa).

opentracing-v030/src/main/java/io/opentracing/v_030/shim/BaseTracerShim.java
carlosalberto: Yeah, I was thinking about it earlier. I'd like to report bad arguments, but then again, these ones should stick to the standard. Will adjust them as expected.

carlosalberto:
Closing this in favor of the new `opentracing-java-v030` repository effort.

PR #223, opened 2017-11-30
Change BINARY to be resizable and stream-oriented.

carlosalberto:
This is an effort to provide a stream-based, resizable BINARY format, inspired by #99 and its feedback.

`BinaryHolder` is an adapter for either `ByteArrayOutputStream` or `ByteArrayInputStream`, depending on whether the format was created for injection or extraction.

I exposed two static methods to try to make purpose clear, i.e. `createInjectionHolder()`, `createExtractionHolder`, instead of using simple constructor(s).

Any feedback is appreciated ;)


yurishkuro:
Glad to see us moving on this. I have a few high level design suggestions.
  * I would prefer to define an interface `Binary` to make it clear which methods are expected from the carrier. The Holder concrete class can be an implementation of that interface.
  * I'm not thrilled with the name Holder, perhaps Adapter
  * what might be even better is to define a single `Adapters` class that has a bunch of static factory  methods, such as `inboundTextMap(Map<String,String>)`, `inboundBinary(byte[])`. This way the helper/holder/adapter classes can be private, which is what std library does, e.g. `Collections`.

carlosalberto:
Hey @yurishkuro thanks for the feedback! So this is an updated PR with the mentioned changes (the only thing was that I left out for now the `TextMap` related methods for `Adapters`, as I'd love to initially have the very BINARY part tuned and ready ;) ) Let me know.

opentracing-api/src/main/java/io/opentracing/propagation/Adapters.java
```diff
@@ -0,0 +1,41 @@
+/*
+ * Copyright 2016-2017 The OpenTracing Authors
```
yurishkuro: nit: as this is new code, (c) is only 2017 (same in other files)

opentracing-api/src/main/java/io/opentracing/propagation/Binary.java
```diff
@@ -0,0 +1,65 @@
...
+public interface Binary {
+    /**
+     * @return whether this instance is inbound and supports reading.
+     */
+    boolean isInbound();
+
+    /**
+     * @return whether this instance is outbound and supports writing.
+     */
+    boolean isOutbound();
```
yurishkuro: are these required?  We didn't have them in TextMap afaik. E.g. I don't know why a Tracer would want to call these.

opentracing-api/src/main/java/io/opentracing/propagation/Binary.java
```diff
@@ -0,0 +1,65 @@
...
+
+    /**
+     * Writes b.length bytes from the specified byte array. The internal
+     * buffer should grow as data is written to it.
+     *
+     * @param b the data.
+     */
+    void write(byte[] b) throws IOException;
```
yurishkuro: I don't have a strong opinion about the exact read/write API, but would it make more sense to model these functions after nio [ReadableByteChannel](https://docs.oracle.com/javase/7/docs/api/java/nio/channels/ReadableByteChannel.html) and [WritableByteChannel](https://docs.oracle.com/javase/7/docs/api/java/nio/channels/WritableByteChannel.html)? They each have just a single method, but it's based on ByteBuffer, which I'm guessing makes them equivalent to several overloaded methods of Input/OutputStream, and certainly more flexible than `write(byte[] b)` (which requires an allocation if you're writing in batches of different sizes). Maybe it's not an issue for propagation codecs specifically.

opentracing-api/src/main/java/io/opentracing/propagation/Binary.java
```diff
@@ -0,0 +1,65 @@
...
+     * @param b the buffer into which the data is read.
+     *
+     * @return the total number of bytes read into the buffer, or -1 if there is no more data.
+     */
+    int read(byte[] b) throws IOException;
+
+    /**
+     * @return A newly allocated byte array with the written data.
+     */
+    byte[] getOutboundData();
```
yurishkuro: seems premature. Can always be implemented with a loop by calling read(), although less efficiently without knowing the full size of the payload.

opentracing-api/src/main/java/io/opentracing/propagation/Binary.java
carlosalberto: Would be more of helper methods (as the `Binary` format either supports injection or extraction, not both). We can remove them if you think they don't bring enough value.

opentracing-api/src/main/java/io/opentracing/propagation/Binary.java
carlosalberto: So this is for the written data (when `inject()` was called), and this is supposed to be called *after* `inject()` to retrieve all those bytes. The alternatives is to let the user pass us a `ByteArrayOutputStream` (or similar) at `Adapters.outboundBinary()`. What do you think?

opentracing-api/src/main/java/io/opentracing/propagation/Adapters.java
carlosalberto: :) Had missed that, will fix.

carlosalberto:
Regarding the naming, what about `Adapters.injectionBinary()` instead of `Adapters.outboundBinary()`, and `Adapters.extractionBinary()` instead of `Adapters.inboundBinary()`?

malafeev:
What's missing for me is MockTracer support. I would like to see tests with MockTracer inject/extract binary data.

opentracing-api/src/main/java/io/opentracing/propagation/Adapters.java
```diff
@@ -0,0 +1,41 @@
...
+ * Unless required by applicable law or agreed to in writing, software distributed under the License
+ * is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express
+ * or implied. See the License for the specific language governing permissions and limitations under
+ * the License.
+ */
+package io.opentracing.propagation;
+
+import java.io.IOException;
+
+public final class Adapters {
```
gkumar7: I do not see a real benefit of having this class. Simply calling the appropriate constructor of BinaryAdapter in a call to inject or extract should do.

opentracing-api/src/main/java/io/opentracing/propagation/Binary.java
gkumar7: +1 to remove, does not add much value

opentracing-api/src/main/java/io/opentracing/propagation/BinaryAdapter.java
```diff
@@ -0,0 +1,73 @@
...
+        outputStream.write(b);
+    }
+
+    public int read(byte [] b) throws IOException {
+        if (inputStream == null)
+            throw new UnsupportedOperationException();
+
+        return inputStream.read(b);
+    }
+
```
gkumar7: Limiting the methods of ```ByteArrayOutputStream``` and ```ByteArrayInputStream``` to only ```read(byte[] b)``` and ```write(byte[] b)``` may limit tracing implementation libraries. Suppose this class simply had two getters:

```java
public InputStream getInput() {
  if (inputStream == null) {
    throw new UnsupportedOperationException();
  }
  return inputStream;
}

public OutputStream getOutput() {
  if (outputStream == null) {
    throw new UnsupportedOperationException();
  }
  return outputStream;
}
```
This would allow much greater flexibility to the tracing implementations (by not being rigid on what/how to read and when to close) and yet satisfy the api as defined by ```inject``` and ```extract```.

opentracing-api/src/test/java/io/opentracing/propagation/BinaryAdapterTest.java
```diff
@@ -0,0 +1,64 @@
...
+    @Test
+    public void inboundTest() throws IOException {
+        byte[] ctx = new byte[] { 1, 2, 3, 4, 4, 3, 2, 1 };
+        Binary binary = Adapters.inboundBinary(ctx);
+        assertEquals(true, binary.isInbound());
+        assertEquals(false, binary.isOutbound());
+
+        byte[] buff = new byte[4];
+
+        assertEquals(4, binary.read(buff));
```
gkumar7: nit: let's define some constants

opentracing-api/src/main/java/io/opentracing/propagation/Adapters.java
carlosalberto: The idea is to put other formats here, so we can abstract the actual implementations, so we don't have to expose `BinaryAdapter` and others.

opentracing-api/src/main/java/io/opentracing/propagation/BinaryAdapter.java
carlosalberto: I personally think exposing the actual stream objects would be overloading the API, trying to cover all cases and scenarios eventually. In any case, if this is truly needed, can be added in the future.

opentracing-api/src/test/java/io/opentracing/propagation/BinaryAdapterTest.java
carlosalberto: This is a very, very simple test, no need to put constants. Maybe we can have `assertEquals(buff.length, binary.read(buff))` and that should be an improvement already ;)

carlosalberto:
@malafeev Oh, sure, will cook some examples - the thing is that I need to add support for the `BINARY` format itself for `MockTracer` (which hopefully can be done soon).

opentracing-api/src/main/java/io/opentracing/propagation/Binary.java
carlosalberto: Makes sense. Lets get rid of them.

opentracing-api/src/main/java/io/opentracing/propagation/BinaryAdapter.java
carlosalberto: On a second thought: we *may* as well remove the write/read methods and end up exposing the streams too (after prototyping the `MockTracer` support for BINARY) ;)

opentracing-api/src/main/java/io/opentracing/propagation/Binary.java
hypnoce: Agreed. ByteBuffer provides a greater level of flexibility while still being able to wrap (without copy) a byte[] with ByteBuffer.wrap

opentracing-mock/src/main/java/io/opentracing/mock/MockTracer.java
```diff
@@ -112,6 +119,81 @@ protected void onSpanFinished(MockSpan mockSpan) {
...
+                        for (Map.Entry<String, String> entry : ctx.baggageItems()) {
+                            objStream.writeUTF(entry.getKey());
+                            objStream.writeUTF(entry.getValue());
+                        }
+
+                        objStream.flush(); // *need* to flush ObjectOutputStream.
+                        binary.write(ByteBuffer.wrap(stream.toByteArray()));
+
+                    } catch (IOException e) {
+                        throw new IllegalArgumentException("Corrupted data");
```
malafeev: shouldn't be RuntimeException?

opentracing-api/src/main/java/io/opentracing/propagation/Adapters.java
```diff
@@ -0,0 +1,85 @@
...
+     * Creates an outbound Binary instance used for injection, backed up
+     * by the specified OutputStream as output.
+     *
+     * @param stream The OutputStream used as the output.
+     *
+     * @return The new Binary carrier used for injection.
+     */
+    public static Binary injectBinary(OutputStream stream) {
+        if (stream == null)
+            throw new IllegalArgumentException("stream");
```
malafeev: isn't better "stream cannot be null" message?

malafeev: I would use curly braces in all `if` blocks even with only one statement

carlosalberto: Oh, definitely. Missed this one.

opentracing-mock/src/main/java/io/opentracing/mock/MockTracer.java
carlosalberto: Makes sense.

malafeev:
@carlosalberto great!

carlosalberto:
This is an updated PR including:

* `ByteBuffer` for `read()`/`write()`.
* `injectBinary()` and `extractBinary()` names for the adapters (instead of `inbound`/`outbound`)
* (Basic) BINARY propagator for MockTracer (using the old ObjectDataStream classes).
* Overloads taking both streams and channels in `Adapters`.

With the support for BINARY in `MockTracer` it should be easy to test this approach out and verify it's good enough. Let me know. Hopefully we are closer to the final PR ;) @yurishkuro @malafeev @gkumar7

opentracing-api/src/main/java/io/opentracing/propagation/Adapters.java
carlosalberto: Oh, btw, if I use `2017` alone the build fails. Guess we need to issue a PR for fixing this part in pom.xml ;)

malafeev:
@carlosalberto LGTM

PR #226, opened 2017-11-30
Make MockTracer::ChildOf not fail if there is a null argument passed

elgris:
According to [SpanBuilder interface](https://github.com/opentracing/opentracing-java/blob/master/opentracing-api/src/main/java/io/opentracing/Tracer.java#L106), `ChildOf` called with `null` parent should be a noop. This pull request fixes this behaviour for MockTracer.

### DIFF
--- a/opentracing-mock/src/test/java/io/opentracing/mock/MockSpanTest.java
+++ b/opentracing-mock/src/test/java/io/opentracing/mock/MockSpanTest.java
@@ -78,4 +78,12 @@ public class MockSpanTest {
         }
         Assert.assertEquals(1, tracer.finishedSpans().get(0).generatedErrors().size());
     }
+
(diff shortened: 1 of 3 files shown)


opentracing-mock/src/test/java/io/opentracing/mock/MockSpanTest.java
```diff
@@ -78,4 +78,12 @@ public void testAddBaggageAfterFinish() {
         }
         Assert.assertEquals(1, tracer.finishedSpans().get(0).generatedErrors().size());
     }
+
+    @Test
+    public void testChaildOfWithNullParentDoesNotThrowException() {
```
sjoerdtalsma: typo: `Chaild` -> `Child`

PR #238, opened 2017-12-28
Reinstate BaseSpan interface for backwards compatibility

### LINKED ISSUE #237: [0.31.0-RC2] missing BaseSpan
I am trying to upgrade Jaeger client to RC2. It contains an integration test using tchannel that is compiled against 0.30.0. Even though the instrumentation never uses BaseSpan explicitly, if fails at runtime:
```
com.uber.jaeger.crossdock.resources.behavior.http.TraceBehaviorResourceTest > testJoinTraceTChannel[1: sampled(false)] FAILED
    java.lang.NoClassDefFoundError: io/opentracing/BaseSpan
        at com.uber.tchannel.tracing.Tracing.startOutboundSpan(Tracing.java:107)
        at com.uber.tchannel.api.SubChannel.sendOutRequest(SubChannel.java:266)
        at com.uber.tchannel.api.SubChannel.sendRequest(SubChannel.java:245)
        at com.uber.tchannel.api.SubChannel.send(SubChannel.java:192)
        at com.uber.jaeger.crossdock.resources.behavior.TraceBehavior.callDownstreamTChannel(TraceBehavior.java:108)
        at com.uber.jaeger.crossdock.resources.behavior.http.TraceBehaviorResourceTest.testJoinTraceTChannel(TraceBehaviorResourceTest.java:174)

        Caused by:
        java.lang.ClassNotFoundException: io.opentracing.BaseSpan
            at java.net.URLClassLoader.findClass(URLClassLoader.java:381)
            at java.lang.ClassLoader.loadClass(ClassLoader.java:424)
            at sun.misc.Launcher$AppClassLoader.loadClass(Launcher.java:335)
            at java.lang.ClassLoader.loadClass(ClassLoader.java:357)
            ... 6 more
```

Simply adding the BaseSpan interface back to the module seems to fix this issue (I will post a PR once confirmed).

yurishkuro:
Fixes #237

### DIFF
--- a/opentracing-api/src/main/java/io/opentracing/BaseSpan.java
+++ b/opentracing-api/src/main/java/io/opentracing/BaseSpan.java
@@ -1,0 +1,40 @@
+/*
+ * Copyright 2016-2017 The OpenTracing Authors
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License"); you may not use this file except
+ * in compliance with the License. You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software distributed under the License
+ * is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express
+ * or implied. See the License for the specific language governing permissions and limitations under
+ * the License.
+ */
+package io.opentracing;
+
+import java.util.Map;
+
+/**
+ * BaseSpan is carried over from 0.30.0 API to preserve backwards compatibility with instrumentation
+ * that was compiled against v0.30.0. Even if that instrumentation does not refer to BaseSpan directly,
+ * it is still required at runtime because most of the Span methods in v0.30.0 were inherited from BaseSpan.
+ * 
+ * Keeping this interface around allows for easier transition to v0.31.0.
+ * 
+ * @see Span
+ * @deprecated since v0.31.0
+ */
+public interface BaseSpan {
+    // SpanContext context();
+    // S setTag(String key, String value);
+    // S setTag(String key, boolean value);
+    // S setTag(String key, Number value);
+    // S log(Map<String, ?> fields);
+    // S log(long timestampMicroseconds, Map<String, ?> fields);
+    // S log(String event);
+    // S log(long timestampMicroseconds, String event);
+    // S setBaggageItem(String key, String value);
+    // String getBaggageItem(String key);
+    // S setOperationName(String operationName);
+}
--- a/opentracing-api/src/main/java/io/opentracing/BaseSpan.java
+++ b/opentracing-api/src/main/java/io/opentracing/BaseSpan.java
@@ -13,28 +13,15 @@
  */
 package io.opentracing;
 
-import java.util.Map;
-
 /**
  * BaseSpan is carried over from 0.30.0 API to preserve backwards compatibility with instrumentation
  * that was compiled against v0.30.0. Even if that instrumentation does not refer to BaseSpan directly,
- * it is still required at runtime because most of the Span methods in v0.30.0 were inherited from BaseSpan.
+ * it is still required at runtime (https://github.com/opentracing/opentracing-java/issues/237).
  * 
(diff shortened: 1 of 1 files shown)


yurishkuro:
> As a small question: maybe it's worth adding a `@deprecated`

Doh! Forgot to hit Cmd-S.

tedsuo:
If no one blocks, I propose we merge this tomorrow afternoon (Friday jan 5th).

pavolloffay:
I have compiled https://github.com/jaegertracing/jaeger-client-java/pull/313 with `tchannel` `0.7.7` and OT with this PR.

and I sill get other compatibility issues (CNDF) for other classes see: https://pastebin.com/1HNyZMA2. I am still not sure what this PR solves.

pavolloffay:
I was very curious why this actually works and is necessary because `tchannel` does not use BaseSpan https://github.com/uber/tchannel-java/blob/tchannel-0.7.7/tchannel-core/src/main/java/com/uber/tchannel/tracing/Tracing.java it's not imported in the whole repository, therefore it should work without this PR. 

However, there is an anonymous class on line 107 https://github.com/uber/tchannel-java/blob/tchannel-0.7.7/tchannel-core/src/main/java/com/uber/tchannel/tracing/Tracing.java#L107 and compiler decided to cast parameters when setting the tag to `BaseSpan` - in `Tags.ERROR.set(span, true);`

See decompiled code:
screenshot of decompiling java and android applications

So with this PR there is `BaseSpan` symbol present. Then why does it not producer ClassCastException? Becase in 0.30.0  `AbstractTag.set` accepts generic `BaseSpan<?>`  https://github.com/opentracing/opentracing-java/blob/master/opentracing-api/src/main/java/io/opentracing/tag/AbstractTag.java#L29 and if the last section from https://docs.oracle.com/javase/tutorial/java/generics/bridgeMethods.html holds. Then there is `AbstractTag.set(object, ...)` which is then internally casted to `Span`. 


If you think about this it practically solves only this use case and it's using hack that java compiler generates "bridge method" (part of type erasure)

tedsuo:
Thanks for the detailed review @pavolloffay!  After running this by a number of people, I feel like it should go in.

pavolloffay:
My comment https://github.com/opentracing/opentracing-java/pull/238#issuecomment-355632684 is not entirely correct. 

So why does it actually work? 

Tchannel code does not import BaseSpan explicitly but java compiler adds cast to BaseSpan in all `AbstractTag.set(AbstractSpan)` calls. It can be seen in the screenshot above. 

This fix works in jaeger-client because the branch https://github.com/uber/tchannel-java/blob/tchannel-0.7.7/tchannel-core/src/main/java/com/uber/tchannel/tracing/Tracing.java#L111 - `if(response.isError())` is never executed so JVM is happy even with incompatible class introduced in this PR. If the code executes it produces:

```java
java.lang.reflect.InvocationTargetException
	at sun.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
	at sun.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:62)
	at sun.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
	at java.lang.reflect.Method.invoke(Method.java:498)
	at org.codehaus.mojo.exec.ExecJavaMojo$1.run(ExecJavaMojo.java:294)
	at java.lang.Thread.run(Thread.java:748)
Caused by: java.lang.NoSuchMethodError: io.opentracing.tag.BooleanTag.set(Lio/opentracing/BaseSpan;Ljava/lang/Boolean;)V
	at sk.loffay.ot.test.TracedClass.invoke(TracedClass.java:20)
	at sk.loffay.ot.test.module2.Main.main(Main.java:17)
	... 6 more
```

So this fix helps only if the code is actually never executed and introduces even more difficult runtime issues. If somebody wants to play with it https://github.com/pavolloffay/ot-compatibility-test

tedsuo:
Good catch.

yurishkuro:
let's release without this. I spoke to our teams, we will coordinate a new release of Jaeger and TChannel libs to go in sync. Most services internally do not register these dependencies explicitly, but via internal wrapper libs, so we mostly need to bump deps in those.

PR #241, opened 2018-01-03
Auto finish scope manager

carlosalberto:
Creating this PR in order to move the `AutoFinishScopeManager` from the examples module into the util one, so this can be accessible to other libraries (even if we don't support the `Continuation` concept out of the box anymore).

I'm playing with this possibility as both the v030 and the akka support will optionally offer this flavor of `ScopeManager`, and thought it could be helpful for other people/projects overall.

Opinions on this? Shall we put it here and expose it publicly?


pavolloffay:
LGTM, I would be interested how this will work with brave-opentracing

carlosalberto:
@pavolloffay sure!

@yurishkuro Objections?

yurishkuro:
fine with me

carlosalberto:
Merged, preparing it to be part of the final release.

Issue #239, opened 2018-01-02
Update Copyright year


malafeev:
Year in Copyright is 2016-2017

malafeev:
fixed

PR #244, opened 2018-01-09
Update the Copyright year to 2018.

carlosalberto:
Preparing the merge of  `v0.31.0` into `master` prior to the release, I came with the need to update the Copyright year for the merged files. So a choice is to do a full pass once, or else do it manually, as we go. 

Opinions?

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -13,7 +13,7 @@ If you are adding a new file it should have a header like below. This can be aut
 
 ```
 /**
- * Copyright 2016-2017 The OpenTracing Authors
+ * Copyright 2016-2018 The OpenTracing Authors
  *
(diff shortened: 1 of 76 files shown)


yurishkuro:
once is better (less overhead), although technically incorrect

carlosalberto:
Forgot to mention this will cover #239

malafeev:
am I right that there is no automatic check of copyright year during the build?
I don't see that `license.git.copyrightYears` is set  (`src/etc/header.txt`).

pavolloffay:
@malafeev a year should be derived from the git history

pavolloffay:
There is also section in release.md https://github.com/opentracing/opentracing-java/blob/master/RELEASE.md#first-release-of-the-year

malafeev:
thanks @pavolloffay

Issue #243, opened 2018-01-08
Avoiding `Binary` format to be based on byte buffers


raphw:
Just a small side-line note: I do not think it is a good idea to bind the `Binary` format onto `ByteBuffer`s. Frameworks such as Netty are explicitly built around avoiding `ByteBuffer` usage as they can be rather performance heavy and uses its own `ByteBuf` abstraction. It would be of advantage to hooking any form of byte-cosumer / byte-producer into this API.

Also, I do not think that the average user would want to write a binary format to a buffer but rather to a stream. And even when using NIO, I cannot see how an implementor of the API could push or poll bytes asynchronously since the implementation needs to block on the tracer's `inject` and `extract` methods. I therefore think that `Binary` should rather be some form of callback carrier such as an input or output stream what are at least interfaces that can delegate to a byte byffer if needed without any practical performance impact.

Finally, for the current API, it is not clear to me how the return values of binary are supposed to be used. A `ByteBuffer` already has internal offsets which are altered during read/write. This only opens a possibility for inconsistencies.

Thanks for clarifying and considering my suggestion.

yurishkuro:
@raphw are you referring to the current API in master, or to the next version here: https://github.com/opentracing/opentracing-java/blob/v0.31.0/opentracing-api/src/main/java/io/opentracing/propagation/Binary.java#L34 ?

raphw:
I am refering to the next version that you linked.

yurishkuro:
The next version mirrors nio stream interface. Why wouldn't it work?

For example, for inject() flow
  * the instrumentation creates a WritableByteChannel backed by whatever, then wraps it in BinaryAdapter
  * the tracer.inject() writes the trace context to a ByteBuffer (which can be pooled to avoid memory allocations) , and then writes its content to the channel

How would you like the API to look?

raphw:
The `ReadableByteChannel` and `WritableByteChannel` abstractions are a concept from Java NIO for implementing asynchronous I/O. The idea is that a single thread can read/write from multiple such channels and suspend a channel if the channels buffer is exhausted. This comes with the advantage that a single thread does not need to block a thread for a single I/O transaction to complete but can handle multiple such transactions.

[...]

PR #245, opened 2018-01-09
Make Binary use byte arrays instead of ByteBuffer instances.

carlosalberto:
This is an approach for 'solving' #243, as a compromise between streams and `ByteBuffer` approaches.

Adapters for `Channel`s have been removed, but we can include them in next releases if still desired.

`MockTracer` has been updated as well.


carlosalberto:
Closing as we have decided to roll back the current binary format until consensus is achieved.

PR #246, opened 2018-01-11
Binary release adjustments

carlosalberto:
This is a PR to update `MockTracer` and the inline documentation for `Binary` prior to the 0.31 release (after the feedback we got in #243).

* MockTracer's Binary propagator must loop over `Binary.write()` when injecting a context (instead of a *single* write() ).
* Clarify that the `Binary` methods for reading/writing *work* like the ones of a (synchronous) `Channel`, but they don't need to belong to an actual `Channel`.

### DIFF
--- a/opentracing-api/src/main/java/io/opentracing/propagation/Binary.java
+++ b/opentracing-api/src/main/java/io/opentracing/propagation/Binary.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2016-2017 The OpenTracing Authors
+ * Copyright 2016-2018 The OpenTracing Authors
  *
  * Licensed under the Apache License, Version 2.0 (the "License"); you may not use this file except
  * in compliance with the License. You may obtain a copy of the License at
@@ -36,7 +36,8 @@ public interface Binary {
      * Writes a sequence of bytes to this carrier from the given buffer.
      * The internal buffer is expected to grow as more data is written.
      *
-     * The behavior of this method is expected to be the same as WritableByteChannel.write().
+     * The behavior of this method is expected to be the same as WritableByteChannel.write(),
+     * with the exception of being synchronous.
      *
      * @param buffer The buffer from which bytes are to be retrieved.
      *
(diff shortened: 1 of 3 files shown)


opentracing-api/src/main/java/io/opentracing/propagation/Binary.java
```diff
@@ -33,10 +33,11 @@
  */
 public interface Binary {
     /**
-     * Writes a sequence of bytes to this channel from the given buffer.
+     * Writes a sequence of bytes to this carrier from the given buffer.
      * The internal buffer is expected to grow as more data is written.
      *
-     * The behavior of this method is expected to be the same as WritableByteChannel.write().
+     * The behavior of this method is expected to be the same as WritableByteChannel.write(),
+     * with the exception of being synchronous.
```
yurishkuro: Not sure if this is accurate, I think we want to make it return void. I.e. either write all content or throw exception.

carlosalberto: A single operation then? That would work just fine actually. That would simplify things a lot. The question would be what to do for `read()` then - throw an `IOException` if the passed `ByteBuffer` is too small (as you mentioned in the mentioned discussion)?

(Returning a `ByteBuffer` from `read()` wouldn't be so nice IHMO ;) )

yurishkuro: no, it can still be multiple calls if the tracer's buffer is smaller than the serialized context. The point is that the carrier must accept all or nothing. But if you want to keep returning int it's fine too.

I'd rather punt on this whole thing in 0.31. Yes, its going to be a breaking change in the next release, but nobody uses the existing binary API anyway.

carlosalberto: After prototyping stuff a little bit, I agree the best of is to expect the `ByteBuffer` to be written *all* (this `void write(...)`), even if we allow multiple calls to it. Will update this PR earlier today with an update to `Adapters` ;)

carlosalberto:
Closing as we have decided to roll back the current binary format until consensus is achieved.

PR #248, opened 2018-01-11
Revert to the 0.30 Binary format.

carlosalberto:
Since consensus hasn't been reached for the new
binary format, we will roll it back for now,
in order to have the release the 0.31 branch,
and iterate over it in the mean time.


carlosalberto:
cc @yurishkuro @tedsuo

PR #251, opened 2018-01-16
Add a CHANGELOG file

tedsuo:
In the same manner as the golang changelog: https://github.com/opentracing/opentracing-go/blob/master/CHANGELOG.md

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,13 +1,9 @@
-Changes by Version
-==================
+#Changes by Version
 
-v0.31.0 (2018-01-12)
---------------------
-
-- `BaseSpan` and `ActiveSpan` are simplified into a single `Span` class.
-- `Scope` replaces `ActiveSpan`, removing the continuation concept. There is no longer a capture phase. Scopes are only activated and then closed.
-- `ScopeManager` replaces `ActiveSpanSource`, and is expected to be passed to the tracer at initialization time.
-- `ScopeManager.activate(Span,bool)` creates a `Scope` for the current context. The `finishOnClose` flag optionally calls `Span.finish()` on `Scope.close()`.
-- `ScopeManager.active()` returns the `Scope` associated with the current context.
-- `Tracer.activeSpan()` returns the current value of `Tracer.scopeManager().active().span()` as a convenience.
-- `ThreadLocalScopeManager` replaces `ThreadLocalActiveSpanSource`. It still uses thread-local storage for tracking active spans, but removes the reference counting used previously for span lifetime handling.
+## v0.31.0 (2018-01-12) 
+* `BaseSpan` and `ActiveSpan` are simplified into a single `Span` class.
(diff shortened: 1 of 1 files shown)


CHANGELOG.md
```diff
@@ -0,0 +1,13 @@
+Changes by Version
+==================
```
pavolloffay: Could be used the same formatting symbols as in other markdown files in this repository. e.g. `#` for headers, `*` for lists

CHANGELOG.md
```diff
@@ -0,0 +1,13 @@
...
+
+v0.31.0 (2018-01-12)
+--------------------
```
pavolloffay: Could we add one list item per added change? All the following would be in the one list item. Also, add a link to the PR, in some projects there is also a name of the commiter.

Two major changes are missing there:
* introduction of log fields
* removed `span.log(object)` methods

yurishkuro:
should we adopt the http://keepachangelog.com/en/1.0.0/ format?

CHANGELOG.md
tedsuo: I was trying to be uniform by copying the style in the go CHANGELOG ;) : https://raw.githubusercontent.com/opentracing/opentracing-go/master/CHANGELOG.md\

We also use `.rst` instead of `.md` in some repos. 🤷‍♂️ 

I prefer `#` and `*` so I'll switch to that. Maybe we can standardize and change the others as part of the upcoming "language maintainer" initiative.

CHANGELOG.md
tedsuo: I'm not sure what you mean by "All the following would be in the one list item." Do you mean only one list item for all the Scpoce/ActiveSpan changes?

tedsuo:
@yurishkuro http://keepachangelog.com/en/1.0.0/ seems nice but I'd like to get a language maintainer group together to start making these decisions. There's a set of things we should standardize on, such as PR/Issue templates, release procedures, etc...

CHANGELOG.md
pavolloffay: Yes as it was part of one PR

tedsuo: I looked though the PRs... unfortunately I don't think the Pull Requests line up cleanly, as we went through many revisions. I will try to find a middle ground.

tedsuo:
@pavolloffay @carlosalberto @yurishkuro  I updated this PR, sorry for the delay.

- The PRs in the v0.31 were too messy to reference, as a result of numerous revisions, so I did not include them.
- We could include a link to the tracking issue, but I'm not sure how helpful that is (https://github.com/opentracing/opentracing-java/issues/189)
- Instead I simplified the line items to objects which were added and removed from the core API, and method changes on the Tracer object. Hopefully that's the right level of detail.

pavolloffay:
LGTM, although at least one minor yet important thing is missing. Removed `log.(Object)` methods.

pavolloffay:
I will merge this and open subsequent PR to add more links in there.

PR #252, opened 2018-01-19
Binary format proposal.

carlosalberto:
* Uses ByteBuffer to read/write data *synchronously*.
* write() is expected to read all passed data.
* read() may need many calls in case the buffer is not big enough.
* Adapter for InputStream/OutputStream.
* MockTracer BINARY propagator.

### DIFF
--- a/opentracing-api/src/main/java/io/opentracing/propagation/Binary.java
+++ b/opentracing-api/src/main/java/io/opentracing/propagation/Binary.java
@@ -1,0 +1,51 @@
+/*
+ * Copyright 2016-2018 The OpenTracing Authors
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License"); you may not use this file except
+ * in compliance with the License. You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software distributed under the License
+ * is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express
+ * or implied. See the License for the specific language governing permissions and limitations under
+ * the License.
+ */
+package io.opentracing.propagation;
+
+import java.io.IOException;
+import java.nio.ByteBuffer;
+
+/**
+ * Binary is an interface defining the required operations for a binary carrier for
+ * Tracer.inject() and Tracer.extract(). Binary can be defined either as inbound (extraction)
+ * or outbound (injection), and is expected to work synchronously.
+ *
+ * When Binary is defined as inbound, read() must be used to read data,
+ * and it is an error to call read().
+ *
+ * When Binary is defined as outbound, write() must be used to write data,
+ * and it is an error to call write().
+ *
+ * @see Format.Builtin#BINARY
+ * @see io.opentracing.Tracer#inject(SpanContext, Format, Object)
+ * @see io.opentracing.Tracer#extract(Format, Object)
+ */
+public interface Binary {
+    /**
+     * Writes a sequence of bytes to this carrier from the given buffer.
+     * The internal buffer is expected to grow as more data is written.
+     *
+     * @param buffer The buffer from which bytes are to be retrieved.
+     */
+    void write(ByteBuffer buffer) throws IOException;
+
+    /**
+     * Reads a sequence of bytes into the given buffer.
+     *
+     * @param buffer The buffer into which bytes are to be transferred.
+     *
+     * @return The number of bytes read, possibly zero, or -1 if the carrier has reached end-of-stream.
+     */
+    int read(ByteBuffer buffer) throws IOException;
+}
--- a/opentracing-api/src/main/java/io/opentracing/propagation/BinaryAdapter.java
+++ b/opentracing-api/src/main/java/io/opentracing/propagation/BinaryAdapter.java
@@ -1,0 +1,90 @@
+/*
+ * Copyright 2016-2018 The OpenTracing Authors
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License"); you may not use this file except
+ * in compliance with the License. You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software distributed under the License
+ * is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express
+ * or implied. See the License for the specific language governing permissions and limitations under
+ * the License.
+ */
+package io.opentracing.propagation;
+
+import java.io.IOException;
+import java.io.OutputStream;
+import java.io.InputStream;
+import java.nio.ByteBuffer;
+
+/**
+ * BinaryAdapter is a built-in carrier for Tracer.inject() and Tracer.extract(). BinaryAdapter
+ * is backed up by either an OutputStream or an InputStream, depending
+ * on whether it's defined as injection or extraction, respectively.
+ */
+final class BinaryAdapter implements Binary {
+    private final OutputStream outputStream;
+    private final InputStream inputStream;
+
+    /**
+     * Create an outbound BinaryAdapter backed by the specified OutputStream.
+     */
+    BinaryAdapter(OutputStream outputStream) {
+        this.outputStream = outputStream;
+        this.inputStream = null;
+    }
+
+    /**
+     * Create an inbound BinaryAdapter backed by the specified InputStream.
+     */
+    BinaryAdapter(InputStream inputStream) {
+        this.inputStream = inputStream;
+        this.outputStream = null;
+    }
+
+    OutputStream outputStream() {
+        return outputStream;
+    }
+
+    InputStream inputStream() {
+        return inputStream;
+    }
+
+    public void write(ByteBuffer buffer) throws IOException {
+        if (buffer == null) {
+            throw new NullPointerException();
+        }
+        if (outputStream == null) {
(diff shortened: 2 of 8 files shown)


carlosalberto:
See #253

opentracing-api/src/main/java/io/opentracing/propagation/BinaryAdapter.java
```diff
@@ -0,0 +1,90 @@
...
+            throw new UnsupportedOperationException();
+        }
+
+        // Need to read always as:
+        // 1. InputStream.available() always returns 0 for some implementations.
+        // 2. Even buffer.remaining() == 0, we need to know whether we have reached
+        //    the end of the stream or not.
+        byte[] b = new byte[buffer.remaining()];
+        int available = inputStream.read(b);
+        if (available < 0)
```
malafeev: please add curly braces for `if` body

opentracing-mock/src/main/java/io/opentracing/mock/MockTracer.java
```diff
@@ -112,6 +119,81 @@ protected void onSpanFinished(MockSpan mockSpan) {
...
+                        spanId = objStream.readLong();
+                        traceId = objStream.readLong();
+
+                        while (objStream.available() > 0) {
+                            baggage.put(objStream.readUTF(), objStream.readUTF());
+                        }
+                    } catch (IOException e) {
+                    } finally {
+                        if (objStream != null) {
+                            try { objStream.close(); } catch (Exception e2) {}
```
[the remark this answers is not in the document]
carlosalberto: I think it's not important here - this is closing an object used within this method, and it's more of a cleanup task.

opentracing-api/src/main/java/io/opentracing/propagation/BinaryAdapter.java
carlosalberto: Will update 👍

opentracing-api/src/main/java/io/opentracing/propagation/Binary.java
```diff
@@ -0,0 +1,51 @@
...
+ * and it is an error to call write().
+ *
+ * @see Format.Builtin#BINARY
+ * @see io.opentracing.Tracer#inject(SpanContext, Format, Object)
+ * @see io.opentracing.Tracer#extract(Format, Object)
+ */
+public interface Binary {
+    /**
+     * Writes a sequence of bytes to this carrier from the given buffer.
+     * The internal buffer is expected to grow as more data is written.
```
yurishkuro: what does this line refer to? Whose internal buffer? Seems we can remove it without loss.

opentracing-mock/src/main/java/io/opentracing/mock/MockTracer.java
```diff
@@ -112,6 +119,81 @@ protected void onSpanFinished(MockSpan mockSpan) {
...
+
+                    } catch (IOException e) {
+                        throw new RuntimeException("Corrupted state");
+                    } finally {
+                        if (objStream != null) {
+                            try { objStream.close(); } catch (Exception e2) {}
+                        }
+                    }
+                } else {
+                    throw new IllegalArgumentException("Unknown carrier");
```
yurishkuro: can do this at the top to reduce nesting

carlosalberto: Oh, was following the style the `TEXT_MAP` propagator has :) Anyhow, I can definitely change it to be top level as well.

opentracing-api/src/main/java/io/opentracing/propagation/Binary.java
carlosalberto: Internal storage, basically (which would mean that many writes are possible). So based on your comment, it sounds like it definitely needs improvement (either by mentioning that multiple writes are allowed, or removing this line altogether).

tylerbenson:
@carlosalberto I haven't had time to review the whole discussion yet, but does this fully incorporate @raphw's feedback from before?

carlosalberto:
@tylerbenson Some of it, but we are still not taking into account two things that were requested:

1) Not telling in advance the output size (`Tracer.inject()`)
2) Not over-the-wire support, but instead, in-memory buffering is expected (just like `HttpHeaders` and `TextMap`, where you write everything to some in-memory buffer and *after* that you pass it to the transmission and let the protocol handle it).

I'd expect, in turn:

1) Some (or a few) proof of concept implementations.
2) Early/experimental `Tracer`s support and their feedback on this.

raphw:
I can only repeate my critique of this API:

1. **Byte buffers are meant to interact with data either by on or off heap.** This makes byte buffers useful if a server can allow the Tracer to directly serialize or deserialize data from for example a TCP buffer without the JVM needing to also allocate these byte on the Java heap what is expensive in terms of GC. This idea is not reflected by this API proposal and only puts additional weight onto the tracer as most implementations [will look like in this example](https://github.com/opentracing/opentracing-java/pull/252/files#diff-f1dfccca05f6e717c95eca15a8c6622eR142). A good, NIO-friendly API must therefore allow providing `ByteBuffer`s from outside of the tracer.

2. **Byte Buffers are typically used in the context of asynchrous I/O**. An important factor with NIO is that an external buffer such a native TCP storage might not be able to hold all the bytes that belong to the baggage information at once. If I, from the outside, can provide you a buffer, I cannot normally guarantee you that I have all data ready. I would then ask the tracer to do a partial deserialization/serialization, suspend the current channel and handle the next channel from the very same thread. This is the core idea of NIO selectors. With OpenTracing being a synchronous API that does not allow such partial deserialization, I would need to copy the data into a continous heap array and provide this array wrapped by a byte buffer to support the general case. The current API is therefore currently equivalent to one that is operating on byte arrays.

3. **There is no binary contract implied**. Binary protocols that allow embedding random bytes typically add a marker byte to indicate the start and end of a custom segment. This way, a server can accept baggage data from a client without even having a tracer installed by just cropping the segment. By allowing to write to byte buffers explicitly, the user of a tracer has no chance to escape any bytes that resemble the marker bytes breaking the protocol. The only way to work around this would be to write to an intermediate buffer what is breaking the core assumption of NIO.

None of these problems occurs by offering a stream-based approach:

1. Streams can read and write one byte at a time what avoids additional allocation of byte arrays. (Streams can be directed to write or read to or from a direct byte buffer without the mentioned need for intermediaries.)
2. Streams can easily be decorated, allowing for escaping of mentioned marker bytes.
3. **But** streams are synchronous. But since this is also true for the baggage injection and extraction APIs, this is not a problem here.

tylerbenson:
@raphw What would you then suggest?  Maybe I'm missing something, but writing instrumentation to support propagation over a generic binary interface without breaking the protocol when only one side is instrumented seems to be a significant challenge.  Perhaps it would help if we had some specific use cases.  What protocols/frameworks are expected to use this binary format? (perhaps protobuf, thrift, grpc?)  Maybe it would be better to define those propagation interfaces specifically like we do for http, instead of lumping everything into a binary format?  Which formats/protocols are more common when working with NIO?

(Forgive me for my ignorance.  I've written a fair bit of instrumentation, but relatively little for systems without generic headers/metadata/parameters that can be used for propagation details.)

tedsuo:
@raphw @tylerbenson I'd love to see the requirements for binary format expressed as a set of tests/examples. This worked very well for designing context propagation, where the requirements were similarly nuanced. If we keep discussing in english, we'll probably never resolve this. :)

raphw:
Binary Protocols typically include such optional sections either by:

1. Disclosing the size of the segment with some meta data, e.g. by a fixed-length name and a fixed-length size indication followed by the amount of bytes disclosed. This way, if the server recieves this segment but does not know a dispatcher for such a named area, it just discards the bytes.
2. Starting the segment by a name and by indicating the end of the section by a marker byte. All bytes in the payload that would indicate the same byte as this marker byte are then escaped. E.g. any null byte is replaced by two null bytes within the payload. If a null byte is not followed by another null byte, it describes the end of the segment. Again, if no dispatcher is found, the payload can be discarded.

The first option requires to know the binary size in advance, the second option requires a way to manipulate any written byte before it is written to a buffer. My original suggestion using streams would allow for that.

carlosalberto:
Hery everybody - sorry for the delay. Recently I took up on testing the current `BINARY` proposal to do tracing for Cassandra on the server side (through a plugin), and the experimental code consuming this specific `BINARY` support exists in https://github.com/carlosalberto/java-cassandra-server

Please feel free to test it out. Cassandra itself takes a `ByteBuffer` as payload for outgoing query requests, and receives a `ByteBuffer` from the server side as well (described in the mentioned repository containing the Cassandra plugin).

I have the impression most of the times, as the `SpanContext` tends to be small, the payload will be buffered and transferred in a single step.

That being said, I do realize that @raphw usage may not be supported with the current approach, and that may need an additional format (maybe something like `BINARY_STREAMING`).

Let me know ;)

@tedsuo @tylerbenson @yurishkuro

raphw:
the more I think of it, I believe that it would actually be the easiest to extract a `TextMap` and push it into the stream depending on the underlying binary protocol. Until there is an actual use case, maybe it makes sense to table the binary extraction?

carlosalberto:
@raphw Thanks for the feedback. Well, I'd say using OT for instrumenting Cassandra on the server side is already an actual case (albeit a simple one ;) ).

@yurishkuro Any opinion on this?

yurishkuro:
I agree that this is not a pressing matter (strictly speaking "binary" can be just a text_map serialized in a way specific to the protocol being instrumented, like Cassandra), but would be nice to resolve.

carlosalberto:
Closing this PR in favor of #276 (which was just merged).

PR #262, opened 2018-03-21
Deprecate the StringTag.set() overload taking a StringTag.

### LINKED ISSUE #229: StringTag.Set(Span, StringTag) bug
Was this supposed to set the key to the key?
```
    public void set(Span span, StringTag tag) {
        span.setTag(super.key, tag.key);
    }
```

carlosalberto:
It's not a common scenario to use a Tag *key* as the tag value. Would fix #229

### DIFF
--- a/opentracing-api/src/main/java/io/opentracing/tag/StringTag.java
+++ b/opentracing-api/src/main/java/io/opentracing/tag/StringTag.java
@@ -25,6 +25,10 @@ public class StringTag extends AbstractTag<String> {
         span.setTag(super.key, tagValue);
     }
 
+    /**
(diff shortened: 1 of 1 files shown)


carlosalberto:
Merged as it's a rather simple (and obvious) change. Let me know if this needs to be undone later ;)

cwe1ss:
@carlosalberto fyi: this has been merged into the `v0.32.0` branch but there have also already been merges into the `master` branch (#263, #264).

carlosalberto:
@cwe1ss Hey, yes, was wondering about that yesterday. I will back port this deprecation, as once as I get my mind on whats the best (direct cherry-pick or PR ;) )

PR #263, opened 2018-03-28
Enabling NOOP classes to work with pre-existing production code

natehart:
In special circumstances, code may automatically generate spans,
negating the need to always check if `Tracer.activeSpan()` is `null`. Under these
circumstances, the Noop Tracer and ScopeManager will cause NPEs when they return
`null` values when their `active` methods are accessed via the GlobalTracer. This
prevents code from running successfully when using Noop Tracers.

This minor change allows users to depend that their code will genuinely Noop
under all use cases.

Signed-off-by: Nate Hart <nhart@tableau.com>

### DIFF
--- a/opentracing-noop/src/main/java/io/opentracing/noop/NoopScopeManager.java
+++ b/opentracing-noop/src/main/java/io/opentracing/noop/NoopScopeManager.java
@@ -36,7 +36,7 @@ class NoopScopeManagerImpl implements NoopScopeManager {
 
     @Override
     public Scope active() {
-        return null;
+        return NoopScope.INSTANCE;
     }
 
     static class NoopScopeImpl implements NoopScopeManager.NoopScope {
--- a/opentracing-noop/src/main/java/io/opentracing/noop/NoopTracer.java
+++ b/opentracing-noop/src/main/java/io/opentracing/noop/NoopTracer.java
@@ -32,7 +32,7 @@ final class NoopTracerImpl implements NoopTracer {
 
     @Override
     public Span activeSpan() {
-        return null;
+        return NoopSpanImpl.INSTANCE;
     }
 
     @Override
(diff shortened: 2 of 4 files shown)


natehart:
@yurishkuro I made a minor change to the MockTracer to work with the type assumptions baked into the tests. Not only does this allow the tests to propagate MockSpans as expected, but it brings the Propagator constructor closer into line with the no-argument constructor's default values. You may want to put eyes on it before I merge in case I changed a very intentional design decision.

opentracing-mock/src/main/java/io/opentracing/mock/MockTracer.java
```diff
@@ -59,7 +59,7 @@ public MockTracer(ScopeManager scopeManager, Propagator propagator) {
      * Create a new MockTracer that passes through any calls to inject() and/or extract().
      */
     public MockTracer(Propagator propagator) {
-        this(NoopScopeManager.INSTANCE, propagator);
+        this(new ThreadLocalScopeManager(), propagator);
```
yurishkuro: can this be done as another PR? It's completely unrelated.

natehart: Sure, but we'll have to merge that one first.

The MockTracer assumes that `ScopeManager.active()` either returns null or a scope that with a MockSpan. Since we are now returning a NoopScope (and thus a NoopSpan) from the NoopScopeManager, it tries to cast the NoopSpan to a MockSpan and blows up. We won't be able to get the Mock tests to pass until we remove the NoopScopeManager from the Mocks.

natehart:
@yurishkuro I've done some digging into why code coverage is dropping so drastically, but I can't figure out where the problem is. I bet there is some null check in a test somewhere that was assuming a null return, but I'll be darned if I can figure out where it is. Unfortunately, I don't have any more time to spend on this. Merge it if you think it's worthwhile, otherwise you can close the PR.

PR #264, opened 2018-03-28
Updating the Mock Tests To Use Proper Scope Manager

natehart:
Currently, the no-argument constructor for the MockTracer uses a `ThreadLocalScopeManager`
and a `Propagator.TEXT_MAP` as default values. But when providing a Propagator,
the constructor injects a `NoopScopeManager` instead of the `ThreadLocalScopeManager`.

This commit brings the Propagator constructor closer into line with the default
values for this mock.

cc: @yurishkuro

Signed-off-by: Nate Hart <nhart@tableau.com>


yurishkuro:
LGTM

natehart:
@yurishkuro Sorry, I don't have write access to merge.

PR #268, opened 2018-04-11
Add release instructions to update change log

MikeGoldsmith:
Added instructions to RELEASE.md that describe required changes for CHANGELOG.md during release process.


RELEASE.md
```diff
@@ -17,6 +17,11 @@ This repo uses semantic versions. Please keep this in mind when choosing version
    This part is controlled by [`travis/publish.sh`](travis/publish.sh). It creates a bunch of new commits, bumps
    the version, publishes artifacts, and syncs to Maven Central.
 
+1. **Update Change Log**
+
+   Update CHANGELOG.md with a list changes since the last release. Each entry must include the release number,
+   date and a bulleted list of changes where each change is summarized in a single sentence.
+
```
yurishkuro: typically this is done before the release. In Jaeger we use the convention that a commit titled "Preparing release x.y.z" contains changes to the changelog, and it's that commit that gets tagged to trigger a release. Then it follows by another commit "Back to development" that bumps the version to n+1.(snapshot|dev) and adds an empty entry to the changelog. Mostly we adopted this after using `zest.releaser` in Python, that automates all those steps.

MikeGoldsmith: Maybe we could look at automating those steps too?

I'll just update the release process for now though as it's a larger task to investigate / complete that work.

yurishkuro: yeah, I wish there was a single cross-language tool for that

MikeGoldsmith:
@yurishkuro I've moved the change log stage to after alerting there is a release happening and before creating a tag commit.

Issue #186, opened 2017-09-22
Make GlobalTracer uses more threadsafe


redijedi:
I have a multithreaded app which causes some unpredictable behaviors when trying to get a globally registered tracer. A specific complication appears to have some relationship to byteman when using byteman rules. I have found that sometimes I call `GlobalTracer.get()` and receive a noop, while other times I get a tracer initialized from some byteman execution. I attempted to resolve this by conditionally registering the tracer myself:
```
        Tracer tracer;
        if (GlobalTracer.isRegistered()) {
            tracer = GlobalTracer.get();
        } else {
            tracer = TracerResolver.resolveTracer();
            GlobalTracer.register(tracer);
        }
```
However, this was affected by a race condition somewhere. I was able to resolve it by synchronizing on the `GlobalTracer.class`:
```
        Tracer tracer;
        synchronized (GlobalTracer.class) {
            if (GlobalTracer.isRegistered()) {
                tracer = GlobalTracer.get();
            } else {
                tracer = TracerResolver.resolveTracer();
                GlobalTracer.register(tracer);
            }
        }
```

but this seems like something that would be nice to incorporate into the API itself. Note that this still causes the agent to log an exception:
```
! java.lang.IllegalStateException: There is already a current global Tracer registered.
! at io.opentracing.util.GlobalTracer.register(GlobalTracer.java:92)
! at io.opentracing.contrib.agent.OpenTracingHelper.initTracer(OpenTracingHelper.java:80)
````
but it appears to work anyway.

I'll submit a method proposal for streamlining this shortly.

redijedi:
It's rough, but here's the general idea: https://github.com/redijedi/opentracing-java/commit/42faf243bfb4d4a59c2ff6e5cb54a73cd5d376d7

wu-sheng:
@redijedi By using either  `public static synchronized Tracer getOrRegister(...)` and `synchronized (GlobalTracer.class)`, I can resolve the race condition, but you are facing the performance problem. Every time you try to get the tracer, you must get the Lock first, one way or another.

I prefer the current API. And for your circumstance, you can try to initial/resolver/register the GlobalTracer first. For a multithreaded app, I think you need multi threads in providing service or business logic, but you can use single thread to process the tracer initialization stage.

redijedi:
[...]

PR #269, opened 2018-04-13
Make GlobalTracer uses more threadsafe

sjoerdtalsma:
As discussed in #186 replace `void register(Tracer)` with a more
atomic `boolean registerIfAbsent(Tracer)` method.


opentracing-util/src/main/java/io/opentracing/util/GlobalTracer.java
```diff
@@ -84,43 +83,88 @@ public static Tracer get() {
...
-        }
-        if (tracer instanceof GlobalTracer) {
-            LOGGER.log(Level.FINE, "Attempted to register the GlobalTracer as delegate of itself.");
-            return; // no-op
+    public static synchronized boolean registerIfAbsent(final TracerSupplier supplier) {
+        if (supplier != null && !isRegistered()) {
+            final Tracer suppliedTracer = supplier.get();
+            if (suppliedTracer != null && !(suppliedTracer instanceof GlobalTracer)) {
+                GlobalTracer.tracer = suppliedTracer;
+                return true;
```
yurishkuro: what is the point of null checks? This being a configuration step I'd rather see NPE than silent success.

sjoerdtalsma: The null checks are there because the new API returns `true` for registration and `false` for _no registration_, which seems a valid way to handle this.
Just my opinion though, there is something to be said for throwing NPE's as well.

yurishkuro: true/false for registration or no registration is fine, but null supplier or tracer are violations of the method's API, imo, it's better to report them than work around them.

sjoerdtalsma: Valid point. I'd want a NPE _with_ a message what was `null` though. Unfortunately we don't have `Objects.requireNonNull` in java 6 yet..

yurishkuro: `if (supplier == null) throw new NullPointerException("supplier must not be null");` no bueno?

opentracing-util/src/main/java/io/opentracing/util/TracerSupplier.java
```diff
@@ -0,0 +1,30 @@
...
+ * the License.
+ */
+package io.opentracing.util;
+
+import io.opentracing.Tracer;
+
+/**
+ * Functional interface for {@code Supplier<Tracer>} as the API is still Java 1.6+ compatible.
+ */
+public interface TracerSupplier {
```
pavolloffay: Is this interface required to resolve #186 ?

opentracing-util/src/main/java/io/opentracing/util/GlobalTracer.java
```diff
@@ -84,43 +83,86 @@ public static Tracer get() {
...
+     * <p>
+     * Every application intending to use the global tracer is responsible for registering it once
+     * during its initialization.
+     *
+     * @param tracer Tracer to use as global tracer.
+     * @throws RuntimeException if there is already a current tracer registered
+     * @see #registerIfAbsent(Tracer)
+     * @deprecated Please use 'registerIfAbsent' instead which does not attempt a double registration.
+     */
+    public static void register(final Tracer tracer) {
```
pavolloffay: If reprecating add `@Deprecated`

opentracing-util/src/main/java/io/opentracing/util/TracerSupplier.java
sjoerdtalsma: If you want to use it with lambda's while maintaining jdk 1.6 compatibility, sadly I think it is.

pavolloffay: Making it "lambda" like is not required to resolve #186. If there is no requirement for it I am -1 on adding this interface.

sjoerdtalsma: Ok, noted. However obtaining and configuring a tracer is potentially expensive / complex behaviour that may not be necessary. This is -in my opinion- a good enough reason to introduce the method _here_.
How do we reach consensus on this one; let people vote?

yurishkuro: I would actually be in favor of **only** accepting TracerSupplier (btw, isn't "provider" a more accepted term?), because the semantics of "if absent" imply the tracer may not be needed and accepting an actual tracer instance in that case is promoting a bad pattern of constructing an (expensive) instance that may not be needed. E.g. the correct behavior of the caller if we allow Tracer would be to do something like
```
Tracer tracer  = ... // create
if (!GlobalTracer.registerIfAbsent(tracer)) {
  tracer.close();  // release resources
}
```
With "provider" pattern the above is not necessary.

pavolloffay: I think you are talking about lazy tracer initialization. Then it makes sense

yurishkuro: btw we can still keep the resolver use case as one liner if it implements a new method
```
GlobalTracer.registerIfAbsent(Tracerresolver.provider());
```

sjoerdtalsma: I have no problems removing the `registerIfAbsent(Tracer)` method, as it can be replaced with the trivial supplier inner-class (as the current delegated implementation) or a lambda in somewhat decent jvm version.

sjoerdtalsma: @yurishkuro I thought of `provider`, but chose `TracerSupplier` as the Java 8 functional interface would be a `Supplier<Tracer>`. This is also why I chose the name for the `get()` method: That's what the `Supplier<T>` has in java 8.

sjoerdtalsma: I'm not very biased to either choice so if somebody has strong feelings about the one or the other, that's okay by me.

sjoerdtalsma: @pavolloffay @yurishkuro Re: `TracerSupplier`
>Is this interface required to resolve #186 ?

Come to think of it, technically, I think we could also get away with `Callable<Tracer>`, which is available in Java 6. The trivial new interface doesn't really bother me, but perhaps you have strong opinions on this?

Using a `Callable<Tracer>` as provider does have a slight disadvantage that we will have to catch and handle the thrown `Exception`

yurishkuro:
Callable sgtm

opentracing-util/src/main/java/io/opentracing/util/TracerSupplier.java
sjoerdtalsma: Interface `TracerSupplier` has been replaced by `Callable<Tracer>`

sjoerdtalsma:
@pavolloffay Have your comments been addressed?

sjoerdtalsma:
@pavolloffay @yurishkuro Are there any remaining concerns regarding this PR?

Otherwise it may be a good idea to merge this soon, to allow people to adapt to the deprecation as soon as possible.

yurishkuro:
lgtm, 🚢

opentracing-util/src/main/java/io/opentracing/util/GlobalTracer.java
```diff
@@ -84,43 +82,70 @@ public static Tracer get() {
...
+     * @throws RuntimeException if there is already a current tracer registered
+     * @see #registerIfAbsent(Callable)
+     * @deprecated Please use 'registerIfAbsent' instead which does not attempt a double registration.
      */
-    public static synchronized boolean isRegistered() {
-        return !(GlobalTracer.tracer instanceof NoopTracer);
+    @Deprecated
+    public static void register(final Tracer tracer) {
+        if (!registerIfAbsent(provide(tracer))
+                && !tracer.equals(GlobalTracer.tracer)
```
pavolloffay: I am not sure if these checks are mandatory here. `registerIfAbsent(provide(tracer)` is called before and should contain the same checks and throw the same exception.

opentracing-util/src/main/java/io/opentracing/util/GlobalTracer.java
```diff
@@ -84,43 +82,70 @@ public static Tracer get() {
...
-     * Registration is a one-time operation, attempting to call it more often will result in a runtime exception.
+     * The tracer is provided through a {@linkplain Callable} that will only be called if the global tracer is absent.
+     * Registration is a one-time operation. Once a tracer has been registered, all attempts at re-registering
+     * will return {@code false}.
      * <p>
      * Every application intending to use the global tracer is responsible for registering it once
      * during its initialization.
      *
-     * @param tracer Tracer to use as global tracer.
-     * @throws RuntimeException if there is already a current tracer registered
```
pavolloffay: You should document exceptions thrown by this method

opentracing-util/src/main/java/io/opentracing/util/GlobalTracer.java
sjoerdtalsma: @pavolloffay These checks are necessary to keep the current behaviour backward-compatible.

This method is deprecated and must be removed in a future version. Until then, the error-behaviour must not change.

opentracing-util/src/main/java/io/opentracing/util/GlobalTracer.java
```diff
@@ -84,43 +82,73 @@ public static Tracer get() {
...
-        }
-        if (tracer instanceof GlobalTracer) {
-            LOGGER.log(Level.FINE, "Attempted to register the GlobalTracer as delegate of itself.");
-            return; // no-op
-        }
-        if (isRegistered() && !GlobalTracer.tracer.equals(tracer)) {
-            throw new IllegalStateException("There is already a current global Tracer registered.");
+    public static synchronized boolean registerIfAbsent(final Callable<Tracer> provider) {
+        requireNonNull(provider, "Cannot register GlobalTracer from provider <null>.");
+        if (!isRegistered()) try {
```
wu-sheng: This is not a common style in Java, even it is legal
```
 if (!isRegistered()) try {
```
Please consider
```
 if (!isRegistered()) {
     try {
     }
}
```

sjoerdtalsma: Personally (matter of taste) I find the intention better communicated to keep " if _condition_ try " on a single line.

But I'll change it. Merging the deprecation soon is more important for me.

wu-sheng: You could keep your habit :)  It is a suggestion only.

sjoerdtalsma: @wu-sheng I will in my own projects. 
Since this is a public API however, I agree that any uncommon style should be avoided when possible. :+1:

opentracing-util/src/test/java/io/opentracing/util/GlobalTracerTest.java
```diff
@@ -58,7 +48,9 @@ public void testGet_SingletonReference() {
 
     @Test
     public void testMultipleRegistrations() {
-        GlobalTracer.register(mock(Tracer.class));
+        assertThat(GlobalTracer.registerIfAbsent(provide(mock(Tracer.class))), is(true));
```
pavolloffay: @sjoerdtalsma to be sure could you please keep old tests for `register` method? And add new for new?

sjoerdtalsma:
The last activity on this PR is almost a month ago with three approvals given already.
@pavolloffay do you have any remaining concerns regarding this PR?

Otherwise I suggest we merge this soon, so people can get used to the deprecated method and its replacement.

pavolloffay:
@sjoerdtalsma sorry for long delay, thanks for the PR!

pavolloffay:
I think we could/should still keep `GlobalTracer.register(tracer)` or `registerIfAbsent(Tracer)` for simpler use cases. Integrations in runtimes usually controls tracer initialization.

pavolloffay:
@sjoerdtalsma do you want to tackle it in this PR? If not let's merge this and do it sepratetely.

pavolloffay:
I am also a bit concerned about callable. Maybe a separate interface would be better. Callable somehow indicates use of executors which is not the case here.

sjoerdtalsma:
@pavolloffay 
>I think we could/should still keep GlobalTracer.register(tracer) or registerIfAbsent(Tracer) for simpler use cases. Integrations in runtimes usually controls tracer initialization.

I'm not sure I understand what you mean here. `registerIfAbsent` is a _replacement_ of `register`. Do you know of a use-case for `register` that isn't possible with `registerIfAbsent`? It is entirely possible that I misunderstood your comment.

> I am also a bit concerned about callable. Maybe a separate interface would be better. Callable somehow indicates use of executors which is not the case here.

My ideal would be to use a `Supplier<Tracer>` but that doesn't exist in our java version yet.  
I actually **started** with a separate `TracerSupplier` interface but [you asked](https://github.com/opentracing/opentracing-java/pull/269#discussion_r181664552) whether it was necessary. Discussion and agreement amounted to just using `Callable` and wrap the checked exceptions as the most simple solution and not worthy of a separate interface just for this purpose.

I'm not against re-introducing the separate interface again, but maybe we're running in circles on this one?

sjoerdtalsma:
Maybe someone should just merge this in and if there are many questions re: Callable, replace it by TracerSupplier at that time?

pavolloffay:
@sjoerdtalsma I mean we should also keep simple `registerIfAbsent(Tracer tracer)` without callable. Usually, runtimes control tracer initialization so there is no need for callable!

sjoerdtalsma:
@pavolloffay 
> This can go in `registerIfAbsent(Tracer tracer)` can be added in a separate PR

It may be good to discuss the `Tracer` variant in a separate PR indeed, re: @yurishkuro's comment here: https://github.com/opentracing/opentracing-java/pull/269#discussion_r181815569

Thanks everyone for their review, could someone with merge rights merge this then? :wink:

pavolloffay:
@sjoerdtalsma think does not work for me, I will merge this now

PR #270, opened 2018-04-24
Fix try/catch/finally example

felixbarny:
It actually compiles now and also includes how to log an exception


yurishkuro:
thanks!

Issue #291, opened 2018-07-23
Remove finishSpanOnClose flag


felixbarny:
This is a follow up on https://github.com/opentracing/opentracing-java/issues/267. The underlying issue of that one is that you can't remember the `finishSpanOnClose` flag when implementing a bridge for `ScopeManager#active()`. 

The point of this issue is to remove the `finishSpanOnClose` flag altogether, as it is only useful in try-with-resources constructs which are a bad practice because you can't can't log exceptions as the span is closed before the catch block. To cite the main [README.md](https://github.com/opentracing/opentracing-java#starting-a-new-span):

> Note that startActive(true) finishes the span on Scope.close(). Use it carefully because the try-with-resources construct finishes the span before the catch or finally blocks are executed, which makes logging exceptions and setting tags impossible. It is recommended to start the span and activate it later in try-with-resources. This makes the span available in catch and finally blocks.

So we are basically officially admitting that `finishSpanOnClose` is a bad practice. 

You could argue that try-with-resources is valid if you don't want to log exceptions. But what if a user then later decides that they do want to log exceptions? Do they still remember that they can't use try-with-resources + a catch block then? I would most likely forget about that.

The `finishSpanOnClose` just enables some syntactic sugar after all which turns out to be problematic. As flag essentially advocates a bad practice, I suggest removing it.

Transition:
Of course, we can't just remove the flag which would introduce a breaking change.

Instead, I propose adding the method `io.opentracing.Tracer.SpanBuilder#startActive()` without arguments, which has the same semantics like calling `spanBuilder.startActive(false)`. Also, deprecate `io.opentracing.Tracer.SpanBuilder#startActive(boolean finishSpanOnClose)`. This method can then be removed on a later release (for example `1.0.0-RC1`)

The `Scope` interface can still extend from `Closeable`, because logging an exception is still valid to do after the scope is closed.

I think, even after doing this, https://github.com/opentracing/opentracing-java/issues/267 (Deprecate or remove `ScopeManager.active()`) is still relevant, but maybe less severe.

felixbarny:
See also this brave issue for more information about why try-with-resources is considered an anti-pattern: https://github.com/openzipkin/brave/issues/676

adriancole:
note this antipattern is also well documented in wingtips too. There's significant documentation burden to explain how to overcome it
https://github.com/Nike-Inc/wingtips#warning-about-error-handling-when-using-try-with-resources-to-autoclose-spans

[...]

PR #301, opened 2018-08-26
Remove finish span on close

carlosalberto:
Addresses #291 

`SpanBuilder.startActive(Span, boolean)` and `ScopeManager.activate(Span, bool)` have been kept, but as deprecated methods.

Updated the `opentracing-testbed` code to use this change, and added a new, small, `error-reporting` case.

cc @felixbarny @adriancole @yurishkuro @tedsuo @sjoerdtalsma

### DIFF
--- a/opentracing-api/src/main/java/io/opentracing/ScopeManager.java
+++ b/opentracing-api/src/main/java/io/opentracing/ScopeManager.java
@@ -26,23 +26,37 @@ import io.opentracing.Tracer.SpanBuilder;
 public interface ScopeManager {
 
     /**
-     * @deprecated use {@link #activate} instead.
-     * Make a {@link Span} instance active.
+     * Set the specified {@link Span} as the active instance for the current
+     * context (usually a thread).
      *
-     * @param span the {@link Span} that should become the {@link #active()}
-     * @param finishSpanOnClose whether span should automatically be finished when {@link Scope#close()} is called
-     * @return a {@link Scope} instance to control the end of the active period for the {@link Span}. It is a
-     * programming error to neglect to call {@link Scope#close()} on the returned instance.
-     */
-    @Deprecated
-    Scope activate(Span span, boolean finishSpanOnClose);
-
-    /**
-     * Make a {@link Span} instance active.
+     * <p>
+     * The returned {@link Scope} represents the active state for the span.
+     * Once its active period is due, {@link Scope#close()} ought to be called.
+     * To ease this operation, {@link Scope} supports try-with-resources.
+     * Observe the span will not be automatically finished when {@link Scope#close()}
+     * is called.
      *
-     * The span will not be automatically finished when {@link Scope#close()} is called.
+     * <p>
+     * This {@link Scope} instance can be accessed at any time through {@link #active()},
+     * in case it is not possible for the user to store it (when used through middleware
+     * or start/finish event hooks, for example). The corresponding {@link Span} can be
+     * accessed through {@link #activeSpan()} likewise.
      *
-     * @param span the {@link Span} that should become the {@link #active()}
+     * <p>
+     * Usage:
+     * <pre><code>
+     *     Span span = tracer.buildSpan("...").start();
+     *     try (Scope scope = tracer.scopeManager().activate(span)) {
+     *         span.setTag("...", "...");
+     *         ...
+     *     } catch (Exception e) {
+     *         span.log(...);
+     *     } finally {
+     *         span.finish(); // Optionally finish the Span.
+     *     }
+     * </code></pre>
+     *
+     * @param span the {@link Span} that should become the {@link #activeSpan()}
      * @return a {@link Scope} instance to control the end of the active period for the {@link Span}. It is a
      * programming error to neglect to call {@link Scope#close()} on the returned instance.
      */
@@ -53,10 +67,12 @@ public interface ScopeManager {
      * {@link Span}.
      *
      * <p>
-     * If there is an {@link Scope non-null scope}, its wrapped {@link Span} becomes an implicit parent
-     * (as {@link References#CHILD_OF} reference) of any
-     * newly-created {@link Span} at {@link Tracer.SpanBuilder#startActive(boolean)} or {@link SpanBuilder#start()}
-     * time rather than at {@link Tracer#buildSpan(String)} time.
+     * Observe that {@link Scope} is expected to be used only in the same thread where it was
+     * created, and thus should not be passed across threads.
+     *
+     * <p>
+     * Because both {@link #active()} and {@link #activeSpan()} reference the current
+     * the active state, they both will be either null or non-null.
      *
      * @return the {@link Scope active scope}, or null if none could be found.
      */
@@ -68,10 +84,38 @@ public interface ScopeManager {
      * <p>
      * If there is an {@link Span non-null active span}, it becomes an implicit parent
      * (as {@link References#CHILD_OF} reference) of any
-     * newly-created {@link Span} at {@link Tracer.SpanBuilder#startActive(boolean)} or {@link SpanBuilder#start()}
+     * newly-created {@link Span} at {@link SpanBuilder#start()} or {@link Tracer.SpanBuilder#startActive(boolean)}
      * time rather than at {@link Tracer#buildSpan(String)} time.
+     *
+     * <p>
+     * Because both {@link #active()} and {@link #activeSpan()} reference the current
+     * the active state, they both will be either null or non-null.
      *
      * @return the {@link Span active span}, or null if none could be found.
      */
     Span activeSpan();
+
+    /**
+     * @deprecated use {@link #activate(Span)} instead.
+     * Set the specified {@link Span} as the active instance for the current
+     * context (usually a thread).
+     *
+     * <p>
+     * Finishing the {@link Span} upon {@link Scope#close()} is discouraged,
+     * as reporting errors becomes impossible:
+     * <pre><code>
+     *     try (Scope scope = tracer.scopeManager().activate(span, true)) {
+     *     } catch (Exception e) {
+     *         // Not possible to report errors, as
+     *         // the span has been already finished.
+     *     }
+     * </code></pre>
+     *
+     * @param span the {@link Span} that should become the {@link #activeSpan()}
+     * @param finishSpanOnClose whether span should automatically be finished when {@link Scope#close()} is called
+     * @return a {@link Scope} instance to control the end of the active period for the {@link Span}. It is a
+     * programming error to neglect to call {@link Scope#close()} on the returned instance.
+     */
+    @Deprecated
+    Scope activate(Span span, boolean finishSpanOnClose);
 }
--- a/opentracing-api/src/main/java/io/opentracing/ScopeManager.java
+++ b/opentracing-api/src/main/java/io/opentracing/ScopeManager.java
@@ -52,7 +52,9 @@ public interface ScopeManager {
      *     } catch (Exception e) {
      *         span.log(...);
      *     } finally {
-     *         span.finish(); // Optionally finish the Span.
+     *         // Optionally finish the Span if the operation it represents
+     *         // is logically completed at this point.
+     *         span.finish();
      *     }
      * </code></pre>
      *
@@ -72,7 +74,7 @@ public interface ScopeManager {
      *
      * <p>
      * Because both {@link #active()} and {@link #activeSpan()} reference the current
-     * the active state, they both will be either null or non-null.
+     * active state, they both will be either null or non-null.
      *
      * @return the {@link Scope active scope}, or null if none could be found.
      */
@@ -82,14 +84,8 @@ public interface ScopeManager {
      * Return the currently active {@link Span}.
      *
      * <p>
-     * If there is an {@link Span non-null active span}, it becomes an implicit parent
-     * (as {@link References#CHILD_OF} reference) of any
-     * newly-created {@link Span} at {@link SpanBuilder#start()} or {@link Tracer.SpanBuilder#startActive(boolean)}
-     * time rather than at {@link Tracer#buildSpan(String)} time.
-     *
-     * <p>
      * Because both {@link #active()} and {@link #activeSpan()} reference the current
-     * the active state, they both will be either null or non-null.
+     * active state, they both will be either null or non-null.
      *
      * @return the {@link Span active span}, or null if none could be found.
      */
--- a/opentracing-api/src/main/java/io/opentracing/ScopeManager.java
+++ b/opentracing-api/src/main/java/io/opentracing/ScopeManager.java
@@ -49,8 +49,8 @@ public interface ScopeManager {
     Scope activate(Span span);
 
     /**
-     * Return the currently active {@link Scope} which can be used to access the currently active
-     * {@link Scope#span()}.
+     * Return the currently active {@link Scope} which can be used to deactivate the currently active
+     * {@link Span}.
      *
      * <p>
      * If there is an {@link Scope non-null scope}, its wrapped {@link Span} becomes an implicit parent
@@ -61,4 +61,17 @@ public interface ScopeManager {
      * @return the {@link Scope active scope}, or null if none could be found.
      */
     Scope active();
+
+    /**
+     * Return the currently active {@link Span}.
+     *
+     * <p>
+     * If there is an {@link Span non-null active span}, it becomes an implicit parent
+     * (as {@link References#CHILD_OF} reference) of any
+     * newly-created {@link Span} at {@link Tracer.SpanBuilder#startActive(boolean)} or {@link SpanBuilder#start()}
+     * time rather than at {@link Tracer#buildSpan(String)} time.
+     *
+     * @return the {@link Span active span}, or null if none could be found.
+     */
+    Span activeSpan();
 }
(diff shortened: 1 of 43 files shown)


opentracing-api/src/main/java/io/opentracing/Tracer.java
```diff
@@ -166,16 +166,40 @@
...
+        @Deprecated
+        Scope startActive(boolean finishSpanOnClose);
+
+        /**
+         * Returns a newly started and activated {@link Scope}.
+         *
+         * The span will not be automatically finished when {@link Scope#close()} is called.
+         *
+         * <p>
+         * The returned {@link Scope} supports try-with-resources, but is usually
```
felixbarny: Why is it not usually used with try-with-resources?

opentracing-api/src/main/java/io/opentracing/Tracer.java
```diff
@@ -166,16 +166,40 @@
...
+         * used like this:
          * <pre><code>
-         *     try (Scope scope = tracer.buildSpan("...").startActive(true)) {
+         *     Scope scope = null;
+         *     try {
+         *         scope = tracer.buildSpan("...").startActive();
          *         // (Do work)
          *         scope.span().setTag( ... );  // etc, etc
+         *     } finally {
```
felixbarny: add catch clause

opentracing-api/src/main/java/io/opentracing/Tracer.java
```diff
@@ -190,16 +214,15 @@
...
+         * {@code tracer.scopeManager().activate(spanBuilder.start())}.
          *
-         * @param finishSpanOnClose whether span should automatically be finished when {@link Scope#close()} is called
          * @return a {@link Scope}, already registered via the {@link ScopeManager}
          *
          * @see ScopeManager
          * @see Scope
          */
-        Scope startActive(boolean finishSpanOnClose);
+        Scope startActive();
```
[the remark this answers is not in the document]
felixbarny: Could you elaborate?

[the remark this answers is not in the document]
carlosalberto: Oh, that crossed my mind too. So I think that, if it stays, we would need to be clear about the implications of its usage - that is needs to be used with a standard `try` block, i.e.

```java
Scope scope = tracer.buildSpan("foo").startActive();
try {
} catch (Exception e) {
  Tags.ERROR.set(scope.span(), true);
} finally {
  scope.span().finish();
  scope.close();
}
```

(That's why I used such usage in its updated inline documentation ;) )

I'd be up for keeping it (with the proper documentation), but definitely understand it's not a perfect thing, and I'm open to have it removed if deemed so (in such case I will update the PR).

opentracing-api/src/main/java/io/opentracing/Tracer.java
carlosalberto: That's related to Yuri's comment about removing it - essentially, that, as opposed to using `ScopeManager.activate()`, here we **need** an usual `try` block to do the cleanup (that is, having a reference to the `Span` to finish it after any error has occurred).

```java
try (Scope scope = tracer.buildSpan("foo").startActive()) {
} catch (Exception e) {
  // scope and scope.span() are not reachable here nor in the finally section :/
}
```

opentracing-api/src/main/java/io/opentracing/Tracer.java
carlosalberto: Will add it 👍

opentracing-api/src/main/java/io/opentracing/ScopeManager.java
```diff
@@ -26,15 +26,28 @@
 public interface ScopeManager {
 
     /**
+     * @deprecated use {@link #activate} instead.
```
adriancole: nit move to the bottom and say when it will be removed (ideally next release as would the other deprecated methods added prior.. deprecation pre 1.0 is a strange thing but anyway)

opentracing-api/src/main/java/io/opentracing/ScopeManager.java
```diff
@@ -26,15 +26,28 @@
...
      * @param span the {@link Span} that should become the {@link #active()}
      * @param finishSpanOnClose whether span should automatically be finished when {@link Scope#close()} is called
      * @return a {@link Scope} instance to control the end of the active period for the {@link Span}. It is a
      * programming error to neglect to call {@link Scope#close()} on the returned instance.
      */
+    @Deprecated
     Scope activate(Span span, boolean finishSpanOnClose);
 
+    /**
+     * Make a {@link Span} instance active.
```
adriancole: add some description of what active means as the word isn't intuitive.

For example "Makes this span visible via commands like (XX) until the corresponding scope is close"

Also, note that it is a programming error to leave the scope unclosed or to attempt to close it on a different thread.

opentracing-api/src/main/java/io/opentracing/Tracer.java
```diff
@@ -166,16 +166,40 @@
         SpanBuilder withStartTimestamp(long microseconds);
 
         /**
+         * @deprecated use {@link #startActive} instead.
```
adriancole: same.. move down

opentracing-api/src/main/java/io/opentracing/Tracer.java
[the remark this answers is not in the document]
adriancole: This is a strange method, please remove it.

opentracing-api/src/main/java/io/opentracing/Tracer.java
adriancole: If this method isn't deleted, I think it would be cleaner if done similar to census and brave's "scoped span" Instead of messing with scope, mess with span.finish(). In other words, return a Span, and do mention it is a programming error to not call span.finish on the same thread. Another way is to leave it out..

Here's the javadoc from brave's ScopedSpan:

```java
  /**
   * Closes the {@link CurrentTraceContext#newScope(TraceContext) scope} associated with this span,
   * then reports the span complete, assigning the most precise duration possible.
   */
  public abstract void finish();
```

opentracing-api/src/main/java/io/opentracing/Tracer.java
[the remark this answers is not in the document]
carlosalberto: Seems we have found agreement on this point, so I will update this PR in the next days to have it removed ;)

opentracing-api/src/main/java/io/opentracing/ScopeManager.java
carlosalberto: I don't think there's a date to remove this one (we should discuss this topic, but another time ;) ), but we could say "in the next releases" (and definitely before 1.0 ;) )

opentracing-api/src/main/java/io/opentracing/ScopeManager.java
carlosalberto: Thanks, will add such explanation.

carlosalberto:
Hey @adriancole 

Thanks for the feedback. So my personal take for your points is:

1. That's not a bad thing, but then people would need to remember to do something like:

```java
Span span = tracer.buildSpan("one").startActive();
try {
} catch (Exception e) {
  ...
  span.finish();
} finally {
   tracer.scopeManager.active().close();
}
```

This being implicit kinda makes me wonder (but definitely thing to consider).

2. A special `Span` is something that would add complexity to the current (relatively simple) API. For example: what happens if this very special `Span` is passed to another thread? I'd leave it as a potential addition in the future.

So the thing about removing `Scope.span()` is probably fine, as long as we have `ScopeManager.activeSpan()` (so we always expose the current active `Span` and the active `Scope` without explicitly linking them together). However, I'd like to hear the opinion of other people here.

Opinions? @felixbarny @yurishkuro @tedsuo

yurishkuro:
are we shooting for this?
```java
Span span = tracer.buildSpan("nasty").start();
try (Scope scope = tracer.activate(span)) {
  // do nasty stuff
} catch (Exception e) {
  span.SetTag(...);
  ...
} finally {
  span.finish();
}
```

It makes the usage slightly more verbose than the current api, especially without try-with-resource (only if <=1.6, who cares), but avoids the misuse. `scope.span()` is removed.

felixbarny:
I agree with @yurishkuro 
* Remove* `SpanBuilder#startActive`
* Remove* `Scope.span()` (this is only useful in combination with `SpanBuilder#startActive`)
* Remove* `ScopeManager.activate(Span, boolean)`

*deprecate and eventually remove

carlosalberto:
Hey,

Was wondering, after the feedback, about this (small variation):

* `SpanBuilder.startActive()` and `Scope.span()` gets removed
* `ScopeManager.activeSpan()` and `ScopeManager.active()` (so we have the two of them available, but *not* linked to each other, thus less chances to have users pass `Scope` objects arounds).
* `Trace.activeSpan()` and `Tracer.activate()` as aliases to `ScopeManager`'s respective methods.

felixbarny:
Sounds fine to me

opentracing-api/src/main/java/io/opentracing/Tracer.java
```diff
@@ -166,16 +166,40 @@
...
+         * The span will not be automatically finished when {@link Scope#close()} is called.
+         *
+         * <p>
+         * The returned {@link Scope} supports try-with-resources, but is usually
+         * used like this:
          * <pre><code>
-         *     try (Scope scope = tracer.buildSpan("...").startActive(true)) {
+         *     Scope scope = null;
+         *     try {
+         *         scope = tracer.buildSpan("...").startActive();
```
adriancole: please inline this before the try. That removes the null checking later.

Also, as mentioned I think the scope having an accessor to the span is problematic.

carlosalberto:
Hey all, 

This PR is updated to reflect the latest requested changes, namely:

* Deprecate `SpanBuilder.startActive(boolean)`, `ScopeManager.activate(Span, boolean)` and `Scope.span()`.
* Removed `SpanBuilder.startActive()` (parameterless overload, which had been added early in this PR).
* Add `ScopeManager.activeSpan()` and leave `ScopeManager.active()` in place (So `ScopeManager.active()` is still reachable, but designed to stay for more advanced use, as well as still exposing `ScopeManager`, in case the user wants to unwrap it himself).
* Add `Tracer.activateSpan()`, as a shorthand for `tracer.scopeManager().activate(span)`.
* Updated the testbed module and inline docs overall.

@yurishkuro @adriancole @felixbarny

opentracing-testbed/src/test/java/io/opentracing/testbed/activate_deactivate/README.md
```diff
@@ -8,7 +8,8 @@ This example shows the usage of the `Continuation` concept (now part of the `Aut
 return new Thread(new Runnable() {
     @Override
     public void run() {
-	try (Scope scope = tracer.buildSpan("parent").startActive(false)) {
+	Span span = tracer.buildSpan("parent").start();
+	try (Scope scope = tracer.activateSpan(span)) {
```
tylerbenson: I thought the span must be manually finished now.  (There are other cases of this same problem elsewhere in this PR too.)

opentracing-api/src/main/java/io/opentracing/ScopeManager.java
```diff
@@ -26,26 +26,96 @@
...
+     * Usage:
+     * <pre><code>
+     *     Span span = tracer.buildSpan("...").start();
+     *     try (Scope scope = tracer.scopeManager().activate(span)) {
+     *         span.setTag("...", "...");
+     *         ...
+     *     } catch (Exception e) {
+     *         span.log(...);
+     *     } finally {
+     *         span.finish(); // Optionally finish the Span.
```
yurishkuro: `// Optionally finish the Span if the operation it represents is logically completed at this point.`

opentracing-api/src/main/java/io/opentracing/ScopeManager.java
```diff
@@ -26,26 +26,96 @@
...
      *
      * @return the {@link Scope active scope}, or null if none could be found.
      */
     Scope active();
+
+    /**
+     * Return the currently active {@link Span}.
+     *
+     * <p>
+     * If there is an {@link Span non-null active span}, it becomes an implicit parent
```
yurishkuro: Is it necessary to have this mentioned here? It think this comment only belongs to the span builder methods. Scope Manager as an abstraction is not tied to the builder's behavior.

opentracing-api/src/main/java/io/opentracing/ScopeManager.java
```diff
@@ -26,26 +26,96 @@
...
+     * (as {@link References#CHILD_OF} reference) of any
+     * newly-created {@link Span} at {@link SpanBuilder#start()} or {@link Tracer.SpanBuilder#startActive(boolean)}
+     * time rather than at {@link Tracer#buildSpan(String)} time.
+     *
+     * <p>
+     * Because both {@link #active()} and {@link #activeSpan()} reference the current
+     * the active state, they both will be either null or non-null.
```
yurishkuro: s/the active/active/

opentracing-api/src/main/java/io/opentracing/Tracer.java
```diff
@@ -26,11 +26,19 @@
...
+     * @return the active {@link Span}. This is a shorthand for {@code Tracer.scopeManager().activeSpan()}.
      */
     Span activeSpan();
 
+    /**
+     * Make a {@link Span} instance active for the current context (usually a thread).
+     * This is a shorthand for {@code Tracer.scopeManager().activate(span)}.
+     *
+     * @return a {@link Scope} instance to control the end of the active period for the {@link Span}. It is a
+     * programming error to neglect to call {@link Scope#close()} on the returned instance.
```
yurishkuro: and may lead to memory leaks and the scope may remain in the thread-local stack.

opentracing-api/src/main/java/io/opentracing/Tracer.java
```diff
@@ -40,19 +48,20 @@
...
-     *   // of an implicit CHILD_OF Reference for "workScope.span()" when `startActive()` is invoked.
-     *   try (Scope workScope = tracer.buildSpan("DoWork").startActive()) {
-     *       workScope.span().setTag("...", "...");
-     *       // etc, etc
-     *   }
-     *
-     *   // It's also possible to create Spans manually, bypassing the ScopeManager activation.
-     *   Span http = tracer.buildSpan("HandleHTTPRequest")
+     *   // Note: if there is a `tracer.activeSpan()` instance, it will be used as the target
+     *   // of an implicit CHILD_OF Reference when `start()` is invoked.
```
yurishkuro: unless another span reference is explicitly provided to the builder

opentracing-testbed/src/test/java/io/opentracing/testbed/activate_deactivate/README.md
carlosalberto: Hey Tyler - the *activate deactivate* and *multiple callbacks* examples use the more exoteric `AutoFinishScopeManager`, which was actually ignoring the `finishSpanOnClose` parameter anyway, doing the `Span.finish()` call by itself eventually.

I think we could keep this examples along with this specific `ScopeManager` implementation for now, as it shows a very, very specific case (having to finish a `Span` being active in more than one thread/callback). Observe this implementation has its specific and well-known limitations and, IIRC, was kept for the `opentracing-java-v030` shim layer and also for potential usage along with `Scala` frameworks, given their callback nature.

Another options is to refactor the `AutoFinishScopeManager` code to have a similarity with the new API in all ways (for example, by adding a new method that does the conditional finish in the `Span`, maybe `Scope.Continuation.finishSpan()` or similar, deprecating the current API there). *If* we do this part, I'd suggest leaving it for another PR, as I think it's time to merge the current one (I can propose myself to do that change).

Let me know ;)

carlosalberto:
@yurishkuro Thanks for the edit notes, appreciated. Please take a look whenever you can ;)

opentracing-testbed/src/test/java/io/opentracing/testbed/activate_deactivate/README.md
tylerbenson: Ah, I didn't notice the distinction in context here.  Carry on.

carlosalberto:
Hey @felixbarny @adriancole Wondering if you are fine with this PR? If yes, I'd like to have it merged, and probably have a Release Candidate to test things out soon after.

@yurishkuro Docs edits look good to you? ;)

felixbarny:
LGTM

opentracing-api/src/main/java/io/opentracing/ScopeManager.java
```diff
@@ -26,26 +26,92 @@
...
-     * If there is an {@link Scope non-null scope}, its wrapped {@link Span} becomes an implicit parent
-     * (as {@link References#CHILD_OF} reference) of any
-     * newly-created {@link Span} at {@link Tracer.SpanBuilder#startActive(boolean)} or {@link SpanBuilder#start()}
-     * time rather than at {@link Tracer#buildSpan(String)} time.
+     * Because both {@link #active()} and {@link #activeSpan()} reference the current
+     * active state, they both will be either null or non-null.
      *
      * @return the {@link Scope active scope}, or null if none could be found.
      */
     Scope active();
```
tylerbenson: I thought we were talking about deprecating this method.

carlosalberto:
Thanks @felixbarny ! 

I will proceed to merge this PR tomorrow as if seems to have found a proper trade-off of features, unless there's any complain from anybody ;)

carlosalberto:
Merged, thanks everybody for their feedback! Remember that this is not a final change, and it could be changed if/as needed (we will likely be doing a Release Candidate to test this and other changes out) ;)

sjoerdtalsma:
@carlosalberto I didn't participate in this PR as my opinion is already voiced by others, but I forgot to thank you for creating this!
So thanks :smile:

PR #310, opened 2018-10-12
Allow injecting into maps of type Map<String,Object>

tylerbenson:
For the inject case, it doesn’t matter so much that the generic type is declared as String.


yurishkuro:
not sure I follow this change. It still refers to string, and string is not extensible: `public final class String`

carlosalberto:
@yurishkuro If I'm correct, it allows a `Map<String, Object>` to be passed, so `String` objects are appended/set as values to the underlying map.

yurishkuro:
we should have a test demonstrating that.

tylerbenson:
@yurishkuro ok I modified the map in the test to use Object instead of String.  Is that what you were asking for?

Ideally we could do something similar for `TextMapExtractAdapter`, but that's not quite as straightforward.

carlosalberto:
@tylerbenson Thanks for the update

@yurishkuro Ok to merge this PR?

PR #311, opened 2018-10-12
Add generic typed setTag/withTag

### LINKED ISSUE #271: Add setTag/withTag overloads that accept `io.opentracing.tag` types
To use the fluent syntax with Span/SpanBuilder and the tag types, one currently has to write `span.setTag(Tags.HTTP_STATUS.getKey(), 200)`. This has two issues:
* It loses the type restrictions, offered by the tag types. People could call e.g. `span.setTag(Tags.HTTP_STATUS.getKey(), "foo")`
* Having to add `.getKey()` is an unnecessary inconvenience.

#### Proposed solution
Add overloads on `Span` and `SpanBuilder` that accept Tag types so that people can write `span.setTag(Tags.HTTP_STATUS, 200)`.

We've been discussing this in https://github.com/opentracing/opentracing-csharp/pull/72 and compared a few options - I'll copy [my latest comment](https://github.com/opentracing/opentracing-csharp/pull/72#issuecomment-372632290) for your convenience:

>I created [a similar perf test](https://github.com/opentracing/opentracing-csharp/commit/da68ce6f7cdeb22a70c1ec48c688fd9502deedb9) with BenchmarkDotNet.
>
>I've been testing the following signatures with `MockTracer` (calling `SetObjectTag()`) and `NoopTracer` (doing a no-op):
>```cs
>ISpan SetTag_Int(string key, int value);
>
>ISpan SetTag_AbstractGeneric<TTagValue>(AbstractTag<TTagValue> tag, TTagValue value);
>
>ISpan SetTag_AbstractTyped(AbstractTag<int> tag, int value);
>
>ISpan SetTag_IntTag(IntTag tag, int value);
>```
>Result:
>``` ini
>
>BenchmarkDotNet=v0.10.13, OS=Windows 10 Redstone 3 [1709, Fall Creators Update] (10.0.16299.248)
>Intel Core i5-4300U CPU 1.90GHz (Haswell), 1 CPU, 4 logical cores and 2 physical cores
>Frequency=2435765 Hz, Resolution=410.5486 ns, Timer=TSC
>.NET Core SDK=2.1.4
>  [Host]     : .NET Core 2.0.5 (CoreCLR 4.6.26020.03, CoreFX 4.6.26018.01), 64bit RyuJIT
>  DefaultJob : .NET Core 2.0.5 (CoreCLR 4.6.26020.03, CoreFX 4.6.26018.01), 64bit RyuJIT
>
>
>```
>|                       Method |      Mean |     Error |    StdDev |
>|----------------------------- |----------:|----------:|----------:|
>|              MockSpan_IntTag | 77.927 ns | 1.5573 ns | 1.5992 ns |
>| MockSpan_AbstractTag_Generic | 90.484 ns | 1.5414 ns | 1.4418 ns |
>|   MockSpan_AbstractTag_Typed | 77.102 ns | 1.5754 ns | 1.6178 ns |
>|          MockSpan_IntTag_Key | 80.962 ns | 1.5437 ns | 1.4440 ns |
>|               MockSpan_Const | 76.666 ns | 1.5525 ns | 1.5943 ns |
>|      MockSpan_Const_Overload | 77.437 ns | 1.5908 ns | 1.4880 ns |
>|              NoopSpan_IntTag |  4.040 ns | 0.1024 ns | 0.0958 ns |
>| NoopSpan_AbstractTag_Generic | 13.419 ns | 0.2553 ns | 0.2263 ns |
>|   NoopSpan_AbstractTag_Typed |  4.625 ns | 0.1388 ns | 0.2573 ns |
>|          NoopSpan_IntTag_Key |  5.052 ns | 0.1431 ns | 0.1757 ns |
>|               NoopSpan_Const |  3.021 ns | 0.0989 ns | 0.0925 ns |
>|      NoopSpan_Const_Overload |  3.068 ns | 0.0985 ns | 0.1054 ns |
>
>Using one `SetTag<T>(AbstractTag<T> tag, T value)` method is considerably slower so I would rule that out. It's a little bit more future-proof but it doesn't properly support the existing `IntOrStringTag` and adding new types is a breaking change anyway, as we also have to add a regular `SetTag(string key, NewType value)`.
>
>It's surprising to me that using `AbstractTag<int>` is sometimes faster than using `IntTag` directly (I ran it a few times). I would still rule that out though as this method still doesn't properly support `IntOrStringTag`.
>
>I would therefore prefer using the actual Tag-types directly! This would ensure, all existing Tag types are supported and it also looks better in IntelliSense IMO. Adding a new Tag type is a breaking change but as I've said before, this is breaking anyway.
>
>To sum it up, I would prefer the following additions to our interfaces:
>
>```cs
>// ISpan
>ISpan SetTag(BooleanTag tag, bool value);
>ISpan SetTag(IntOrStringTag tag, string value);
>ISpan SetTag(IntTag tag, int value);
>ISpan SetTag(StringTag tag, string value);
>
>// ISpanBuilder
>ISpanBuilder WithTag(BooleanTag tag, bool value);
>ISpanBuilder WithTag(IntOrStringTag tag, string value);
>ISpanBuilder WithTag(IntTag tag, int value);
>ISpanBuilder WithTag(StringTag tag, string value);
>```
----------

Java might have different performance characteristics of course!

Adding overloads should be non-breaking for instrumentation code but it obviously is a breaking change for tracers. 

Note that we could add this in a non-breaking way in C# by making them "extension methods" (see [comment](https://github.com/opentracing/opentracing-csharp/pull/72#issuecomment-369864433)) but this gives tracers less flexibility and it would make the `NoopTracer` a little less efficient (because the extension method will still be executed and the no-op will happen later). Not sure if there's something similar in Java?

tylerbenson:
Reduces the need to call Tags.SOMETAG.getKey() and adds type safety for setTag if done from the span.

Fixes #271

### DIFF
--- a/opentracing-api/src/main/java/io/opentracing/Span.java
+++ b/opentracing-api/src/main/java/io/opentracing/Span.java
@@ -13,6 +13,7 @@
  */
 package io.opentracing;
 
+import io.opentracing.tag.AbstractTag;
 import java.util.Map;
 
 /**
@@ -43,6 +44,9 @@ public interface Span {
 
     /** Same as {@link #setTag(String, String)}, but for numeric values. */
     Span setTag(String key, Number value);
+
+    /** Same as {@link #setTag(String, String)}, but with typed value. */
+    <T> Span setTag(AbstractTag<T> key, T value);
 
     /**
      * Log key:value pairs to the Span with the current walltime timestamp.
--- a/opentracing-api/src/main/java/io/opentracing/Tracer.java
+++ b/opentracing-api/src/main/java/io/opentracing/Tracer.java
@@ -14,6 +14,7 @@
 package io.opentracing;
 
 import io.opentracing.propagation.Format;
+import io.opentracing.tag.AbstractTag;
 
 /**
  * Tracer is a simple, thin interface for Span creation and propagation across arbitrary transports.
@@ -161,6 +162,9 @@ public interface Tracer {
 
         /** Same as {@link Span#setTag(String, Number)}, but for the span being built. */
         SpanBuilder withTag(String key, Number value);
+
+        /** Same as {@link AbstractTag#set(Span, T)}, but for the span being built. */
+        <T> SpanBuilder withTag(AbstractTag<T> tag, T value);
 
         /** Specify a timestamp of when the Span was started, represented in microseconds since epoch. */
         SpanBuilder withStartTimestamp(long microseconds);
--- a/opentracing-mock/src/main/java/io/opentracing/mock/MockSpan.java
+++ b/opentracing-mock/src/main/java/io/opentracing/mock/MockSpan.java
@@ -14,6 +14,9 @@
 package io.opentracing.mock;
 
 import io.opentracing.References;
+import io.opentracing.Span;
+import io.opentracing.SpanContext;
+import io.opentracing.tag.AbstractTag;
 import java.util.ArrayList;
 import java.util.Collections;
 import java.util.HashMap;
@@ -21,9 +24,6 @@ import java.util.List;
 import java.util.Map;
 import java.util.Objects;
 import java.util.concurrent.atomic.AtomicLong;
-
-import io.opentracing.Span;
-import io.opentracing.SpanContext;
 
 /**
  * MockSpans are created via MockTracer.buildSpan(...), but they are also returned via calls to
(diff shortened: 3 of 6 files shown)


yurishkuro:
+1

felixbarny:
Nice one.
Why is this considered a breaking change?

carlosalberto:
+1

tylerbenson:
One thing to note, I didn't follow @cwe1ss's suggestion to split the method declaration out by type. I don't have anything against that.  If y'all prefer the non-generic option I can do that instead.  Thoughts?

tedsuo:
LGTM

carlosalberto:
@yurishkuro Approved - is it ok if I go ahead and merge?

PR #313, opened 2018-10-18
Allow clearing the current Scope.

carlosalberto:
Decided to take on being able to clear the current `Scope`, if any - this is done to prevent leaking `Scope` objects that were left around unintentionally.

If merged, this could become part of the next iteration of https://github.com/opentracing/specification/pull/122 ;)

@yurishkuro @adriancole @felixbarny @tylerbenson

### DIFF
--- a/opentracing-api/src/main/java/io/opentracing/ScopeManager.java
+++ b/opentracing-api/src/main/java/io/opentracing/ScopeManager.java
@@ -92,6 +92,16 @@ public interface ScopeManager {
     Span activeSpan();
 
     /**
+     * Clear the current active state, setting both {@link #active()} and
+     * {@link #activeSpan()} to null for the current context (usually a thread).
+     *
+     * <p>
+     * This method can be called to discard any previously leaked {@link Scope} objects
+     * that were unintentionally left active.
+     */
+    void clear();
+
+    /**
      * @deprecated use {@link #activate(Span)} instead.
      * Set the specified {@link Span} as the active instance for the current
      * context (usually a thread).
--- a/opentracing-util/src/main/java/io/opentracing/util/AutoFinishScopeManager.java
+++ b/opentracing-util/src/main/java/io/opentracing/util/AutoFinishScopeManager.java
@@ -41,4 +41,9 @@ public class AutoFinishScopeManager implements ScopeManager {
         AutoFinishScope scope = tlsScope.get();
         return scope == null ? null : scope.span();
     }
+
+    @Override
+    public void clear() {
+        tlsScope.set(null);
+    }
 }
--- a/opentracing-util/src/main/java/io/opentracing/util/ThreadLocalScopeManager.java
+++ b/opentracing-util/src/main/java/io/opentracing/util/ThreadLocalScopeManager.java
@@ -45,4 +45,9 @@ public class ThreadLocalScopeManager implements ScopeManager {
         Scope scope = tlsScope.get();
         return scope == null ? null : scope.span();
     }
+
+    @Override
+    public void clear() {
+        tlsScope.set(null);
+    }
 }
--- a/opentracing-noop/src/main/java/io/opentracing/noop/NoopScopeManager.java
+++ b/opentracing-noop/src/main/java/io/opentracing/noop/NoopScopeManager.java
@@ -49,6 +49,10 @@ class NoopScopeManagerImpl implements NoopScopeManager {
         return NoopSpan.INSTANCE;
     }
 
+    @Override
+    public void clear() {
+    }
+
     static class NoopScopeImpl implements NoopScopeManager.NoopScope {
         @Override
         public void close() {}
--- a/opentracing-util/src/test/java/io/opentracing/util/AutoFinishScopeManagerTest.java
+++ b/opentracing-util/src/test/java/io/opentracing/util/AutoFinishScopeManagerTest.java
@@ -57,4 +57,18 @@ public class AutoFinishScopeManagerTest {
         Scope missingSpan = source.active();
         assertNull(missingSpan);
     }
+
+    @Test
+    public void clear() throws Exception {
+        Span span = mock(Span.class);
+
+        Scope scope = source.activate(span);
+        assertNotNull(scope);
+        assertNotNull(source.active());
+        assertNotNull(source.activeSpan());
+
+        source.clear();
+        assertNull(source.active());
+        assertNull(source.activeSpan());
+    }
 }
--- a/opentracing-util/src/test/java/io/opentracing/util/ThreadLocalScopeManagerTest.java
+++ b/opentracing-util/src/test/java/io/opentracing/util/ThreadLocalScopeManagerTest.java
@@ -114,4 +114,18 @@ public class ThreadLocalScopeManagerTest {
         assertNull(source.active());
         assertNull(source.activeSpan());
     }
+
+    @Test
+    public void clear() throws Exception {
+        Span span = mock(Span.class);
+
+        Scope scope = source.activate(span);
+        assertNotNull(scope);
+        assertNotNull(source.active());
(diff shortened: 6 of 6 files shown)


opentracing-util/src/main/java/io/opentracing/util/AutoFinishScopeManager.java
```diff
@@ -41,4 +41,9 @@ public Span activeSpan() {
         AutoFinishScope scope = tlsScope.get();
         return scope == null ? null : scope.span();
     }
+
+    @Override
+    public void clear() {
+        tlsScope.set(null);
```
sjoerdtalsma: Shouldn't this be `tlsScope.remove()` ?

opentracing-util/src/main/java/io/opentracing/util/ThreadLocalScopeManager.java
```diff
@@ -45,4 +45,9 @@ public Span activeSpan() {
         Scope scope = tlsScope.get();
         return scope == null ? null : scope.span();
     }
+
+    @Override
+    public void clear() {
+        tlsScope.set(null);
```
sjoerdtalsma: Same as above, shouldn't this be `tlsScope.remove()` ?

opentracing-api/src/main/java/io/opentracing/ScopeManager.java
```diff
@@ -91,6 +91,16 @@
...
 
+    /**
+     * Clear the current active state, setting both {@link #active()} and
+     * {@link #activeSpan()} to null for the current context (usually a thread).
+     *
+     * <p>
+     * This method can be called to discard any previously leaked {@link Scope} objects
+     * that were unintentionally left active.
+     */
+    void clear();
```
tylerbenson: Is there any difference in expected behavior between this calling 
```java
activate(null)
```

opentracing-util/src/main/java/io/opentracing/util/AutoFinishScopeManager.java
felixbarny: I think both have the same semantics but `tlsScope.remove()` removes the entry in the thread local map whereas `set(null)` still keeps it, which leads to fewer allocations of `java.lang.ThreadLocal.ThreadLocalMap.Entry` objects.

opentracing-api/src/main/java/io/opentracing/ScopeManager.java
felixbarny: Some implementations may keep a stack of activated scopes. My understanding is that `clear()` would clear the whole stack and `activate` would push another element on the stack.

Not quite sure but `clear()` may have slightly different semantics depending on whether the ScopeManager maintains a stack of Scopes or only holds the currently active scope.

sjoerdtalsma: :+1+ My context manager -by default- maintains a stack.

opentracing-util/src/main/java/io/opentracing/util/AutoFinishScopeManager.java
sjoerdtalsma: @felixbarny sounds ok.
@carlosalberto  If this is the reasoning, could you please add an (inline) comment so future developers also know this?

opentracing-api/src/main/java/io/opentracing/ScopeManager.java
felixbarny: Sorry for the OT, but I'm thinking about doing the same. What are your reasons for that? Is your context manager on GitHub?

felixbarny: In our specific implementation, we have to get ahold of the bottom of the stack aka. entry span. Also, the scopes don't need to reference the previously active span. This simplifies scopes and reduces GC overhead a little, as the scope object does not have to have to store a reference to the active span in a field.

sjoerdtalsma: _Since you ask_: My context manager is 'pluggable' and has no real opinion on stacking or not. The default context uses a stack so you can easily change e.g. the current `Locale` just in a small piece of code and have everything else the way it was. Current supporte contexts include `MDC`, Spring security context, `Locale`, originating `ServletRequest`, _and_ OpenTracing `Span`. You can plug additional types via a ServiceLoader SPI.
For many situations, having a Closeable context that can be nested is the 'easiest' for users (as it returns things to the way they were). For OpenTracing `Spans` however, I just delegate to the configured `ScopeManager`.  
_More info_: https://github.com/talsma-ict/context-propagation  
For questions, feel free to ping me at gitter, always curious what others think.

sjoerdtalsma: Tradeoffs for stacking: 
+ : Support nesting
+ : User experience: 'return things they way they were' / 'works as expected' in many cases
- : Obviously a bigger chance of memory leaks if open-close semantics are not correctly used.
- : More memory consumption / garbage collecting

felixbarny: > More memory consumption / garbage collecting

Not sure if that is true. The stack won't be collected unless the thread is collected.

opentracing-util/src/main/java/io/opentracing/util/AutoFinishScopeManager.java
carlosalberto: @sjoerdtalsma Hey, added a small note - does it look fine to you?

sjoerdtalsma: Yes, as this would have stopped me from asking 'why not remove()' :wink:

sjoerdtalsma:
@felixbarny 

> Who is supposed that method in which circumstances?

I can imagine some servlet filter right before returning the thread to the pool. Or a custom job scheduler, -again- also before returning a worker thread to the pool. i.e. the one who controls threads, but may not control the correct activation / closing of current scopes of all called code.

sjoerdtalsma:
Long time ago in the times of application servers and class loaders, I even went as far to create a list of all `static ThreadLocal` uses in our entire codebase (including libraries) and `remove()` all of them after each web request.
It was a hack, but it made several memory leaks disappear and gave us time to perform a more thorough analysis of the root cause(s).

CodingFabian:
I do not fully understand the purpose. Is this a band aid that can be used whenever something unexpected happens? A util for libraries to clear up the mess a user might have created or something that has legitimate use by a user?
users could just decide to not commit a span or create a new one.

While this api is technically no problem, I wonder what the guidelines for users are. How is this api intended to be used by end users.

carlosalberto:
Hey @CodingFabian 

So this is more of a preventive use case, I'd say, and is not expected to be used by final users most of the time (same as `ScopeManager.active()`). 

Quoting @sjoerdtalsma:

> I can imagine some servlet filter right before returning the thread to the pool

And the issue around https://github.com/opentracing-contrib/java-jaxrs/issues/109 could use this one.

carlosalberto:
Hey @CodingFabian Did the reason given above sounds right to you? Let me know ;)

@yurishkuro @opentracing/opentracing-java-maintainers Opinion on this?

opentracing-api/src/main/java/io/opentracing/ScopeManager.java
adriancole: I don't think this is a good idea. It is a hack, but a similar hack can be achieved by allowing a closure that clears a scope. ex passing a null tracecontext to the thing that scopes the active span

adriancole: the benefit of passing null is you can still provide for "descoping that null" which is in easiest form returning a noop scope object when it was already null.

ohbus (approved):

Issue #305, opened 2018-09-14
Add get to TextMap


adriancole:
I complained about this a couple years ago, but re-complaining as it seems possible some things can change now.

TextMap is accidentally constrained to an awkward pattern where you can add keys by name, but not retrieve by name. This is awkward because getting by name involves creating an iterator then scanning it. It also encourages people to cast to vendor types [like this](https://github.com/apache/geronimo-opentracing/blob/2d3799fb2fe9350a72d4b607322d08bcef8d719d/src/main/java/org/apache/geronimo/microprofile/opentracing/impl/GeronimoTracer.java#L106) for routine functionality.

I got bored looking to find a single implementation of TextMap that doesn't support get by name natively or that doesn't instantiate an intermediate java Map due to the awkwardness! The net result is guaranteed less efficient operations for the most common propagation use cases.

Can we please fix this as it will remove a lot of spaghetti. The fix is to add `@Nullable String get(String key)`

yurishkuro:
in many cases the native representation of the headers is a multi-map. Did you mean `List<String> get(String key)`? Or will the adapter have to concatenate into one string?

felixbarny:
The current method `Iterator<Map.Entry<String, String>> iterator();` also suggests that multi-values are not expected (although technically, there could be multiple entries with the same key). I'm not sure if those would be needed for the tracing use-case. I'd say, the adapter should get the first header. If you think there is a use case to get the list of headers, I'd add two methods, so that no objects have to be allocated for the common case:
 *  `@Nullable String getFirst(String key)`
 * `List<String> get(String key)`

Otherwise, big +1. Seems like a no-brainer to me.

yurishkuro:
The W3C tracestate header is by definition a multi-value header. Http protocol does not guarantee how those are transmitted: could be as repeated headers, could be as a comma-separated single header.

felixbarny:
> The W3C tracestate header is by definition a multi-value header.

So adding the `getFirst` and `get`/`getAll` methods makes even more sense, it seems?

tylerbenson:
I realize that the header definition is multi-map, but I don't know of a good case for TextMap to support that format.  @yurishkuro do you have any examples where that is important?

yurishkuro:
As I mentioned above, W3C `tracestate` header is explicitly multi-value, so even if the sender puts all values into a single header the routing proxies are within their rights to rewrite that as one value per header. The currently on-hold `correlation-context` header will also be multi-value.

[...]

PR #316, opened 2018-10-23
Split Inject and Extract Builtin interfaces

### LINKED ISSUE #315: Separate TextMap into read/write interfaces.
It currently has both combined into a single interface, but implementations commonly only implement one of the methods.  Even the included adapters do this:
https://github.com/opentracing/opentracing-java/blob/master/opentracing-api/src/main/java/io/opentracing/propagation/TextMapInjectAdapter.java#L39
https://github.com/opentracing/opentracing-java/blob/master/opentracing-api/src/main/java/io/opentracing/propagation/TextMapExtractAdapter.java#L43

This seems like an indication of a design problem.

I propose we separate this interface into two different interfaces, then let the current TextMap interface extend from each of them.

At this same time, we could include changes proposed in #305.

Thoughts?

tylerbenson:
If we had separate formats for inject and extract this refactor would have been much cleaner.

This allows us to better avoid partially implementing the TextMap interface and throwing an exception for the other method.

Since Binary is a new addition, I applied a similar refactoring to that.

Closes #315.

### DIFF
--- a/opentracing-api/src/main/java/io/opentracing/propagation/TextMap.java
+++ b/opentracing-api/src/main/java/io/opentracing/propagation/TextMap.java
@@ -25,28 +25,5 @@ import java.util.Map;
  * @see io.opentracing.Tracer#inject(SpanContext, Format, Object)
  * @see io.opentracing.Tracer#extract(Format, Object)
  */
-public interface TextMap extends Iterable<Map.Entry<String, String>> {
-    /**
-     * Gets an iterator over arbitrary key:value pairs from the TextMapReader.
-     *
-     * @return entries in the TextMap backing store; note that for some Formats, the iterator may include entries that
-     * were never injected by a Tracer implementation (e.g., unrelated HTTP headers)
-     *
-     * @see io.opentracing.Tracer#extract(Format, Object)
-     * @see Format.Builtin#TEXT_MAP
-     * @see Format.Builtin#HTTP_HEADERS
-     */
-    Iterator<Map.Entry<String,String>> iterator();
-
-    /**
-     * Puts a key:value pair into the TextMapWriter's backing store.
-     *
-     * @param key a String, possibly with constraints dictated by the particular Format this TextMap is paired with
-     * @param value a String, possibly with constraints dictated by the particular Format this TextMap is paired with
-     *
-     * @see io.opentracing.Tracer#inject(io.opentracing.SpanContext, Format, Object)
-     * @see Format.Builtin#TEXT_MAP
-     * @see Format.Builtin#HTTP_HEADERS
-     */
-    void put(String key, String value);
+public interface TextMap extends TextMapInject, TextMapExtract {
 }
--- a/opentracing-api/src/main/java/io/opentracing/propagation/BinaryAdapters.java
+++ b/opentracing-api/src/main/java/io/opentracing/propagation/BinaryAdapters.java
@@ -27,7 +27,7 @@ public final class BinaryAdapters {
      *
      * @return The new {@link Binary} carrier used for extraction.
      */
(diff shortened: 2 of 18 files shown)


carlosalberto:
So I like this change definitely, but I'm wondering about breaking the API (which we could actually do for `BINARY` as we are breaking it anyway, and mostly because nobody uses it currently, AFAIK).

Wondering if there's something we could do to help with such API break.

opentracing-api/src/main/java/io/opentracing/propagation/TextMap.java
```diff
@@ -25,28 +25,5 @@
...
-     *
-     * @param key a String, possibly with constraints dictated by the particular Format this TextMap is paired with
-     * @param value a String, possibly with constraints dictated by the particular Format this TextMap is paired with
-     *
-     * @see io.opentracing.Tracer#inject(io.opentracing.SpanContext, Format, Object)
-     * @see Format.Builtin#TEXT_MAP
-     * @see Format.Builtin#HTTP_HEADERS
-     */
-    void put(String key, String value);
+public interface TextMap extends TextMapInject, TextMapExtract {
```
yurishkuro: are we deprecating this then?

tylerbenson: I think it is still useful if someone wants to create a single wrapper for both inject/extract purposes.  I don't think it's necessary to deprecate.

tylerbenson:
@carlosalberto what API breakage are you referring to that you'd like to avoid?  I thought the way I made this change still maintained backwards compatibility.

I would still like to consider #305 which would definitely be a breaking change, but that isn't included in this yet.

opentracing-api/src/main/java/io/opentracing/propagation/Binary.java
```diff
@@ -13,6 +13,7 @@
  */
 package io.opentracing.propagation;
 
+import io.opentracing.SpanContext;
```
carlosalberto: Do we really need it here? ;)

carlosalberto:
Hey @tylerbenson 

Sorry for the late feedback, was busy with other things. So I think it indeed does not actually break the API, but I'm a little bit curious (more than concerned) with users having to **change** their injection code to:

```java
tracer.inject(span.context(), Format.Builtin.TEXT_MAP_INJECT, carrier);
```

Or else have tracer authors check for both `TEXT_MAP` and `TEXT_MAP_INJECT` (and `TEXT_MAP` and `TEXT_MAP_EXTRACT` for the extraction scenario, etc). Probably this last one is the one making more sense, backwards-compatibility wise (interestingly enough, in `MockTracer` we wouldn't need to change the handling at all, as we don't check the `Format<C>` object, but only its `C` type ;) )

tylerbenson:
@carlosalberto I believe the way I wrote it allows the new built-in formats to be opt-in for API usage.  For tracer authors, I think it depends on implementation, though I think using the type is probably the best option.

opentracing-api/src/main/java/io/opentracing/propagation/Binary.java
carlosalberto: Fair enough

carlosalberto:
Hey all,

I'd like to get this PR moving - specially as it will **not** break things (as @tylerbenson refactored things nicely). 

One thing I still wonder is whether we should deprecate `TextMap` or keep it, as @tyler also mentioned:

> I think it is still useful if someone wants to create a single wrapper for both inject/extract purposes

Thoughts? @opentracing/opentracing-java-maintainers

tylerbenson:
@carlosalberto We have a few additional approvals. I think this is ready for merge now.

Issue #314, opened 2018-10-22
0.32.0 Release


carlosalberto:
This issue tracks the Release Candidate iterations towards the the new 0.32 API for OpenTracing-Java.

**RC Branch**: https://github.com/opentracing/opentracing-java/tree/v0.32.0

## Current Status
* **Trace Identifiers** have been added (`SpanContext.toTraceId()` and `SpanContext.toSpanId()`).
* Finishing `Span` upon `Scope.close()` has been deprecated.
* `SpanBuilder.startActive()` has been deprecated.
* `Scope.span()` has been deprecated.
* `ScopeManager.activeSpan()` has been added.
* Small refactoring of the `BINARY` format (now we let the users know about the required buffer size).
* Other API sugar for making the developers' life easier ;)

## Changes that might be included for the final release
* `ScopeManager.clear()`
* `GlobalTracer` registration improvements.
* `MockTracer` improvements.

## Release Process
* The new API is being developed on the **v0.32.0** branch.
* Changes to the release candidate will be released on maven as `0.32.0.RC1`, `0.32.0.RC2`, etc.

carlosalberto:
Hey everybody!

I created this Issue as a way to track the incoming RC iterations towards the 0.32.0 Release (just as we did with the previous 0.31 one). Feel free to comment and leave any feedback on this. The plan is to do a first Release Candidate on the next days ;)

cc @yurishkuro @opentracing/opentracing-java-maintainers @felixbarny

adriancole:
I understand how deprecations on api signatures work. I don't understand how one would accomplish a deprecated behavior like this:

"Finishing Span upon Scope.close() has been deprecated."

What does this mean? We still finish on close but somehow people know this won't happen later? If so, I disagree with this solution as it will just delay problem. Or are you deprecating methods that allow this? Maybe this can be clarified?

adriancole:
Initial is that you should very quickly release a v0.33.0 which removes the deprecated methods entirely.

I'll add some feedback based on thoughts in the diff https://github.com/opentracing/opentracing-java/compare/v0.31.0...v0.32.0

* `ScopeManager#active()` -- still there and undeprecated.. it should not be there
* `Span#setTag(AbstractTag<T> key, T value)` and `SpanBuilder#withTag(AbstractTag<T> key, T value)`-- very very awkward to make an api require an abstract class as a parameter.
[...]

PR #317, opened 2018-10-24
Do not use AbstractTag<T> for Span.setTag(), but the actual impls.

carlosalberto:
This is done to not use an abstract class for our public API,
so we take overloads of BooleanTag, IntTag and StringTag.

Addresses a small detail mentioned in #314 

@yurishkuro @tylerbenson

### DIFF
--- a/opentracing-api/src/main/java/io/opentracing/tag/Tag.java
+++ b/opentracing-api/src/main/java/io/opentracing/tag/Tag.java
@@ -15,17 +15,6 @@ package io.opentracing.tag;
 
 import io.opentracing.Span;
 
-public abstract class AbstractTag<T> implements Tag<T> {
-    protected final String key;
-
-    public AbstractTag(String tagKey) {
-        this.key = tagKey;
-    }
-
-    @Override
-    public String getKey() {
-        return key;
-    }
-
-    protected abstract void set(Span span, T tagValue);
+public interface Tag<T> {
+    String getKey();
 }
--- a/opentracing-api/src/main/java/io/opentracing/tag/Tag.java
+++ b/opentracing-api/src/main/java/io/opentracing/tag/Tag.java
@@ -17,4 +17,5 @@ import io.opentracing.Span;
 
 public interface Tag<T> {
     String getKey();
+    void set(Span span, T value);
 }
--- a/opentracing-mock/src/main/java/io/opentracing/mock/MockSpan.java
+++ b/opentracing-mock/src/main/java/io/opentracing/mock/MockSpan.java
@@ -16,7 +16,9 @@ package io.opentracing.mock;
 import io.opentracing.References;
 import io.opentracing.Span;
 import io.opentracing.SpanContext;
-import io.opentracing.tag.AbstractTag;
+import io.opentracing.tag.BooleanTag;
+import io.opentracing.tag.IntTag;
+import io.opentracing.tag.StringTag;
 import java.util.ArrayList;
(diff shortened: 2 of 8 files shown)


yurishkuro:
Why not use a typed interface? I'm not a fan of overloaded methods (even though it is consistent with the plain set methods).

carlosalberto:
@yurishkuro 

You refer to something like this?

```java
// IntTag and the others would be inheriting this class.
public class TypedTag<T> extends AbstractTag<T> {
    protected void set(Span span, T tagValue)
    {   
        span.setTag(super.key, tagValue.toString());
    }
}
```

carlosalberto:
Hey @yurishkuro sorry, re-read your message today and figured you simply wanted the same, but through an interface ;)

opentracing-api/src/main/java/io/opentracing/tag/Tag.java
```diff
@@ -0,0 +1,20 @@
...
+ * is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express
+ * or implied. See the License for the specific language governing permissions and limitations under
+ * the License.
+ */
+package io.opentracing.tag;
+
+import io.opentracing.Span;
+
+public interface Tag<T> {
+    String getKey();
```
yurishkuro: How about setValue(Span, T) instead?

opentracing-mock/src/main/java/io/opentracing/mock/MockSpan.java
```diff
@@ -137,8 +137,8 @@ public MockSpan setTag(String key, Number value) {
     }
 
     @Override
-    public <T> Span setTag(AbstractTag<T> key, T value) {
-        return setObjectTag(key.getKey(), value);
+    public <T> MockSpan setTag(Tag<T> tag, T value) {
+        return setObjectTag(tag.getKey(), value);
```
yurishkuro: If you add setValue to the interface you won't need to use object.

opentracing-api/src/main/java/io/opentracing/tag/Tag.java
carlosalberto: Had considered the same, but was afraid of making `AbstractTag.set()` public (right now is protected) ;)

Although now I think we could leave `set()` as abstract and have `setValue()`, as an alternative ;) Would that work for you?

yurishkuro: I would rather make set() public. I don't know why it's protected now, since all concrete classes make it public.

carlosalberto:
Hey @yurishkuro 

I added `Tag.set()` and updated `AbstractTag` as well - but left the call to `MockTracer.setObjectTag()` as this extra checks that no tags are added after a `Span` was finished, and would love to keep it around:

```java
    // An error will be reported if this Span is already finished.
    private synchronized MockSpan setObjectTag(String key, Object value) {
        finishedCheck("Adding tag {%s:%s} to already finished span", key, value);
```

Let me know ;)

yurishkuro:
I think it should do `tag.set(span, value)` instead. This way the correct typed set() method will be called by the tag itself, and _those_ methods would call setObjectTag

carlosalberto:
@yurishkuro Updated it to looks just like that ;) If you think that's fine, I will merge soon

yurishkuro:
🚢

carlosalberto:
👍

Issue #312, opened 2018-10-17
Unclear lifecycle of spans


felixbarny:
The way the ScopeManager is currently defined makes the lifecycle of Span objects unclear. This complicates span object recycling aka object pooling.

Recycling objects is a way to reduce the allocation overhead by maintaining a pool of pre-allocated Span instances. Instead of instantiating a new span every time a span is started, a span is taken out of the object pool. When the Span is finished and has been reported out of process, the span instance is reset and put back into the object pool. Using this technique, it is possible to add almost no allocation overhead to the monitored application which means less GC overhead caused by the tracer.

IMO, the lifecycle of a Span should be very straightforward. When finish() is called, no other interactions should be allowed so it is safe to recycle and reuse the whole span instance.

There are two APIs which break this assumption.

## Span.context
Quote from The spec of Span.finish:
> With the exception of the method to retrieve the Span's SpanContext, none of the below may be called after the Span is finished.

IMO, the SpanContext should be retrieved before the span is finished if it is intended to be used after the span is finished. The implementations are supposed to return an object which is not tied to the span’s lifecycle, for example making a copy of the internal span context.

## ScopeManager

Even in the current version of the spec, it  “forbids” calls after `finish()`:
> With the exception of the method to retrieve the Span's SpanContext, none of the below may be called after the Span is finished.

Scoping a span in a different thread than the one it is finished from, indirectly breaks this requirement. In the second thread, there is no way to know whether the `activeSpan()` has already finished, as it has no control over the lifecycle of that span. So calling `tracer.activeSpan().addTag("foo", "bar")` is never legal if the active span is finished in a different thread. The Span reference returned by `tracer.activeSpan()` might already be recycled and used to record an entirely different operation.

Example scenario:

Thread A | Thread B |  
-- | -- | --
Span span1 = startSpan() |   |  
Scope scopeSpan1 = scopeManager.activate(span1) |   |  
Executor.sumbit(Runnable) | Runnable.run |  
  | Scope scopeSpan1 = scopeManager.activate(span1) |  
scopeSpan1.close() |   |  
span1.finish() |   |  
  | tracer.activeSpan().addTag("foo", "bar") | Is this span finished or not? The caller has no way to know that. Worst case is that it now represents an entirely different operation.
  | Span span2 = startSpan() |  
  | span2.finish() |  
  | scopeSpan1.finish() |  

[...]

PR #319, opened 2018-10-30
Add ScopeManager#activate(SpanContext)

felixbarny:
Similar to `ScopeManager#activate(Span)` but used in cases where the thread in which the activation is performed does not have control over the life cycle of the span.

This puts the burden of being aware of the lifecycle to the one writing the instrumentation (for example [`TracedRunnable`](https://github.com/opentracing-contrib/java-concurrent/blob/master/src/main/java/io/opentracing/contrib/concurrent/TracedRunnable.java)) instead of the user of `Tracer.activeSpan()`. The `EarlySpanFinishTest` demonstrates this.

It also simplifies tracer implementations as with this change tracers don't have to guard against racy calls to `Span#finish()` and `ScopeManager#activate(Span)`. This is really painful to do if the tracer implementation relies on no methods being called on the span after `finish()` (for example when the tracer recycles the span object).

This addresses the problems laid out in https://github.com/opentracing/opentracing-java/issues/312

### DIFF
--- a/opentracing-api/src/main/java/io/opentracing/ScopeManager.java
+++ b/opentracing-api/src/main/java/io/opentracing/ScopeManager.java
@@ -40,7 +40,8 @@ public interface ScopeManager {
      * This {@link Scope} instance can be accessed at any time through {@link #active()},
      * in case it is not possible for the user to store it (when used through middleware
      * or start/finish event hooks, for example). The corresponding {@link Span} can be
-     * accessed through {@link #activeSpan()} likewise.
+     * accessed through {@link #activeSpan()} likewise. You can also get ahold of the
+     * active {@link Span}'s {@link SpanContext} via {@link #activeSpanContext()}.
      *
      * <p>
      * Usage:
@@ -58,11 +59,42 @@ public interface ScopeManager {
      *     }
      * </code></pre>
      *
+     * <p>
+     * Note: You may only activate spans when you own its life cycle.
+     * That means you must make sure that no other thread calls {@link Span#finish()}
+     * while the scope is still active.
+     * If you can't guarantee that, use {@link #activate(SpanContext)} instead.
+     *
      * @param span the {@link Span} that should become the {@link #activeSpan()}
      * @return a {@link Scope} instance to control the end of the active period for the {@link Span}. It is a
      * programming error to neglect to call {@link Scope#close()} on the returned instance.
      */
     Scope activate(Span span);
+
+    /**
+     * Similar to {@link #activate(Span)} but used in cases where the thread in which the
+     * activation is performed does not have control over the life cycle of the span.
+     *
+     * <p>
+     * One example of that is when performing an activation in the {@link Runnable#run()}
+     * method of a traced {@link Runnable} wrapper which is executed by an
+     * {@link java.util.concurrent.ExecutorService}.
+     *
+     * <p>
+     * This {@link Scope} instance can be accessed at any time through {@link #active()},
+     * in case it is not possible for the user to store it (when used through middleware
+     * or start/finish event hooks, for example). The corresponding {@link SpanContext} can be
+     * accessed through {@link #activeSpanContext()} likewise.
+     * In contrast to {@link #activate(Span)}, {@link #activeSpan()} will return {@code null}.
+     * This prevents users of the {@link #activeSpan()} API to accidentally interacting with
+     * already {@linkplain Span#finish() finished} spans.
+     *
+     * @param spanContext the {@link SpanContext} that should become the {@link #activeSpanContext()}
+     * @see #activate(Span)
+     * @return a {@link Scope} instance to control the end of the active period for the {@link Span}. It is a
+     * programming error to neglect to call {@link Scope#close()} on the returned instance.
+     */
+    Scope activate(SpanContext spanContext);
 
     /**
      * Return the currently active {@link Scope} which can be used to deactivate the currently active
@@ -84,12 +116,24 @@ public interface ScopeManager {
      * Return the currently active {@link Span}.
      *
      * <p>
-     * Because both {@link #active()} and {@link #activeSpan()} reference the current
-     * active state, they both will be either null or non-null.
+     * Note that {@link #activeSpan()} can return {@code null} while {@link #active()} is non-null
+     * in case of a {@linkplain #activate(SpanContext) span context activation}.
      *
      * @return the {@link Span active span}, or null if none could be found.
      */
     Span activeSpan();
+
+    /**
+     * Return the currently active {@link SpanContext}, which was activated by activating either
+     * a {@link #activate(Span) Span} or a {@link #activate(SpanContext) SpanContext}.
+     *
+     * <p>
+     * Because both {@link #active()} and {@link #activeSpanContext()} reference the current
+     * active state, they both will be either null or non-null.
+     *
+     * @return the {@link SpanContext active span context}, or null if none could be found.
+     */
+    SpanContext activeSpanContext();
 
     /**
      * @deprecated use {@link #activate(Span)} instead.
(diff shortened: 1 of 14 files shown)


yurishkuro:
@felixbarny I understand the reasons why you want this, but I'm really troubled by the exponential increase of mental complexity of using such an API. Every time you want to use active span now you have to check for null. It makes life much harder for everyone who's using the API correctly already, ie not accessing the span after finish.

felixbarny:
> It makes life much harder for everyone who's using the API correctly already, ie not accessing the span after finish.

Not really because `ScopeManager#active()` would only return `null` when you don't own the lifecycle of the span. Which means that you are not allowed to call any methods on the returned span instance anyway. So when using the API correctly, you don't have to introduce any new `null` checks.

felixbarny:
> this puts a tracking burden on this function which is only needed when one creates bugs.

What do you mean with tracking burden? You mean tracking when a span is finished and then return `null` thereafter? This is precisely the thing I want to avoid having to do with this change.

This API addition forces instrumentation authors to think about whether the activation should allow code in scope to have access to the active span or if it has only relevance for setting parent-child relationships. They should choose the latter one in case the scope of the activation may outlive the corresponding span. That guards users against accidentally interacting with an already finished span.

> An alternative could be that implementations can be free to return a NoOp scope

Not sure I follow. Do you mean a noop span? With this change, you would still have a scope but it might only hold a SpanContext and no Span.

carlosalberto:
I totally agree with @yurishkuro opinion on the complexity part, and also think that as @adriancole says, having a no-op active `Span` (and related `Scope`) could help. 

(I think we had already talked briefly in the past about this possibility, so it's probably a good time to resurrect it ;) )

felixbarny:
> I totally agree with @yurishkuro opinion on the complexity part

Maybe there is something I'm missing but as explained, there is no additional complexity in terms of `null` checks as when the using the API correctly. Because when you use the API correctly, you may only get the active span when you are allowed to, that is if there is no chance for the span to already be finished. In that case, there are no new null checks.

Whether `Tracer.activeSpan()` and `ScopeManager.active()` should return a noop span rather than `null` is another discussion.

opentracing-api/src/main/java/io/opentracing/ScopeManager.java
```diff
@@ -58,12 +59,43 @@
...
+     * In contrast to {@link #activate(Span)}, {@link #activeSpan()} will return {@code null}.
+     * This prevents users of the {@link #activeSpan()} API to accidentally interacting with
+     * already {@linkplain Span#finish() finished} spans.
+     *
+     * @param spanContext the {@link SpanContext} that should become the {@link #activeSpanContext()}
+     * @see #activate(Span)
+     * @return a {@link Scope} instance to control the end of the active period for the {@link Span}. It is a
+     * programming error to neglect to call {@link Scope#close()} on the returned instance.
+     */
+    Scope activate(SpanContext spanContext);
```
adriancole: @felixbarny specifically what I mean here is that it is strange to have something that is meant for try/finally and returns null. Wouldn't you expect a NPE here because someone looks at something made for try/finally, but returns null?

felixbarny: The Scope would always be non null. But the scope can hold either a Span or just a SpanContext.

adriancole: my mistake. I misread this and I don't think others would misread it

opentracing-api/src/main/java/io/opentracing/ScopeManager.java
```diff
@@ -58,12 +59,43 @@
...
+     * Note: You may only activate spans when you own its life cycle.
+     * That means you must make sure that no other thread calls {@link Span#finish()}
+     * while the scope is still active.
+     * If you can't guarantee that, use {@link #activate(SpanContext)} instead.
+     *
      * @param span the {@link Span} that should become the {@link #activeSpan()}
      * @return a {@link Scope} instance to control the end of the active period for the {@link Span}. It is a
      * programming error to neglect to call {@link Scope#close()} on the returned instance.
      */
     Scope activate(Span span);
```
adriancole: would be nice to deprecate this, but I expect this to not be a popular suggestion

opentracing-api/src/main/java/io/opentracing/ScopeManager.java
```diff
@@ -58,12 +59,43 @@
...
 
+    /**
+     * Similar to {@link #activate(Span)} but used in cases where the thread in which the
+     * activation is performed does not have control over the life cycle of the span.
+     *
+     * <p>
+     * One example of that is when performing an activation in the {@link Runnable#run()}
+     * method of a traced {@link Runnable} wrapper which is executed by an
+     * {@link java.util.concurrent.ExecutorService}.
+     *
```
adriancole: add example in code? So far doesn't explicitly mention try/finally. I think this will reinforce it

opentracing-api/src/main/java/io/opentracing/ScopeManager.java
```diff
@@ -84,13 +116,25 @@
...
      *
      * <p>
-     * Because both {@link #active()} and {@link #activeSpan()} reference the current
-     * active state, they both will be either null or non-null.
+     * Note that {@link #activeSpan()} can return {@code null} while {@link #active()} is non-null
+     * in case of a {@linkplain #activate(SpanContext) span context activation}.
      *
      * @return the {@link Span active span}, or null if none could be found.
      */
     Span activeSpan();
```
adriancole: similarly prefer this to be deprecated

opentracing-api/src/main/java/io/opentracing/Tracer.java
```diff
@@ -31,6 +31,11 @@
      */
     Span activeSpan();
 
+    /**
+     * @return the active {@link SpanContext}. This is a shorthand for {@code Tracer.scopeManager().activeSpanContext()}.
+     */
+    SpanContext activeSpanContext();
```
adriancole: personally don't think these shortcuts should be here. just use the scope manager, and change docs of the activateSpan stuff to mention it internally uses the same. I don't know if Tracer.toSpan(SpanContext) has been done yet, but it would clear all of this up, possibly addressing @wu-sheng's gripe about not knowing when a span is made present on another thread.

opentracing-api/src/main/java/io/opentracing/ScopeManager.java
felixbarny: I think we should have a separate discussion about this

opentracing-api/src/main/java/io/opentracing/ScopeManager.java
felixbarny: added an example and copied over the try-with-resources paragraph

opentracing-api/src/main/java/io/opentracing/Tracer.java
felixbarny: In the interest of scoping this pr (see what I did there?) I think it's better to move this discussion to a separate issue. I added this to be in line with the current conventions. I have not heard of `Tracer.toSpan(SpanContext)` yet. What are Wu's gripes you are referring to?

opentracing-api/src/main/java/io/opentracing/ScopeManager.java
carlosalberto: I'd be (initially) against it ;) And yeah, this can definitely happen on a different discussion.

opentracing-api/src/main/java/io/opentracing/Tracer.java
adriancole: this topic is indeed separately addressable, though should be done! I may have misunderstood your SELF link issue as the same end functionality as `Tracer.toSpan(SpanContext)`

PR #320, opened 2018-11-16
Initial implementation of a no-op active Span.

carlosalberto:
Hey all,

I decided to test out, as a prototype, a no-op active `Scope`/`Span` for the current api. This had been suggested (or at least mentioned) in the past. And it could probably help with features such as #319 

Advantages:
* Never have to check for null on `ScopeManager.active()` nor `ScopeManager.activeSpan()`.

Potential issues:
* Tracers need to be aware of this feature - see the code in `MockTracer` which will check against `NoopSpanContext`, and which case it will ignore it as potential parent when creating new `Span`s.
* Test writers might need to check against the same -, i.e. `assertEquals(NoopScope.INSTANCE, scopeManager.active())` instead of `assertNull(scopeManager.active())`.

Thoughts? 

@yurishkuro @felixbarny @tedsuo @sjoerdtalsma @tylerbenson @opentracing/opentracing-java-maintainers 

PS - Did not update the testbed module. Will do that later if we decide to go on and merge this change ;)

### DIFF
--- a/opentracing-mock/src/main/java/io/opentracing/mock/MockTracer.java
+++ b/opentracing-mock/src/main/java/io/opentracing/mock/MockTracer.java
@@ -30,7 +30,7 @@ import io.opentracing.ScopeManager;
 import io.opentracing.Span;
 import io.opentracing.SpanContext;
 import io.opentracing.Tracer;
-import io.opentracing.noop.NoopScopeManager;
+import io.opentracing.noop.NoopSpanContext;
 import io.opentracing.propagation.Binary;
 import io.opentracing.propagation.Format;
 import io.opentracing.propagation.TextMap;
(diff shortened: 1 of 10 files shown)


sjoerdtalsma:
Conceptually, I would have a little trouble understanding what a _no-op_ scope is. Is it a scope? Is it an _empty_ scope? If so, would that be a better name? What is _no-op_ scope precisely?

Ideally, I would prefer to receive an `Optional` result from the `active..` methods. To me, it is much clearer about what is happening.

sjoerdtalsma:
Accidentally closed this pull request, sorry!

carlosalberto:
Hey @sjoerdtalsma 

> Is it a scope? Is it an empty scope? If so, would that be a better name? What is no-op scope precisely?

I'd say it's both an empty scope and an empty span, that basically do nothing. We could probably change the name?

The `Optional` is another possibility to consider, but we don't support Java 8 yet ;)

carlosalberto:
After the lack of interest, I'm closing this PR. We can definitely re-open or create a new ticket if/when this feature is needed ;)

PR #321, opened 2018-11-28
add isRegistered boolean to GlobalTracer

MikeGoldsmith:
- add isRegistered boolean
- set isRegistered when Registering a tracer
- add unit test to verify registering a NoopTracer set the isRegistered bool
- move resetTracer logic to GlobalTracer as private function instead of test utility for easier reusability

Ported from C# fix where the bug was originally reported in
opentracing/opentracing-csharp#113


austinlparker:
@yurishkuro do you know where the ticket you're referencing is? I'm unfamiliar with it.

opentracing-util/src/test/java/io/opentracing/util/GlobalTracerTestUtil.java
```diff
@@ -1,67 +0,0 @@
...
- *     &lt;dependency>
- *         &lt;groupId>io.opentracing&lt;/groupId>
- *         &lt;artifactId>opentracing-util&lt;/artifactId>
- *         &lt;version><em>version</em>&lt;/version>
- *         <strong>&lt;type>test-jar&lt;/type></strong>
- *         &lt;scope>test&lt;/scope>
- *      &lt;/dependency>
- * </code></pre>
- */
-public class GlobalTracerTestUtil {
```
sjoerdtalsma: Removing this class breaks backward compatibility for folks using the `test-jar` dependency!

MikeGoldsmith: I can revert this change, it's not essential to the fix. I should be able to update the isRegistered boolean in the same way the tracer is reset to a NoopTracer.

However, should we be supporting the use of GlobalTracerTestUtil to reset the tracer outside of the GlobalTracer jar?

yurishkuro:
@MikeGoldsmith GitHib mobile wasn't showing that part - looks good.

opentracing-util/src/test/java/io/opentracing/util/GlobalTracerTestUtil.java
sjoerdtalsma: Instead of updating two variables, have you could considered a constant UNREGISTERED_TRACER and check with `==` if a tracer was registered?

opentracing-util/src/test/java/io/opentracing/util/GlobalTracerTestUtil.java
```diff
@@ -59,6 +59,11 @@ public static void setGlobalTracerUnconditionally(Tracer tracer) {
             globalTracerField.setAccessible(true);
             globalTracerField.set(null, tracer);
             globalTracerField.setAccessible(false);
+
+            Field isRegisteredField = GlobalTracer.class.getDeclaredField("isRegistered");
```
sjoerdtalsma: This does something that is not according to the method name.
Setting unconditionally would intuitively leave `isRegistered` to true.
_Resetting_ should clear it though.

MikeGoldsmith:
@sjoerdtalsma Sorry for the delay in addressing your comments.

I've updated `resetGlobalTracer` and `setGlobalTracerUnconditionally` to correctly manage the _isRegistered_ property.

opentracing-util/src/test/java/io/opentracing/util/GlobalTracerTestUtil.java
carlosalberto: I agree with @sjoerdtalsma - IHMO resetting should be the only one touching `isRegistered` (an alternative would be to receive a `isRegistered` parameter, but it's probably not something needed at the moment, and thus could be added later if needed).

carlosalberto:
Hey @MikeGoldsmith 

Left a comment but overall this looks good ;)

sjoerdtalsma:
I think discarding the boolean (as it's just state) and replacing that with an `==` comparison with a unique no-op constant tracer gets you what you want (ability to set the NoopTracer) and reduces the complexity of your solution.

I won't object this PR though.

PR #287, opened 2018-07-13
Add service tag

pavolloffay:
Related to https://github.com/opentracing/specification/issues/77 and https://github.com/opentracing/specification/pull/119

Signed-off-by: Pavol Loffay <ploffay@redhat.com>

PR #322, opened 2018-12-05
Revert PR 287 which added service tag

tylerbenson:
See https://github.com/opentracing/specification/issues/131.

Until this gets resolved, I don't think it should be included in an official release.

Reverts opentracing/opentracing-java#287.

Note, this is being merged into the `v0.32.0` feature branch to be part of the next RC.

### DIFF
--- a/opentracing-api/src/main/java/io/opentracing/tag/Tags.java
+++ b/opentracing-api/src/main/java/io/opentracing/tag/Tags.java
@@ -46,17 +46,6 @@ public final class Tags {
     public static final String SPAN_KIND_CONSUMER = "consumer";
 
     /**
-     * The service name for a span, which overrides any default "service name" property defined
-     * in a tracer's config. This tag is meant to only be used when a tracer is reporting spans
-     * on behalf of another service (for example, a service mesh reporting on behalf of the services
-     * it is proxying). This tag does not need to be used when reporting spans for the service the
-     * tracer is running in.
-     *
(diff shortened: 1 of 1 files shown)


adriancole:
+1

opentracing-api/src/main/java/io/opentracing/tag/Tags.java
```diff
@@ -45,17 +45,6 @@ private Tags() {
...
-    /**
-     * The service name for a span, which overrides any default "service name" property defined
-     * in a tracer's config. This tag is meant to only be used when a tracer is reporting spans
-     * on behalf of another service (for example, a service mesh reporting on behalf of the services
-     * it is proxying). This tag does not need to be used when reporting spans for the service the
-     * tracer is running in.
-     *
-     * @see #PEER_SERVICE
-     */
-    public static final StringTag SERVICE = new StringTag("service");
```
yurishkuro: Why can't this be deprecated instead of removed?

carlosalberto: +1 here. This would be breaking things.

tylerbenson: This tag was not present in `0.31.0`, so this would only be a breaking change for someone using a unofficial build.  That is the main reason I'm pushing to get it removed before `0.32.0` so it doesn't have to go through a deprecation period.

Issue #267, opened 2018-04-11
Deprecate or remove `ScopeManager.active()`


codefromthecrypt:
`ScopeManager.active` is documented as a feature used to get the current span. Perhaps it is a relic from the previous design. It tempts people to call close on it eventhough ScopeManager is pretty clear that scopes should be managed directly.

I would remove this api, possibly for `ScopeManager.activeSpan` if that's the intended use case to serve. Otherwise, please clarify the documentation that calls to close are undefined.

If folks are doing things like subclassing, you can deprecate the api, and they can still offer "advanced" things as they exist today as their subtype would still be able to offer that functionality regardless of it it exists on a supertype.

https://github.com/openzipkin-contrib/brave-opentracing/pull/74#issuecomment-380382654

carlosalberto:
Hey @adriancole thanks for the feedback.

I think that's a very valid point, and the important question here is whether there is a pattern that *requires* the `Scope` to be available around (or a pattern that cannot keep the `Scope` around it close it itself).

Will think some though on it and try to cook some examples (both a) the need for the current design, if any, and b) the problem of exposing it in `ScopeManager`)

@yurishkuro @pavolloffay @tylerbenson  may have some thoughts on it too ;)

cwe1ss:
I need `ScopeManager.active` in C# to access the current `Scope` in scenarios where the library/framework I'd like to instrument only has `OnStart`/`OnStop` hooks without any chance to store the scope somewhere. 

E.g. .NET has its own OpenTracing "span"-like construct called `Activity` and I'm translating them to "real" spans with this code: https://github.com/opentracing-contrib/csharp-netcore/blob/master/src/OpenTracing.Contrib.NetCore/Internal/GenericEventProcessor.cs#L43-L67

If I only had access to `Tracer.activeSpan` I could not close the scope.

tylerbenson:
I'm probably missing something, but I don't understand the problem with that method... Why wouldn't you want a way to get the current scope from the ScopeManager?  How would you get it otherwise except from where it was originally created?  In fact, we use it in a fair bit of [our instrumentation](https://github.com/DataDog/dd-trace-java/search?l=Java&q=scopeManager+active).

When writing instrumentation, it is often useful to know if there is a current active span.  We also have additional logic on our scope that wouldn't be as nice to have on the span.  I suggest leaving it as is.

pavolloffay:
[...]

Issue #218, opened 2017-11-14
How long will we maintain JDK 1.6 compatibility?


yurishkuro:

tedsuo:
I would want to see us sit at a stable v1 before considering a 1.8+ API. In that sense, we maintain 1.6 compatibility "forever."

jpkrohling:
I think we should really, really consider 1.8 as the baseline. As API providers, it would help us a lot to be able to have default implementations on interfaces. For instance, #282 would be a non-breaking change with a code like this on the `SpanContext` interface:

```java
    default <T extends SpanContext> T unwrap() {
        return (T) this;
    }
```

carlosalberto:
This issue came up during #326 and I would like to re-take this discussion on the reasons behind it (to decide the timeline to move to 1.7/1.8).

As a start, this old issue could throw light into the historical context: https://github.com/opentracing/opentracing-java/issues/75

To me, it sounds like the important bit here is to dig into any framework or service that would really need to keep support for 1.6 (at least for the near future).

PR #326, opened 2019-01-09
Deprecate ScopeManager.active()

carlosalberto:
Addresses #267 

1. `ScopeManager.active()` is deprecated.
2. Tests around it are kept, till we fully remove it.
3. Its usage is removed from `testbed` though.
4. Updated `AutoFinishScopeManager` (a rather exoteric manager) to also offer its capture/continue workflow without relying on a publicly visible `Scope`.
5. Added a test case showing the case when the even handler or request handler has no object to store the `Scope` object, being solved by using  a thread-local field.

Probably the important point is 5) - as it shows a concrete scenario that has historically prevented us from removing `active()` altogether. I'm hope this case is not over simplifying this requirement ;)

@opentracing/opentracing-java-maintainers please feel free to leave your opinion, as I'd like to have this included in 0.32 if possible ;)

### DIFF
--- a/opentracing-util/src/test/java/io/opentracing/util/AutoFinishScopeManagerTest.java
+++ b/opentracing-util/src/test/java/io/opentracing/util/AutoFinishScopeManagerTest.java
@@ -37,6 +37,12 @@ public class AutoFinishScopeManagerTest {
     }
 
     @Test
+    public void missingCapturedScope() throws Exception {
+        AutoFinishScope.Continuation missingContinuation = source.captureScope();
+        assertNull(missingContinuation);
+    }
+
+    @Test
     public void activateSpan() throws Exception {
         Span span = mock(Span.class);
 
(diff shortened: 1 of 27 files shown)


opentracing-util/src/test/java/io/opentracing/util/AutoFinishScopeManagerTest.java
```diff
@@ -57,4 +63,36 @@ public void activateSpan() throws Exception {
         Scope missingSpan = source.active();
         assertNull(missingSpan);
     }
+
+    @Test
+    public void captureScope() throws Exception {
+        Span span = mock(Span.class);
+
+        // We can't use 1.7 features like try-with-resources in this repo without meddling with pom details for tests.
```
tylerbenson: I suggest we consider dropping support for 1.6.

yurishkuro: +1, about time.

sjoerdtalsma: Agree, but hopefully in a separate PR?

austinlparker: I think a discussion is appropriate but I'm not sure we should hold 0.32 for it.

carlosalberto: I think we can discuss that as part of #218 and decide what to do from there - and I agree with @austinlparker, I think we shouldn't hold 0.32 for this change to happen ;)

carlosalberto:
Just added a commit to add the new example to the `testbed` README. 

Unless somebody is not happy with this change, I will be merging this PR tomorrow in order to (very soon) issue a second 0.32 release candidate.

carlosalberto:
Merged prior to preparing a second RC for 0.32 ;)

Issue #302, opened 2018-09-10
Add close() method to io.opentracing.Tracer


SevaSafris:
The Tracer interface does not define a close() method. Implementors of the Tracer interface, however, most often implement their own close() method to finalize traces, and to release resources. Since io.opentracing.Tracer.close() is not defined, application code that utilizes Tracer(s) must import the vendor-specific implementation in order for the close() method to be available. This effectively makes the application code coupled to the vendor, since the vendor-specific import is required.

By declaring a close() method in the io.opentracing.Tracer interface, vendor coupling can be avoided.

The close() method can be declared with a noop default, thus allowing for backward compatibility with all implementors of io.opentracing.Tracer that do not implement a close() method.

jpkrohling:
> The close() method can be declared with a noop default

IIRC, this library is still on Java 6, so, no default implementation for interfaces are available :/

SevaSafris:
I just checked the [io.opentracing:parent:pom](https://github.com/opentracing/opentracing-java/blob/master/pom.xml#L38), and it has compiler source and target set to 1.7.

jpkrohling:
There seems to be an override here: https://github.com/opentracing/opentracing-java/blob/master/opentracing-api/pom.xml#L32

SevaSafris:
That override had eluded me. The override is itself a bad practice. So how should this move forward?
1) Declare the close() method without a default: This would force some vendors to have to implement the close() method if their Tracer(s) don't already have it implemented.
2) Remove the override: The class bytecode versions across the submodules are not consistent, which has the potential to lead to frustrating situations for vendors and app developers (this being one of them). I cannot comment on the reason for the override, as there is no comment in the pom where the override is set. Is this a path worth following? It's out of the scope of this issue, so a new issue would have to be created.
3) Do nothing: Keep the jdk1.6 override, and do not declare a Tracer.close(). This would force application code to be coupled to vendor implementations of the Tracer interface.

SevaSafris:
Yes, this is an Option 4. However, this option is a hack that would further bury the primary issues (not having a close() on Tracer, and the jdk1.6 override). I beliebe Option 4 would be to incur a mindful technical debt.

yurishkuro:
[...]

Issue #250, opened 2018-01-12
Tracer should be Closeable


fzakaria:
I would like to guarantee that any Reporter's close at the end of my program 
(in the case they are batching).
Tracer should implement Closeable similar to Jaeger's implementation.

wu-sheng:
Why a tracer need to close? Tracer should have a context about tracing. Batch module tracing is not related to Tracer, I think.

fzakaria:
I wrote it in the description. The use case is to shutdown reporters that may start threads or have file handlers.

wu-sheng:
You are talking about the implementation. How does OT-java API library user know when to close and start a new one? And what happens when users close a tracer and still have Trace-Context unclosed?

Even more, in Java, many tracing (like Apache SkyWalking, and more) use ThreadLocal to manage Tracing Context. You can't clear the context just close a tracer.

In my perspective, If you are facing a batch process, and implementation supports file handler, you should start a Tracer for batch/file at the beginning. Or keep the switch mechanism in implementation level.

yurishkuro:
The question of `close()` method has been raised in other languages as well, but we ended up with the same decisions:
  * for some tracer implementations `close()` method doesn't make sense, e.g. a Noop tracer or a wrapper tracer
  * in most cases the main program is aware which concrete tracer it is instantiating (rather than an abstract `io.opentracing.Tracer`, so the `close()` method would be available on the concrete instance. However, this doesn't work if the app is using dynamic loading of the tracer that Java supports.
  * the application can always do a conditional cast to the Closeable interface, so it's not necessary to pull it into the Tracer interface

carlosalberto:
@yurishkuro nice summary :) Sounds like we should put that in our FAQ (once we create one, maybe in the `specification` repository ;) )?

carlosalberto:
Trying to retake this item @yurishkuro  - think we should create a FAQ file or section in the README for cases like this? 

(I personally think this is not something that could be included in the Specification, as it's more a Language thing ;) )

yurishkuro:
+1 to add to the readme/faq

carlosalberto:
Hey @wu-sheng 

So it looks the way to proceed here is to create a FAQ file - starting with an explanation on this item (and why we don't implement the `Closeable` interface). Sounds fine?

pavolloffay:
I have some contra arguments on @yurishkuro comments. 

[...]

PR #329, opened 2019-02-08
Make Tracer extends Closable.

carlosalberto:
Prior to 0.32, I decided to take a shot at this feature (as part of https://github.com/opentracing/specification/issues/130) - the change itself is actually minimal, and I'm guessing that this could be an easy-to-implement item for Tracer maintainers, but would definitely love some feedback.

Depending on how good or bad the changes are deemed, we might include this in the 0.32 Final Release ;)

PS - Was tempted to leave `MockTracer.close()` as no-op, but preferred to keep the invariant.

@yurishkuro @tylerbenson @austinlparker  @opentracing/opentracing-java-maintainers

### DIFF
--- a/opentracing-api/src/main/java/io/opentracing/Tracer.java
+++ b/opentracing-api/src/main/java/io/opentracing/Tracer.java
@@ -130,6 +130,7 @@ public interface Tracer extends Closeable {
      * <p>
      * For stateless tracers, this can be a no-op.
      */
+    @Override
     void close();
 
     interface SpanBuilder {
--- a/opentracing-api/src/main/java/io/opentracing/Tracer.java
+++ b/opentracing-api/src/main/java/io/opentracing/Tracer.java
@@ -120,12 +120,13 @@ public interface Tracer extends Closeable {
     <C> SpanContext extract(Format<C> format, C carrier);
 
     /**
-     * Closes the Tracer, and flushes the in-memory collection to the configured persistance store.
+     * Closes the Tracer, and tries to flush the in-memory collection to the configured persistance store.
      *
      * <p>
      * The close method should be considered idempotent; closing an already closed Tracer should not raise an error.
      * Spans that are created or finished after a Tracer has been closed may or may not be flushed.
-     * Calling the close method should be considered a synchronous operation.
+     * Calling the close method should be considered a synchronous operation. Observe this call may block for
+     * a relatively long period of time, depending on the internal shutdown.
      * <p>
      * For stateless tracers, this can be a no-op.
      */
--- a/opentracing-api/src/main/java/io/opentracing/Tracer.java
+++ b/opentracing-api/src/main/java/io/opentracing/Tracer.java
@@ -13,13 +13,15 @@
  */
 package io.opentracing;
 
+import java.io.Closeable;
+
 import io.opentracing.propagation.Format;
 import io.opentracing.tag.Tag;
 
 /**
  * Tracer is a simple, thin interface for Span creation and propagation across arbitrary transports.
  */
-public interface Tracer {
+public interface Tracer extends Closeable {
 
     /**
      * @return the current {@link ScopeManager}, which may be a noop but may not be null.
@@ -117,6 +119,17 @@ public interface Tracer {
      */
     <C> SpanContext extract(Format<C> format, C carrier);
 
+    /**
+     * Closes the Tracer, and flushes the in-memory collection to the configured persistance store.
+     *
+     * <p>
+     * The close method should be considered idempotent; closing an already closed Tracer should not raise an error.
+     * Spans that are created or finished after a Tracer has been closed may or may not be flushed.
+     * Calling the close method should be considered a synchronous operation.
+     * <p>
+     * For stateless tracers, this can be a no-op.
+     */
+    void close();
 
     interface SpanBuilder {
 
--- a/opentracing-util/src/main/java/io/opentracing/util/GlobalTracer.java
+++ b/opentracing-util/src/main/java/io/opentracing/util/GlobalTracer.java
@@ -22,6 +22,7 @@ import io.opentracing.noop.NoopTracer;
 import io.opentracing.noop.NoopTracerFactory;
 import io.opentracing.propagation.Format;
 
+import java.io.IOException;
 import java.util.concurrent.Callable;
 
 /**
@@ -212,6 +213,11 @@ public final class GlobalTracer implements Tracer {
     @Override
     public Scope activateSpan(Span span) {
         return tracer.activateSpan(span);
+    }
+
+    @Override
+    public void close() {
+        tracer.close();
     }
 
     @Override
--- a/opentracing-util/src/main/java/io/opentracing/util/GlobalTracer.java
+++ b/opentracing-util/src/main/java/io/opentracing/util/GlobalTracer.java
@@ -22,7 +22,6 @@ import io.opentracing.noop.NoopTracer;
 import io.opentracing.noop.NoopTracerFactory;
 import io.opentracing.propagation.Format;
 
-import java.io.IOException;
 import java.util.concurrent.Callable;
 
 /**
--- a/opentracing-mock/src/main/java/io/opentracing/mock/MockTracer.java
+++ b/opentracing-mock/src/main/java/io/opentracing/mock/MockTracer.java
@@ -24,7 +24,6 @@ import java.io.ObjectOutputStream;
 import java.io.IOException;
 import java.nio.ByteBuffer;
 import java.util.ArrayList;
-import java.util.Collections;
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
--- a/opentracing-mock/src/main/java/io/opentracing/mock/MockTracer.java
+++ b/opentracing-mock/src/main/java/io/opentracing/mock/MockTracer.java
@@ -24,6 +24,7 @@ import java.io.ObjectOutputStream;
 import java.io.IOException;
 import java.nio.ByteBuffer;
 import java.util.ArrayList;
+import java.util.Collections;
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
@@ -53,6 +54,7 @@ public class MockTracer implements Tracer {
     private final List<MockSpan> finishedSpans = new ArrayList<>();
     private final Propagator propagator;
     private final ScopeManager scopeManager;
+    private boolean isClosed;
 
     public MockTracer() {
         this(new ThreadLocalScopeManager(), Propagator.TEXT_MAP);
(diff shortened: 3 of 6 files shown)


jpkrohling:
I'm +1 to this change, but haven't we had this discussion before?

opentracing-util/src/main/java/io/opentracing/util/GlobalTracer.java
```diff
@@ -22,6 +22,7 @@
 import io.opentracing.noop.NoopTracerFactory;
 import io.opentracing.propagation.Format;
 
+import java.io.IOException;
```
sjoerdtalsma: Why the import of IOException?
I don't see it being thrown?

carlosalberto: Oh, good catch. I forgot to remove it during my cleanup.

carlosalberto:
Hey @sjoerdtalsma

So, from the mentioned RFC: 

> In lieu of having to perform potentially unsafe casts of a Tracer to access it's implementation-specific Close method, the interface would ensure that it was available regardless.

Also, there was some previous feedback (https://github.com/opentracing/opentracing-java/issues/250#issuecomment-420210277 and https://github.com/opentracing/opentracing-java/issues/302#issuecomment-419909901) 

Hope that helps ;)

sjoerdtalsma:
@carlosalberto I have read those arguments and there seems to be a genuine need for this.
However, I'm more curious than anything else about a use-case where one is in control of the Tracer's lifecycle (because we need close) but not knowing which implementation he/she created earlier in the lifecycle.

Don't let my curiosity get in the way of merging by the way 😉 ..

tylerbenson:
@sjoerdtalsma perhaps a good example is around the usage of https://github.com/opentracing-contrib/java-tracerresolver, where the user only adds it to their classpath, and does not reference their desired tracer class directly.

jpkrohling:
> However, I'm more curious than anything else about a use-case where one is in control of the Tracer's lifecycle (because we need close) but not knowing which implementation he/she created earlier in the lifecycle

I think I might have mentioned before, but one of the concrete cases where I needed this was when implementing the tracing subsystem for WildFly. Basically, the application server will create a tracer based on the tracer implementation from the application's classpath (possibly packaged inside the application's WAR) and start/close it whenever the deployment has started/finished.

yurishkuro:
My only concern with Close() API is that it requires synchronous, uninterruptible behavior. I don't know why, but Lightstep's tracer for Go implements cancellable Close():

https://github.com/lightstep/lightstep-tracer-go/blob/9a9bdeed74406bebe00f3875387e5ef1e2364c9e/tracer.go#L190

In Jaeger we never had requests to support that. Our Node.js implementation of Close() is asynchronous, but that's more of a side effect of Node.js style.

So I am interested in hearing from existing implementations:
  * do you need Close() to be async (i.e. return a future)?
  * do you need Close() to be cancellable?

If yes on either question, would be good to hear & document the use cases.

jpkrohling:
> * do you need Close() to be async (i.e. return a future)?
> * do you need Close() to be cancellable?

For Java, catching an `InterruptedException` is typically sufficient to handle the case where the close has to be canceled. 

IMO, having the Tracer extend Closeable has a higher value than providing a non-standard async close method, or a close method with a context (or similar).

austinlparker:
@yurishkuro I don't think close is cancellable in go - the context object is only used in the flush so we can ensure there's nothing in the span buffer before closing the tracer. Not really sure about the rationale for that implementation other than making it easier to close + flush in one instruction rather than two.

carlosalberto:
Hey @yurishkuro @jpkrohling 

> IMO, having the Tracer extend Closeable has a higher value than providing a non-standard async close method, or a close method with a context (or similar).

I strongly agree with this. But I'm still curious on this detail, and would love so hear other points of view.

sjoerdtalsma:
@tylerbenson @jpkrohling Thanks for the use cases. With that I'm totally on board with this PR.

_I should've guessed the tracer resolver example though.._ :wink:

austinlparker:
Apparently the rationale behind the Golang `Close` method being cancellable is to prevent process shutdown being blocked by the tracer, so that a shutdown circuit breaker implemented on the context can kill the `Close` method. The more you know!

carlosalberto:
Hey hey @yurishkuro wondering if this makes sense to you? If you think we need more feedback, we could poke people a little bit more ;)

cc @codingfabian @felixbarny

opentracing-api/src/main/java/io/opentracing/Tracer.java
```diff
@@ -117,6 +119,17 @@
      */
     <C> SpanContext extract(Format<C> format, C carrier);
 
+    /**
+     * Closes the Tracer, and flushes the in-memory collection to the configured persistance store.
+     *
+     * <p>
+     * The close method should be considered idempotent; closing an already closed Tracer should not raise an error.
+     * Spans that are created or finished after a Tracer has been closed may or may not be flushed.
```
yurishkuro: this text implies that while  spans may not be flushed, the tracer must still create them like business as usual, right?

opentracing-api/src/main/java/io/opentracing/Tracer.java
```diff
@@ -117,6 +119,17 @@
...
+     * Calling the close method should be considered a synchronous operation.
```
yurishkuro: What should happen if tracer uses sync span export (e.g. via HTTP), is able to retry failures, and the tracing backend is down? This method theoretically can block for a long time?

yurishkuro:
I am being a bit pedantic, but the fact that Lightstep had run into this situation indicates that the use case of `tracer.close()` blocking is not completely theoretical, that's why I want the semantics of that method to be defined more carefully.

opentracing-api/src/main/java/io/opentracing/Tracer.java
carlosalberto: Yes, I'd say so (as mentioned in the spec RFC) - an alternative is to NOT allow Span objects to be created (from where? from the builder? from `start`?)

opentracing-api/src/main/java/io/opentracing/Tracer.java
carlosalberto: Oh, that's a good point. Retries are fine, I think - but that brings us to the timeout point, and whether we should tell the user it can block for long or it should return relatively soon.

Opinions on this?

(I _think_ every Tracer should try to flush for some given time -implementation specific- and then simply give up).

carlosalberto:
> I am being a bit pedantic, but the fact that Lightstep had run into this situation indicates that the use case of `tracer.close()` blocking is not completely theoretical, that's why I want the semantics of that method to be defined more carefully.

I think this is fine. We can iterate a little bit more to have better semantics ;) (worst case: if it takes too long, we can include it in a further release).

I'd love to get feedback on this

cc @austinlparker

yurishkuro:
> every Tracer should try to flush for some given time -implementation specific- and then simply give up

That's reasonable, but if the API was cognizant of that use case it wouldn't have to be implementation specific. I assume the option here is to return CancellableFuture, which would make the tracer not io.Closable.

carlosalberto:
> That's reasonable, but if the API was cognizant of that use case it wouldn't have to be implementation specific.

Agreed.

> I assume the option here is to return CancellableFuture, which would make the tracer not io.Closable.

Also correct. But we could offer this also in the future, for 'advanced' scenarios ;)

What about saying something along "this is a synchronous call, will **try** to flush pending Spans and shutdown related resources. Be aware of the fact it **may** ultimately block for long". This is perhaps too general but would help give better expectations to users.

yurishkuro:
Is the weak guarantee, especially of freeing resources and returning, actually going to be sufficient for the people who need this functionality?

austinlparker:
The people that have asked for the feature, iirc, haven't mentioned anything about wanting this to be cancellable, but have mentioned try-with-resources, so I think the current evidence tips towards implementing closeable rather than cancellable.

jpkrohling:
+1, making the Tracer implement Closeable has real benefits, such as being able to use with closeable streams and try-with-resources. Tracers aren't the only components that might depend on IO resources upon `close`, and the typical behavior is to properly handle signals like thread interruption when it has to exit right away.

yurishkuro:
@jpkrohling I don't follow your point about interruption. Typically this is used when you have work done in thread A and another thread B is waiting for that work to complete (e.g. B waits on `future.get()`). If someone interrupts A, then B gets the interrupted exception. In case of close() method, there's only single thread (most likely `main`), so who would/could interrupt the close() operation?

jpkrohling:
I expect blocking operations to not run in the main thread, so, `Tracer#close` should always run in a worker thread. Otherwise, you are right, but that's an application-level concern that applies to all other `#close` operations they might have (HTTP connections, file system operations, ...)

yurishkuro:
> I expect blocking operations to not run in the main thread, so, `Tracer#close` should always run in a worker thread

Well, then to interrupt it you need another API on the tracer, and there's still nobody to actually do the interrupt because `main` thread is blocked.

Anyway, we're getting into the weeds. My main concern is that we're saying that tracer does not need to provide any guarantee how long it will take to return from `close()`, and to avoid blocking indefinitely we can only suggest that implementations _support some sort of internal configurable timeout_. Returning a future that can be cancelled or waited on with a user-controlled timeout avoids this problem completely in a very clean way. It does not make the tracer Closable.

jpkrohling:
I understand the appeal, but I'm not aware of any standard APIs in Java for that. I'm afraid we'd be building something that deviates from what pretty much everything out there uses.

If we really have a need for the feature you described, we could consider adding an overloaded method, `#close(Context)` (or `#close(int)`) with a deadline/timeout. Until then, it's more prudent to keep it simple and consistent with what people are used to.

austinlparker:
🎉

opentracing-util/src/main/java/io/opentracing/util/GlobalTracer.java
sjoerdtalsma: Will you remove the import before merging?

carlosalberto: Yes ;)

carlosalberto:
Hey all - I've removed a pair of unused imports, and added a pair of minor notes on the `close()` behavior.

The plan is to merge it late Monday (start preparing the 0.32 release), btw ;)

opentracing-api/src/main/java/io/opentracing/Tracer.java
```diff
@@ -117,6 +119,18 @@
...
+     *
+     * <p>
+     * The close method should be considered idempotent; closing an already closed Tracer should not raise an error.
+     * Spans that are created or finished after a Tracer has been closed may or may not be flushed.
+     * Calling the close method should be considered a synchronous operation. Observe this call may block for
+     * a relatively long period of time, depending on the internal shutdown.
+     * <p>
+     * For stateless tracers, this can be a no-op.
+     */
+    void close();
```
t-8ch: Should this `@Override`?

carlosalberto: Done ;)

carlosalberto:
Merged it, thank you all for the feedback!

PR #333, opened 2019-03-08
Do not make isRegistered() synchronized.

carlosalberto:
This one does not really need to be synchronized.
The field itself is volatile and that is enough.

Probably being too pedantic on this, but it feels this is the correct thing to do ;)

@yurishkuro @sjoerdtalsma

### DIFF
--- a/opentracing-util/src/main/java/io/opentracing/util/GlobalTracer.java
+++ b/opentracing-util/src/main/java/io/opentracing/util/GlobalTracer.java
@@ -95,7 +95,7 @@ public final class GlobalTracer implements Tracer {
      *
      * @return Whether a tracer has been registered
      */
-    public static synchronized boolean isRegistered() { return isRegistered; }
+    public static boolean isRegistered() { return isRegistered; }
 
     /**
      * Register a {@link Tracer} to back the behaviour of the {@link #get()}.


opentracing-util/src/main/java/io/opentracing/util/GlobalTracer.java
```diff
@@ -95,7 +95,7 @@ public static Tracer get() {
      *
      * @return Whether a tracer has been registered
      */
-    public static synchronized boolean isRegistered() { return isRegistered; }
+    public static boolean isRegistered() { return isRegistered; }
```
sjoerdtalsma: Agree, I approve of this, but would still love this condition to be deduced from the tracer itself instead of the artificial boolean.
I prefer a unique 'unregistered' instance (behaviourally identical to NoopTracer) that can be compared by reference:
`return tracer != UNREGISTERED`.

This makes the atomic registration also semantically 'just register' the tracer instead of also setting some boolean.

carlosalberto: In that case we would need to allow the `NoopTracerFactory` to have more than an instance (currently a singleton), so we can apply a check against the very initial 'unregistered' instance ;)

Perhaps as an improvement after 0.32 is released?

yurishkuro: what, if any, would be user-visible change in behavior if internal boolean is removed?

sjoerdtalsma: > Perhaps as an improvement after 0.32 is released?

Agree, I approve of this PR, my comment was intended for further improvement.

> what, if any, would be user-visible change in behavior if internal boolean is removed?

I would hope none. Just more consistent internal bookkeeping (single variable instead of two to maintain).

carlosalberto: So we will look into this fine improvement (along probably a change into the no-op tracer factory) right after 0.32 ;)

PR #335, opened 2019-03-13
Deprecate AutoFinishScopeManager

carlosalberto:
The last fix before releasing 0.32 ;)

@yurishkuro

### DIFF
--- a/opentracing-util/src/main/java/io/opentracing/util/AutoFinishScope.java
+++ b/opentracing-util/src/main/java/io/opentracing/util/AutoFinishScope.java
@@ -19,6 +19,13 @@ import io.opentracing.Span;
 
 import java.util.concurrent.atomic.AtomicInteger;
 
+/**
+ * {@link AutoFinishScope} is a {@link Scope} implementation that uses ref-counting
+ * to automatically finish the wrapped {@link Span}.
+ *
+ * @see AutoFinishScopeManager
+ */
+@Deprecated
 public class AutoFinishScope implements Scope {
     final AutoFinishScopeManager manager;
     final AtomicInteger refCount;
--- a/opentracing-util/src/main/java/io/opentracing/util/AutoFinishScope.java
+++ b/opentracing-util/src/main/java/io/opentracing/util/AutoFinishScope.java
@@ -20,6 +20,10 @@ import io.opentracing.Span;
 import java.util.concurrent.atomic.AtomicInteger;
 
 /**
+ * @deprecated use {@link ThreadLocalScope} instead.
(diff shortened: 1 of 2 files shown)


opentracing-util/src/main/java/io/opentracing/util/AutoFinishScopeManager.java
```diff
@@ -18,6 +18,13 @@
 
 import java.util.concurrent.atomic.AtomicInteger;
 
+/**
+ * @deprecated use {@link ThreadLocalScopeManager} instead.
```
sjoerdtalsma: Does this explain you you can still 'automatically' finish spans when scope is closed?
Or do we not want that anymore?

i.e. Please describe the _reason_ for deprecating this functionality

yurishkuro:
The operating mode this classes proposes is in contrast with the main philosophy of 0.32 where we deprecated autoFinish flag in  activate (), because it too easily leads to programming mistakes of using try-with-resource on the scope and not having the right span to handle errors.

I think we should stick with one recommended instrumentation pattern, shown in 0.32 readme, instead of having the instrumentation be dependent on which scope manager is in use.

carlosalberto:
Also, the historical context is that we wanted to preserve this for people migrating from 0.30. 

That being said, I will add this and @yurishkuro 's note and then merge.

opentracing-util/src/main/java/io/opentracing/util/AutoFinishScopeManager.java
carlosalberto: Added a (very) small note. Let me know if you think we need further clarification/details ;)

opentracing-util/src/main/java/io/opentracing/util/AutoFinishScope.java
```diff
@@ -19,6 +19,17 @@
 
 import java.util.concurrent.atomic.AtomicInteger;
 
+/**
+ * @deprecated use {@link ThreadLocalScope} instead.
+ * The operation mode of this class contrasts with the 0.32
+ * deprecation of auto finishing {@link Span}s upon {@link Scope#close()}.
```
yurishkuro: Could also add a link to this issue for more background.

sjoerdtalsma: Great idea!

carlosalberto: Done ;)

carlosalberto:
Hey hey guys - if you are ok with this, I will merge later today, and issue the 0.32 release right after ;)

Let me know.

Issue #334, opened 2019-03-12
ThreadLocalScopeManager should be extensible


mdvorak:
Adding custom behavior for activating and deactivating scope is real pain - one have to copy `ThreadLocalScopeManager` and `ThreadLocalScope` into project and modify them.
It is because while `activate` can be overriden by `ScopeManager`, the actual deactivation happens in `ThreadLocalScope.close()` and is conditional. Even wrapping `ThreadLocalScope` and hooking to `close()` won't work. Thus, copy-paste it is.

Main use-case here is setting [`MDC`](https://logback.qos.ch/manual/mdc.html) context. Even with 0.32 and exposed spanId and traceId, there is no straightforward way to set those to MDC (unless I'm oblivious to something). While setting always MDC is a little opinionated and can be part of more automated libraries like for Spring (as stated in #202), ot-java should provide extension point for both apps and libraries to do that with ease.

Now, I'll prepare PR, just pick a solution :)

**1)** Replace calls to `ThreadLocalScopeManager.tlsScope.set()` in `ThreadLocalScope` with protected method in `ThreadLocalScopeManager`:
```
protected void setThreadScope(ThreadLocalScope scope) {
  this.tlsScope.set(scope);
}
```
This has least impact on current code, but exposes class internals. Question is, whether it is actually an issue.

**2)** Similar to 1), but add listener to `ThreadLocalScopeManager` instead of protected setter, to allow composition over inheritance:
```
public interface Listener {
    void onActiveSpanChanged(Span span);
}
final Listener listener;
```
Whenever `ThreadLocalScope` calls `scopeManager.tlsScope.set()`, it also calls `scopeManager.listener.onActiveSpanChanged`.

**3)** Create some more sophisticated and complex observer system. Problem with above solutions is, that `AutoFinishScopeManager` could have exactly same logic. Is it worth adding complex structures to avoid little code duplication? 
Also question is, what are actually plans with `AutoFinishScopeManager`, since it couples app code with specific configuration, and the implementation itself is error-prone IMO (leaking spans, not closing spans). 

I prefer solution 2), as it allows very simple use in Java8+. Example for Jaeger and Log4j2:
```
new JaegerTracer.Builder("sample")
            .withScopeManager(new ThreadLocalScopeManager(this::setThreadContext))
            .build();
// ...
private void setThreadContext(Span span) {
    if (span instanceof JaegerSpan) {
        JaegerSpan jaegerSpan = (JaegerSpan) span;
[...]

PR #336, opened 2019-03-15
#334 - Added ScopeListener interface to the ThreadLocalScopeManager

mdvorak:
* Added new interface `ScopeListener`
* Modified `ThreadLocalScopeManager`
* Added `NoopScopeListener` to be used as default
* Modified tests

Is there anything else missing?

Also, I've created signature of onActivate, thaking directly span:
```
void onActivate(Span span);
```
Because consumers should not actually need the scope, only its span - but it is a little weird, having Span in the interface called ScopeListener. What do you think? I have already prepared commit to change it to `void onActivate(Scope scope);`

### DIFF
--- a/opentracing-util/src/main/java/io/opentracing/util/NoopScopeListener.java
+++ b/opentracing-util/src/main/java/io/opentracing/util/NoopScopeListener.java
@@ -1,0 +1,31 @@
+/*
+ * Copyright 2016-2019 The OpenTracing Authors
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License"); you may not use this file except
+ * in compliance with the License. You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software distributed under the License
+ * is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express
+ * or implied. See the License for the specific language governing permissions and limitations under
+ * the License.
+ */
+package io.opentracing.util;
+
+import io.opentracing.Span;
+
+public interface NoopScopeListener extends ScopeListener {
+    NoopScopeListener INSTANCE = new NoopScopeListenerImpl();
+}
+
+class NoopScopeListenerImpl implements NoopScopeListener {
+
+    @Override
+    public void onActivate(Span span) {
+    }
+
+    @Override
+    public void onClose() {
+    }
+}
--- a/opentracing-util/src/main/java/io/opentracing/util/NoopScopeListener.java
+++ b/opentracing-util/src/main/java/io/opentracing/util/NoopScopeListener.java
@@ -19,6 +19,9 @@ public interface NoopScopeListener extends ScopeListener {
     NoopScopeListener INSTANCE = new NoopScopeListenerImpl();
 }
 
+/**
+ * A noop (i.e., cheap-as-possible) implementation of a ScopeListener.
+ */
 class NoopScopeListenerImpl implements NoopScopeListener {
 
     @Override
--- a/opentracing-util/src/main/java/io/opentracing/util/ScopeListener.java
+++ b/opentracing-util/src/main/java/io/opentracing/util/ScopeListener.java
@@ -13,11 +13,32 @@
  */
 package io.opentracing.util;
 
+import io.opentracing.Scope;
+import io.opentracing.ScopeManager;
 import io.opentracing.Span;
 
+/**
+ * Listener that can react on changes of currently active {@link Span}.
+ * <p>
+ * The {@link #onActivate(Span)} method will be called, whenever scope changes - that can be both
+ * as result of a {@link ScopeManager#activate(Span, boolean)} call or when {@link Scope#close()}
+ * is closed on a nested scope.
+ * <p>
+ * {@link #onClose()} is called when closing outermost scope - meaning no scope is currently active.
+ *
+ * @see ThreadLocalScopeManager
+ */
 public interface ScopeListener {
 
+    /**
+     * Called whenever a scope was activated (changed).
+     *
+     * @param span Activated span. Never null.
+     */
     void onActivate(Span span);
 
+    /**
+     * Called when outermost scope was deactivated.
+     */
     void onClose();
 }
--- a/opentracing-util/src/main/java/io/opentracing/util/ScopeListener.java
+++ b/opentracing-util/src/main/java/io/opentracing/util/ScopeListener.java
@@ -1,0 +1,23 @@
+/*
+ * Copyright 2016-2019 The OpenTracing Authors
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License"); you may not use this file except
+ * in compliance with the License. You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software distributed under the License
+ * is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express
+ * or implied. See the License for the specific language governing permissions and limitations under
+ * the License.
+ */
+package io.opentracing.util;
+
+import io.opentracing.Span;
+
+public interface ScopeListener {
+
+    void onActivate(Span span);
+
+    void onClose();
+}
--- a/opentracing-util/src/main/java/io/opentracing/util/ThreadLocalScope.java
+++ b/opentracing-util/src/main/java/io/opentracing/util/ThreadLocalScope.java
@@ -35,6 +35,7 @@ public class ThreadLocalScope implements Scope {
         this.finishOnClose = finishOnClose;
         this.toRestore = scopeManager.tlsScope.get();
         scopeManager.tlsScope.set(this);
+        scopeManager.listener.onActivate(wrapped);
     }
 
     @Override
@@ -48,7 +49,13 @@ public class ThreadLocalScope implements Scope {
             wrapped.finish();
         }
 
-        scopeManager.tlsScope.set(toRestore);
+        if (toRestore != null) {
+            scopeManager.tlsScope.set(toRestore);
+            scopeManager.listener.onActivate(toRestore.wrapped);
+        } else {
+            scopeManager.tlsScope.remove();
+            scopeManager.listener.onClose();
+        }
     }
 
     @Override
--- a/opentracing-util/src/main/java/io/opentracing/util/ThreadLocalScopeManager.java
+++ b/opentracing-util/src/main/java/io/opentracing/util/ThreadLocalScopeManager.java
@@ -23,7 +23,17 @@ import io.opentracing.Span;
  * @see ThreadLocalScope
  */
 public class ThreadLocalScopeManager implements ScopeManager {
+
     final ThreadLocal<ThreadLocalScope> tlsScope = new ThreadLocal<ThreadLocalScope>();
+    final ScopeListener listener;
+
+    public ThreadLocalScopeManager() {
+        this(null);
+    }
+
+    public ThreadLocalScopeManager(ScopeListener listener) {
+        this.listener = listener != null ? listener : NoopScopeListener.INSTANCE;
+    }
 
     @Override
     public Scope activate(Span span, boolean finishOnClose) {
--- a/opentracing-util/src/main/java/io/opentracing/util/ThreadLocalScopeManager.java
+++ b/opentracing-util/src/main/java/io/opentracing/util/ThreadLocalScopeManager.java
@@ -19,6 +19,10 @@ import io.opentracing.Span;
 
 /**
  * A simple {@link ScopeManager} implementation built on top of Java's thread-local storage primitive.
+ * <p>
+ * Optionally supports {@link ScopeListener}, to perform additional actions when scope is changed for given thread.
+ * Listener methods are always called synchronously on the same thread, right after activation (meaning {@link #active()}
+ * already returns new a scope).
  *
  * @see ThreadLocalScope
  */
@@ -27,10 +31,18 @@ public class ThreadLocalScopeManager implements ScopeManager {
     final ThreadLocal<ThreadLocalScope> tlsScope = new ThreadLocal<ThreadLocalScope>();
     final ScopeListener listener;
 
+    /**
+     * Default constructor for {@link ThreadLocalScopeManager}, without any listener.
+     */
     public ThreadLocalScopeManager() {
         this(null);
     }
 
+    /**
+     * Constructs {@link ThreadLocalScopeManager} with custom {@link ScopeListener}.
+     *
+     * @param listener Listener to register. When null, noop listener will be used.
+     */
     public ThreadLocalScopeManager(ScopeListener listener) {
         this.listener = listener != null ? listener : NoopScopeListener.INSTANCE;
     }
--- a/opentracing-util/src/test/java/io/opentracing/util/ThreadLocalScopeTest.java
+++ b/opentracing-util/src/test/java/io/opentracing/util/ThreadLocalScopeTest.java
@@ -27,10 +27,12 @@ import org.junit.Test;
 
 public class ThreadLocalScopeTest {
     private ThreadLocalScopeManager scopeManager;
+    private ScopeListener scopeListener;
 
     @Before
     public void before() throws Exception {
-        scopeManager = new ThreadLocalScopeManager();
+        scopeListener = mock(ScopeListener.class);
+        scopeManager = new ThreadLocalScopeManager(scopeListener);
     }
 
     @Test
@@ -63,6 +65,11 @@ public class ThreadLocalScopeTest {
         verify(backgroundSpan, times(1)).finish();
         verify(foregroundSpan, times(1)).finish();
 
+        // Verify listener calls
+        verify(scopeListener, times(2)).onActivate(backgroundSpan);
+        verify(scopeListener, times(1)).onActivate(foregroundSpan);
+        verify(scopeListener, times(1)).onClose();
+
         // And now nothing is active.
         Scope missingSpan = scopeManager.active();
         assertNull(missingSpan);
@@ -71,11 +78,16 @@ public class ThreadLocalScopeTest {
     @Test
     public void testDeactivateWhenDifferentSpanIsActive() {
         Span span = mock(Span.class);
+        Span nestedSpan = mock(Span.class);
 
         Scope active = scopeManager.activate(span, false);
-        scopeManager.activate(mock(Span.class), false);
+        scopeManager.activate(nestedSpan, false);
         active.close();
 
         verify(span, times(0)).finish();
+
+        verify(scopeListener, times(1)).onActivate(span);
+        verify(scopeListener, times(1)).onActivate(nestedSpan);
+        verify(scopeListener, times(0)).onClose();
     }
 }


opentracing-util/src/main/java/io/opentracing/util/ScopeListener.java
```diff
@@ -0,0 +1,44 @@
...
+     * Called whenever a scope was activated (changed).
+     *
+     * @param span Activated span. Never null.
+     */
+    void onActivate(Span span);
+
+    /**
+     * Called when outermost scope was deactivated.
+     */
+    void onClose();
```
yurishkuro: Why not pass the span here too?

opentracing-util/src/main/java/io/opentracing/util/ThreadLocalScope.java
```diff
@@ -48,7 +49,13 @@ public void close() {
             wrapped.finish();
```
yurishkuro: I think this should call onClose first

opentracing-util/src/main/java/io/opentracing/util/ThreadLocalScope.java
```diff
@@ -48,7 +49,13 @@ public void close() {
...
         }
 
-        scopeManager.tlsScope.set(toRestore);
+        if (toRestore != null) {
+            scopeManager.tlsScope.set(toRestore);
+            scopeManager.listener.onActivate(toRestore.wrapped);
+        } else {
+            scopeManager.tlsScope.remove();
+            scopeManager.listener.onClose();
```
yurishkuro: Should be done earlier imo

opentracing-util/src/main/java/io/opentracing/util/ScopeListener.java
mdvorak: Then it would make sense to call `onActivate` with previous span as well (optionally null). I'd think it is better to keep the interface as simple as possible. But its no problem to implement it of course :)

opentracing-util/src/main/java/io/opentracing/util/ThreadLocalScope.java
mdvorak: Are you sure? Listener code will then execute in the context of the previous span, not the one passed in as argument. Same applies to onClose.

opentracing-util/src/main/java/io/opentracing/util/ScopeListener.java
yurishkuro: I am not sure why previous span is relevant - if we're on an executor thread that span may be completely unrelated to the one being activated.

opentracing-util/src/main/java/io/opentracing/util/ThreadLocalScope.java
yurishkuro: again, the "previous span" is not really relevant. This function closes the current scope, so it should invoke `listener.close(currentScope.span)`. Then it can do the housekeeping of the stack and re-activation of the previous span.

mdvorak: Ok, maybe we misunderstood each other. Were you talking about onClose or onActivate?
`onClose` can be called before `scopeManager.tlsScope.remove();` and it makes sense. I'll update that.

Calling `onActivate` before `scopeManager.tlsScope.set(toRestore);` does not make sense - onActivate would run still with active previous span.

opentracing-util/src/main/java/io/opentracing/util/ScopeListener.java
mdvorak: Yes, but previous span is the one being deactivated - same situation as onClose. So, if you don't need deactivated span in onActivate, you should not need it in onClose either.

opentracing-util/src/main/java/io/opentracing/util/ThreadLocalScope.java
mdvorak: On the second thought, what about renaming methods to `onActivated` and `onClosed`, and really call them both after change has happened? After all, this is listener to change, not interceptor. I would naturally expect, that state transition has already happened, when I'm receiving the call.

yurishkuro: sgtm

mdvorak: Renamed

carlosalberto:
> Because consumers should not actually need the scope, only its span - but it is a little weird, having Span in the interface called ScopeListener

I think this is fine, aligning with the recent changes on 0.32 around exposing active `Span`, not the `Scope` object itself.

carlosalberto:
I'm wondering if supporting multiple listeners could be useful ;)

yurishkuro:
Multiple listeners are already possible via composition.

opentracing-util/src/main/java/io/opentracing/util/ScopeListener.java
```diff
@@ -0,0 +1,44 @@
...
+     * Called whenever a scope was activated (changed).
+     *
+     * @param span Activated span. Never null.
+     */
+    void onActivated(Span span);
+
+    /**
+     * Called when outermost scope was deactivated.
+     */
+    void onClosed();
```
yurishkuro: I still think it needs to accept the span represented by the scope being closed.

opentracing-util/src/main/java/io/opentracing/util/ThreadLocalScope.java
```diff
@@ -48,7 +49,13 @@ public void close() {
             wrapped.finish();
         }
 
-        scopeManager.tlsScope.set(toRestore);
+        if (toRestore != null) {
+            scopeManager.tlsScope.set(toRestore);
+            scopeManager.listener.onActivated(toRestore.wrapped);
+        } else {
+            scopeManager.tlsScope.remove();
+            scopeManager.listener.onClosed();
```
yurishkuro: the previous comments got lost - the onClosed must be called regardless of the previous span presence.

carlosalberto:
@yurishkuro sure, but was wondering if we should support this out-of-the-box ;) (not an important thing, just curious about the possibility)

yurishkuro:
We can always add it later, it's a utility class that would dispatch to 2+ underlying listeners. Nobody asked for it yet.

opentracing-util/src/main/java/io/opentracing/util/ThreadLocalScope.java
mdvorak: And thats exactly what i don't want to do, see discussion in #334.
Most common use-case is setting MDC (or ThreadContext for log4j) - and that uses copy-on-write immutable maps.
So, having to do following sequence:
1) onActivated(1) - set MDC (copy internal map)
2) onClose(1) - delete MDC (copy internal map) - **useless**
3) onActivated(2) - set MDC (copy internal map)
4) onClose(2) - delete MDC (copy internal map) - **useless**
5) onActivated(3) - set MDC (copy internal map)
wastes resources. Consumer cannot know, whether onActivated is coming right after onClose.
Thats why originally wanted only protected method for simple extension, or single method interface, so consumer can decide, what is best. This API should be simplistic, if you need something complex, you can copy whole class.

mdvorak: BTW thats why I originally proposed single method interface, with possible null as argument - its ugly, but efficient and versatile.

carlosalberto:
@yurishkuro Fair enough, sounds great to me ;)

opentracing-util/src/main/java/io/opentracing/util/ThreadLocalScope.java
yurishkuro: You're making assumption that MDC is the only use case. If you don't want to clear MDC from onClosed(), you don't have to, it does not negate the usefulness of the callback for other use cases. For example, see https://github.com/census-instrumentation/opencensus-specs/issues/247

mdvorak: You have to clear it in onClosed(), because you can't leave it set on last deactivation (updated example above, sorry I left it out).
Not saying its only use case, but its use case that should be supported. Give users freedom to do what they need, don't force them to follow exact and strict pattern.

Only model, that would actually fulfill both use-cases IMO, is ugly but flexible:
```
interface ScopeListener {
  void onSpanChanged(@Nullable Span activeSpan, @Nullable Span deactivatedSpan);
}
```

Or, I would revisit protected method and simply inheritance - do everyone whatever they want.
```
class ThreadLocalScopeManager {
  protected void setThreadScope(ThreadLocalScope scope) {
    this.tlsScope.set(scope);
  }
}
```
this can serve as "around advice". And it would make the class extensible in many ways, which I don't think is a bad thing.

thegreystone:
So, with the ScopeListener, I would need to track the association between an onActivate and a close of a scope myself, most likely with a thread local. For example, if I start a JFR event in a scope, I would have to add it in a thread local, to close the right one once close is hit.

Brave, for example, uses the concept of span and scope decorators. Upon activation, the scope decorator would expect a new scope to be returned (see ScopeDecorator#decorateScope(...)), into which the event could be pushed and closed when the scope itself is closed. No thread locals required.

sjoerdtalsma:
@thegreystone I like the decorator idea and I think it has come up before.  
However, as I understand it, each tracer can currently assume that all spans are 'their own'. 
How do you propose to allow each tracer operate on decorated spans, expose an _undecorate_ or _unwrap_ method of sorts?
Could it be an idea work this out in a separate PR?

opentracing-util/src/main/java/io/opentracing/util/ScopeListener.java
```diff
@@ -0,0 +1,44 @@
...
+ * @see ThreadLocalScopeManager
+ */
+public interface ScopeListener {
+
+    /**
+     * Called whenever a scope was activated (changed).
+     *
+     * @param span Activated span. Never null.
+     */
+    void onActivated(Span span);
```
felixbarny: I think the method name should reflect whether the activation happens before or after invoking this method. I'd propose `beforeActivated` and `afterClosed`

opentracing-util/src/main/java/io/opentracing/util/ThreadLocalScopeManager.java
```diff
@@ -19,11 +19,33 @@
...
+ * Optionally supports {@link ScopeListener}, to perform additional actions when scope is changed for given thread.
+ * Listener methods are always called synchronously on the same thread, right after activation (meaning {@link #active()}
+ * already returns new a scope).
  *
  * @see ThreadLocalScope
  */
 public class ThreadLocalScopeManager implements ScopeManager {
+
     final ThreadLocal<ThreadLocalScope> tlsScope = new ThreadLocal<ThreadLocalScope>();
+    final ScopeListener listener;
```
felixbarny: Consider making it a list or adding a `CompoundScopeListener`, which maintains a list of `ScopeListeners`

mdvorak:
Moving discussion to bottom.

We've identified 2 requirements:

1. Perform action with span at activation and deactivation, while providing said scope in an argument, for easy handling, as needed in census-instrumentation/opencensus-specs#247
1. Ability to handle change of span (both activation and deactivation of nested scope) in single operation, for MDC update (avoid unnecessary modifications) - this might apply to other handlers as well

Only possible solution I can think of is a bit ugly but universal
```java
interface ScopeListener {
  void onSpanChanged(@Nullable Span activeSpan, @Nullable Span deactivatedSpan);
}
```

I can't think of use case, where I would need handler before action. If so, you can always replace whole ScopeManager with your own.. Which should still be normal option for anyone requesting special behavior. I just wanted simple way to add configuring MDC for logging, without need to copy-paste scope manager into every app.

opentracing-util/src/main/java/io/opentracing/util/ThreadLocalScope.java
mdvorak: Moved discussion to bottom, it will get lost here on another push.

opentracing-util/src/main/java/io/opentracing/util/ScopeListener.java
carlosalberto: I agree on this. It's clearer having a `before` or `after` prefix.

yurishkuro: good call

opentracing-util/src/main/java/io/opentracing/util/ThreadLocalScopeManager.java
yurishkuro: composite listener can be another PR, imo, if someone actually needs it.

opentracing-util/src/main/java/io/opentracing/util/NoopScopeListener.java
```diff
@@ -0,0 +1,34 @@
...
+ * Unless required by applicable law or agreed to in writing, software distributed under the License
+ * is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express
+ * or implied. See the License for the specific language governing permissions and limitations under
+ * the License.
+ */
+package io.opentracing.util;
+
+import io.opentracing.Span;
+
+public interface NoopScopeListener extends ScopeListener {
```
carlosalberto: I'm wondering if we really need to expose this one as public.

yurishkuro:
@mdvorak I don't think we have an agreement on (2) to _have a single operation_. Using lambda or 2-method class is, imo, a very minor thing compared to the cleanliness and intuitiveness of the API.

mdvorak:
> @mdvorak I don't think we have an agreement on (2) to _have a single operation_. Using lambda or 2-method class is, imo, a very minor thing compared to the cleanliness and intuitiveness of the API.

Sorry I mentioned lambda originally, but read carefully previous post. Problem for two-method handler is an implementation, where deactivation always called before activation creates performance (or other) problems. You force implementation to have two separate handlers, when it actually needs to handle both at one step.

tylerbenson:
Here's a suggestion...  Have the `onSpanChanged` method be the primary interface, but include an adapter that converts it to the style that @yurishkuro wants.  Given the constraints proposed by @mdvorak I don't think it could go the other way.

What does the prior art look like?  How does this compare with Brave/Census/etc?

felixbarny:
What about adding an argument for the previously active/now active span.

* `void beforeActivate(Span spanToActivate, @Nullable Span previouslyActive);`
* `void afterDeactivate(Span deactivatedSpan, @Nullable Span nowActive);`

@thegreystone This should eliminate the need for maintaining a separate `ThreadLocal`, right? The advantage compared to Scope wrappers is that this does not allocate additional memory when creating scopes.

@mdvorak would that also solve your MDC use case? You will only have to remove the MDC context if `nowActive` is `null` (when you deactivate the first span of that thread).

mdvorak:
@felixbarny Yes, it would be possible to create efficient handler, but I still find it very complicated to do very simple thing. Also, relying on behavior, where that when `afterDeactivate` is called with `nowActive!=null` can be ignored, and I can expect `beforeActivate` right away seems error prone. From API point of view, I would be relying on specific implementation behavior.

2) I already proposed to provide both previous and next spans, so I agree with you on first point.

felixbarny:
> Also, relying on behavior, where that when afterDeactivate is called with nowActive!=null can be ignored

How is this different when having a single method?

> From API point of view, I would be relying on specific implementation behavior.

Not sure what is implementation specific here - the behavior is clearly specified and does not depend on the implementation.

thegreystone:
> @thegreystone This should eliminate the need for maintaining a separate `ThreadLocal`, right? The advantage compared to Scope wrappers is that this does not allocate additional memory when creating scopes.

Not sure how this would help. I would still need to propagate the event from beforeActivate to afterDeactivate somehow, and the only way I would know how is to use a thread local (in the scope case; in the SpanListener case all bets are off).

thegreystone:
> @thegreystone I like the decorator idea and I think it has come up before.
> However, as I understand it, each tracer can currently assume that all spans are 'their own'.
> How do you propose to allow each tracer operate on decorated spans, expose an _undecorate_ or _unwrap_ method of sorts?

True, doing this in a general fashion requires a bit of extra consideration. Brave simply returns implementations of their own scope interface, and assume they are the only player in town. 

> Could it be an idea work this out in a separate PR?

If we have decorators, we probably do not need listeners. And if we have listeners, it will seem a bit excessive to have decorators too. It may be worth it to consider the options. Anyone from Brave involved in OpenTracing?

thegreystone:
Here are some Brave decorators:
https://github.com/openzipkin/brave/tree/master/context

The JFR use case is the closest to my heart.

mdvorak:
Scope decorators are great, and that was my first idea, except with current implementation it does not work reliably, because of this if:
```java
public class ThreadLocalScope implements Scope {
    @Override
    public void close() {
        if (scopeManager.tlsScope.get() != this) {
            // This shouldn't happen if users call methods in the expected order. Bail out.
            return;
        }
```
decorator can never know, whether scope was actually closed or not.

carlosalberto:
@mdvorak Besides what others already commented/asked, I suggest you cooking a small case under `testbed` -if you have enough cycles- to show why you need a specific event design (and why the simple event @yurishkuro has in mind does not suit your needs).  I know you posted a small explanation regarding MDC, but having actual code could throw additional light into it ;)

sjoerdtalsma:
> decorator can never know, whether scope was actually closed or not

Don't let that influence the listener vs. decorator discussion please. If that's a 'defect' of the current scope manager implementation, we could address that separately from the decorator issue. We could update scope manager to handle out-of-order closing in a predictable manner for example.

felixbarny:
>> @thegreystone This should eliminate the need for maintaining a separate ThreadLocal, right? The advantage compared to Scope wrappers is that this does not allocate additional memory when creating scopes.
>
> Not sure how this would help. I would still need to propagate the event from beforeActivate to afterDeactivate somehow, and the only way I would know how is to use a thread local (in the scope case; in the SpanListener case all bets are off).

I misunderstood what the problem is. Looking at brave's [JfrScopeDecorator](https://github.com/openzipkin/brave/blob/master/context/jfr/src/main/java/brave/context/jfr/JfrScopeDecorator.java) I understand now. You have to have a reference to the `Event` instance which has been created on activation when the deactivation happens.


I see two designs to achieve that:

**Brave-like scope wrappers**
* Pros:
  * Prooven design
  * Simplicity
* Cons:
  * Additional allocations for every scope, proportional to the number of scope wrappers
  * The type of the scope changes, so we will likely need a `unwrap` method

**Context object**

Introducing a span context object which allows to attach custom key/value pairs to it. You can add custom context objects in the before callback and retrieve them from the map in the after callback.

`ScopeContext`
* `void add(String key, Object value)` (lazily allocates the underlying map)
* `<T> T get(String key)`

`ScopeListener`
* `void beforeActivate(Span spanToActivate, ScopeContext context);`
* `void afterDeactivate(Span deactivatedSpan, ScopeContext context);`

* Pros
  * The type of the scope does not change as it is not wrapped
  * Even though we have to allocate `ScopeContext` objects, we only have to allocate once, no matter how many `ScopeListener`s are registered
* Cons
  * Allocation of the `ScopeContext` object for each scope event
  * Adding things to a map causes allocations (the `Map` itself and `Map.Entry` objects)
  * More complicated to work with

Example
```java
public class ExampleScopeListener implements ScopeListener {
    @Override
    public void beforeActivate(Span spanToActivate, ScopeContext context) {
        context.add("foo", "bar");
    }

    @Override
    public void afterDeactivate(Span deactivatedSpan, ScopeContext context) {
        String foo = context.get("foo");
    }
}
```

carlosalberto:
> Introducing a span context object which allows to attach custom key/value pairs to it. You can add custom context objects in the before callback and retrieve them from the map in the after callback.

Btw, `ScopeContext` has been discussed a little bit in the past, so it's probably a good time to re-take it.

felixbarny:
Can you link to that discussion or summarize some of the discussed use cases?

carlosalberto:
Hey @felixbarny 

This is the Issue: https://github.com/opentracing/specification/issues/114

dschulten:
Since this has been dormant for a while, I wanted to mention that there is an opentracing-contrib module which can do sth quite similar: https://github.com/opentracing-contrib/java-api-extensions

whiskeysierra:
Not really. You can't observe Scopes with that.

dschulten:
> Not really. You can't observe Scopes with that.

That is true, you observe when a `Span` starts and finishes, and you can track when the root span finishes (i.e. traceId==spanId). Which is why I said it is similar. At least one can cover the MDC use case with it.

Out of curiosity: What is the compelling reason why one needs to observe a `ThreadLocalScope`?

mdvorak:
I believe I discussed that somewhere - Span is completely different from Scope:
Scope is technical, Span is logical
One span can "span" across different threads, especially in reactive app. Scope is activated and deactivated on every thread change, since it is thread-based. 
So, when you want to observe tracing output, Span is your thing. If you need to extend tracing library, you might need Scopes.

PR #343, opened 2019-04-17
circleci support for opentracing-java

unixlabs:
build jar files using circleci

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -1,0 +1,27 @@
+version: 2 # cirlcir ci build process for OpenTracing API for Java
+jobs:
+  build:
+    branches:
+      only:
+        - develop
+    docker:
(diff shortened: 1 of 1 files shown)


.circleci/config.yml
```diff
@@ -0,0 +1,27 @@
...
+           command: mvn -f opentracing-mock/pom.xml clean install
+      - run:
+           name: build jar file for noop service
+           command: mvn -f opentracing-noop/pom.xml clean install
+      - run:
+           name: build jar file for testbed
+           command: mvn -f opentracing-testbed/pom.xml clean install
+      - run:
+           name: build jar file for util
+           command: mvn -f opentracing-util/pom.xml clean install
```
yurishkuro: This seems unnecessary. The install target at the root level would build all jars in one go and will probably be much faster.

yurishkuro:
How do we test that this works?

PR #345, opened 2019-04-30
Set version to 0.33.1-SNAPSHOT.

carlosalberto:


pavolloffay:
The next snapshot version should be automatically updated by release script. If that is not the case something is broken

carlosalberto:
@pavolloffay As far as I remember, once a release/release-candidate happens, the version is automatically changed - but I did this update for people testing this branch and/or using the bintray snapshots ;)

carlosalberto:
Hey @sjoerdtalsma 

> Can't we integrate this automatically in the release job after a merge-back somehow?

Would be nice to have, yes ;)

PR #352, opened 2019-06-04
Update Tags.java

limfriend:
add tag 'error_kind'


whiskeysierra:
https://static.javadoc.io/io.opentracing/opentracing-api/0.33.0/io/opentracing/log/Fields.html#ERROR_KIND

It's a log field, not a tag.

whiskeysierra:
See also https://github.com/opentracing/specification/blob/master/semantic_conventions.md

limfriend:
> https://static.javadoc.io/io.opentracing/opentracing-api/0.33.0/io/opentracing/log/Fields.html#ERROR_KIND
> 
> It's a log field, not a tag.
'error.kind' is very important , we should be easy to  search and stat it.   it is difficulty to search when ''error.kind' is log field

PR #359, opened 2019-08-12
Add method finishedTraces to MockTracer.

### LINKED ISSUE #277: Add method finishedTraces to MockTracer
MockTracer has `finishedSpans` method, however in tests it's necessary to assert if the spans are in the same trace, therefore, a method like `List<List<MockSpan>> finishedTraces` would be useful.

qudongfang:
Resolves #277

### DIFF
--- a/opentracing-mock/src/main/java/io/opentracing/mock/MockTracer.java
+++ b/opentracing-mock/src/main/java/io/opentracing/mock/MockTracer.java
@@ -84,7 +84,7 @@ public class MockTracer implements Tracer {
     }
 
     /**
-     * @return a copy of all finish()ed MockSpans started by this MockTracer (since construction or the last call to
+     * @return all finish()ed MockSpans started by this MockTracer (since construction or the last call to
      * MockTracer.reset()).
      *
      * @see MockTracer#reset()
@@ -94,15 +94,15 @@ public class MockTracer implements Tracer {
     }
 
     /**
-     * @return a copy of all finish()ed Traces(Spans) started by this MockTracer grouped by traceId and spanId in HashMap format.
+     * @return all finish()ed Traces(Spans) started by this MockTracer grouped by traceId and spanId in HashMap format.
      */
-    public Map<String, Map<String, Span>> finishedTraces() {
-        Map<String, Map<String, Span>> result = new LinkedHashMap<>();
+    public Map<String, Map<String, MockSpan>> finishedTraces() {
+        Map<String, Map<String, MockSpan>> result = new LinkedHashMap<>();
 
         for (MockSpan span: this.finishedSpans()) {
             String traceId = span.context().toTraceId();
 
-            Map<String, Span> spanId2Span = result.get(traceId);
+            Map<String, MockSpan> spanId2Span = result.get(traceId);
             if (null == spanId2Span) {
                 spanId2Span = new LinkedHashMap<>();
                 result.put(traceId, spanId2Span);
--- a/opentracing-mock/src/main/java/io/opentracing/mock/MockTracer.java
+++ b/opentracing-mock/src/main/java/io/opentracing/mock/MockTracer.java
@@ -24,6 +24,7 @@ import java.io.ObjectOutputStream;
 import java.io.IOException;
 import java.nio.ByteBuffer;
 import java.util.ArrayList;
+import java.util.Collections;
 import java.util.HashMap;
 import java.util.LinkedHashMap;
 import java.util.List;
@@ -90,7 +91,7 @@ public class MockTracer implements Tracer {
      * @see MockTracer#reset()
      */
     public synchronized List<MockSpan> finishedSpans() {
-        return new ArrayList<>(this.finishedSpans);
+        return Collections.unmodifiableList(new ArrayList<>(this.finishedSpans));
     }
 
     /**
@@ -112,7 +113,11 @@ public class MockTracer implements Tracer {
             spanId2Span.put(spanId, span);
         }
 
-        return result;
+        for (Map.Entry<String, Map<String, MockSpan>> entry : result.entrySet()) {
+            result.put(entry.getKey(), Collections.unmodifiableMap(entry.getValue()));
+        }
+
+        return Collections.unmodifiableMap(result);
     }
 
     /**
--- a/opentracing-mock/src/main/java/io/opentracing/mock/MockTracer.java
+++ b/opentracing-mock/src/main/java/io/opentracing/mock/MockTracer.java
@@ -93,6 +93,28 @@ public class MockTracer implements Tracer {
     }
 
     /**
+     * @return a copy of all finish()ed Traces(Spans) started by this MockTracer grouped by traceId and spanId in HashMap format.
+     */
+    public Map<String, Map<String, Span>> finishedTraces() {
+        Map<String, Map<String, Span>> result = new HashMap<>();
+
+        for (MockSpan span: this.finishedSpans()) {
+            String traceId = span.context().toTraceId();
+
+            Map<String, Span> spanId2Span = result.get(traceId);
+            if (null == spanId2Span) {
+                spanId2Span = new HashMap<>();
+                result.put(traceId, spanId2Span);
+            }
+
+            String spanId = span.context().toSpanId();
+            spanId2Span.put(spanId, span);
+        }
+
+        return result;
+    }
+
+    /**
      * Noop method called on {@link Span#finish()}.
      */
     protected void onSpanFinished(MockSpan mockSpan) {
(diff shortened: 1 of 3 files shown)


opentracing-mock/src/main/java/io/opentracing/mock/MockTracer.java
```diff
@@ -95,6 +92,28 @@ public synchronized void reset() {
         return new ArrayList<>(this.finishedSpans);
     }
 
+    /**
+     * @return a copy of all finish()ed Traces(Spans) started by this MockTracer grouped by traceId and spanId in HashMap format.
+     */
+    public Map<String, Map<String, Span>> finishedTraces() {
+        Map<String, Map<String, Span>> result = new HashMap<>();
```
sjoerdtalsma: Suggestion to use `LinkedHashMap` to keep insertion order from the original `List` ?

(both `new HashMap` calls)

sjoerdtalsma:
No other comments, thanks for the change

opentracing-mock/src/main/java/io/opentracing/mock/MockTracer.java
qudongfang: Brilliant!

qudongfang:
cc @pavolloffay

opentracing-mock/src/main/java/io/opentracing/mock/MockTracer.java
```diff
@@ -95,6 +93,28 @@ public synchronized void reset() {
         return new ArrayList<>(this.finishedSpans);
     }
 
+    /**
+     * @return a copy of all finish()ed Traces(Spans) started by this MockTracer grouped by traceId and spanId in HashMap format.
+     */
+    public Map<String, Map<String, Span>> finishedTraces() {
```
pavolloffay: It should return a MockSpan not span

opentracing-mock/src/main/java/io/opentracing/mock/MockTracer.java
```diff
@@ -95,6 +93,28 @@ public synchronized void reset() {
         return new ArrayList<>(this.finishedSpans);
     }
 
+    /**
+     * @return a copy of all finish()ed Traces(Spans) started by this MockTracer grouped by traceId and spanId in HashMap format.
```
pavolloffay: It does not return a copy

whiskeysierra:
Alternatively it could be made read-only/unmodifiable using
`Collections.unmodifiable*`. Has to be done deep though, i.e. to all
contained maps as well.

opentracing-mock/src/main/java/io/opentracing/mock/MockTracer.java
qudongfang: I thought It is better to use a more general return data type.

opentracing-mock/src/main/java/io/opentracing/mock/MockTracer.java
qudongfang: Thanks for pointing out, I just copied so words from the comments of `finishedSpans` and didn't realised it was wrong.

qudongfang:
> Alternatively it could be made read-only/unmodifiable using `Collections.unmodifiable*`

Thanks for your suggestion. I am not sure if we really need it to be a copy or read-only/unmodifiable.

pavolloffay:
+1 on unmodifiable

sjoerdtalsma:
Not sure on unmodifiable. Next call creates a new `Map` structure anyway.
Any pressing reasons for the unmodifiable, seeing that the nested maps add complexity?

carlosalberto:
Agreed, I don't think we *have to* add the unmodifiable part. For example, `finishedSpans()` is defined as:

```java
    public synchronized List<MockSpan> finishedSpans() {
        return new ArrayList<>(this.finishedSpans);
    }
```

qudongfang:
> +1 on unmodifiable

Done.


> +1 on unmodifiable

Done.

opentracing-mock/src/main/java/io/opentracing/mock/MockTracer.java
```diff
@@ -86,13 +85,39 @@ public synchronized void reset() {
...
     /**
-     * @return a copy of all finish()ed MockSpans started by this MockTracer (since construction or the last call to
+     * @return all finish()ed MockSpans started by this MockTracer (since construction or the last call to
      * MockTracer.reset()).
      *
      * @see MockTracer#reset()
      */
     public synchronized List<MockSpan> finishedSpans() {
-        return new ArrayList<>(this.finishedSpans);
+        return Collections.unmodifiableList(new ArrayList<>(this.finishedSpans));
```
whiskeysierra: That's a breaking change. Client code may rely on the fact that it gets a copy right now which it can safely mutate.

qudongfang: Yes, It is a breaking change.

I think it is not right to make `finishedTraces` unmodifiable while `finishedSpans` modifiable. @pavolloffay Shall we make them all modifiable?

whiskeysierra:
I wasn't aware that the one right returns a copy/snapshot. I'd keep
consistency and do the same then.

opentracing-mock/src/main/java/io/opentracing/mock/MockTracer.java
```diff
@@ -86,13 +85,39 @@ public synchronized void reset() {
...
+    }
+
+    /**
+     * @return all finish()ed Traces(Spans) started by this MockTracer grouped by traceId and spanId in HashMap format.
+     */
+    public Map<String, Map<String, MockSpan>> finishedTraces() {
```
sjoerdtalsma: I just noticed that method `finishedSpans` is **synchronized**.

May be a good idea to make `finishedTraces` synchronized as well?

opentracing-mock/src/main/java/io/opentracing/mock/MockTracer.java
pavolloffay: This should be reverted. In the other method we can return normal map for the consistency.

opentracing-mock/src/main/java/io/opentracing/mock/MockTracer.java
carlosalberto: +1

opentracing-mock/src/main/java/io/opentracing/mock/MockTracer.java
carlosalberto: > This should be reverted. In the other method we can return normal map for the consistency.

Agreed as well.

qudongfang: Done.

opentracing-mock/src/main/java/io/opentracing/mock/MockTracer.java
qudongfang: Done.

opentracing-mock/src/main/java/io/opentracing/mock/MockTracer.java
```diff
@@ -86,7 +84,7 @@ public synchronized void reset() {
     }
 
     /**
-     * @return a copy of all finish()ed MockSpans started by this MockTracer (since construction or the last call to
+     * @return all finish()ed MockSpans started by this MockTracer (since construction or the last call to
```
carlosalberto: Sorry, could we keep the original, mentioning this is a copy? ;) 

(Same for the new method)

carlosalberto:
Hey, thanks a lot for the updates, appreciated. It looks good now, and just left a small comment about keeping the comment regarding the returned `MockSpans` being a copy. After that we should be good to merge ;)

opentracing-mock/src/main/java/io/opentracing/mock/MockTracer.java
qudongfang: OK.
At the sametime, It is a copy of the `MockSpan` list not a deep copy of `MockSpans`.

opentracing-mock/src/main/java/io/opentracing/mock/MockTracer.java
```diff
@@ -126,8 +146,6 @@ protected void onSpanFinished(MockSpan mockSpan) {
         };
 
         Propagator BINARY = new Propagator() {
-            static final int BUFFER_SIZE = 128;
```
carlosalberto: This change was not here before. We can clean up (potential) dead code in another PR, to keep changes uniform ;)

qudongfang: Sure thing, I have reverted the commit about this.

carlosalberto:
Merged, thanks!

PR #369, opened 2020-01-18
Log an error when Scope.close() is called on the incorrect object.

carlosalberto:
A few users have complained about not getting any warning when closing a non-active `Scope`, so this PR tries to do a simple warning.

(Prior-art wise, grpc's Context implementation does the very same ;) )

### DIFF
--- a/opentracing-util/src/main/java/io/opentracing/util/ThreadLocalScope.java
+++ b/opentracing-util/src/main/java/io/opentracing/util/ThreadLocalScope.java
@@ -17,6 +17,9 @@ import io.opentracing.Scope;
 import io.opentracing.ScopeManager;
 import io.opentracing.Span;
 
+import java.util.logging.Level;
+import java.util.logging.Logger;
+
 /**
  * {@link ThreadLocalScope} is a simple {@link Scope} implementation that relies on Java's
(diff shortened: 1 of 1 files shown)


carlosalberto:
Any opinion or similar on this one @yurishkuro @sjoerdtalsma ?

opentracing-util/src/main/java/io/opentracing/util/ThreadLocalScope.java
```diff
@@ -39,6 +44,8 @@
     public void close() {
         if (scopeManager.tlsScope.get() != this) {
             // This shouldn't happen if users call methods in the expected order. Bail out.
+            log.log(Level.SEVERE, "Scope being closed is not the active instance.",
```
sjoerdtalsma: Personally I would log a warning at this point I think.
Also, wrap the log in an `if` for the level since you forcefully get a stacktrace.

Side note; My own code is a little more lenient with out-of-sequence closing; it won't touch the active scope, but it will 'mark' the wrong one closed. So when the correct one is closed, it will be 'skipped' in the stack unwinding code since it was already closed out of sequence.

I hope you get the idea, sorry it's still rather early in the morning here :wink:

PR #372, opened 2020-02-19
Don't override the return type in buildSpan().

mmimica:
This makes it possible to override `buildSpan(String op)` of MockTracer and return another impl of `Tracer.SpanBuilder`.
A use-case for this would be, for example, to supply an instance of `SpanBuilder` that delegates to original `MockTracer.SpanBuilder`, but also performs additional logging.

### DIFF
--- a/opentracing-mock/src/main/java/io/opentracing/mock/MockTracer.java
+++ b/opentracing-mock/src/main/java/io/opentracing/mock/MockTracer.java
@@ -280,7 +280,7 @@ public class MockTracer implements Tracer {
     }
 
     @Override
-    public SpanBuilder buildSpan(String operationName) {
+    public Tracer.SpanBuilder buildSpan(String operationName) {
         return new SpanBuilder(operationName);
     }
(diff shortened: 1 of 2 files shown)


whiskeysierra:
The proper way of doing this would be composition, not inheritance. You can take a look at https://github.com/zalando/opentracing-toolbox/tree/master/opentracing-proxy to see an example.

mmimica:
> The proper way of doing this would be composition, not inheritance.

Indeed.

yurishkuro (changes requested):
if you extend MockTracer.SpanBuilder, then you do not need this change.

mmimica:
> if you extend MockTracer.SpanBuilder, then you do not need this change.

My extension to SpanBuilder is generic and has nothing to do with mocks, so I'm not going do that.

yurishkuro:
if you're using MockTracer, you should be extending its subcomponents.

You're proposing a breaking change.

PR #381, opened 2021-05-06
docs(Tracer): remove link to nonexistent method

dmedinag:
The previously referred method was removed from the API, hence it no longer exists. Must say too that semantic versioning was used wrong! Removing this method was a breaking change. The same as removing `ScopeManager.activate(Span span, boolean finishSpanOnClose)`. Please be more careful with the versioning of the libs, deprecating a method is better than removing directly :)


whiskeysierra:
Nope, it wasn't used incorrectly. OpenTracing never made it out of 0.y.z:

> Major version zero (0.y.z) is for initial development. Anything MAY change at any time. The public API SHOULD NOT be considered stable.
> 
> https://semver.org/#spec-item-4

venuprasadnaik (approved):

